PR #39, opened 2019-11-21
Add empty table state to MainTable

barrymcgee:
## Done

- Add empty table state to MainTable for when data is waiting on or has no data
- Driveby: fix linting error in Accordion.js

## QA

- Pull branch
- Run `yarn run serve`
- View Storybook at http://localhost:8403/
- Click to MainTable component
- View 'Empty' state
- Sanity check code

### DIFF
--- a/src/components/Accordion/Accordion.js
+++ b/src/components/Accordion/Accordion.js
@@ -22,7 +22,7 @@ const Accordion = ({ className, sections, ...props }) => {
       className={classNames(className, "p-accordion")}
       {...props}
       role="tablist"
-      aria-multiselect="true" //eslint-disable-line jsx-a11y/aria-props
+      ariaMultiselect="true" //eslint-disable-line jsx-a11y/aria-props
     >
       <ul className="p-accordion__list">
         {generateSections(sections, setExpanded, expanded)}
--- a/src/components/Accordion/Accordion.js
+++ b/src/components/Accordion/Accordion.js
@@ -22,7 +22,7 @@ const Accordion = ({ className, sections, ...props }) => {
       className={classNames(className, "p-accordion")}
       {...props}
       role="tablist"
-      ariaMultiselect="true" //eslint-disable-line jsx-a11y/aria-props
+      ariaMultiselectable="true"
     >
       <ul className="p-accordion__list">
         {generateSections(sections, setExpanded, expanded)}
--- a/src/components/MainTable/MainTable.js
+++ b/src/components/MainTable/MainTable.js
@@ -7,6 +7,8 @@ import TableRow from "../TableRow";
 import TableHeader from "../TableHeader";
 import TableCell from "../TableCell";
 
+import "./MainTable.scss";
+
 const updateSort = (setSortKey, setSortDirection, sortKey, sortDirection) => {
   let newDirection = null;
   if (sortDirection === "none") {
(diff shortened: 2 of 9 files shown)


barrymcgee:
Note: 

The table row on the header does not appear - Vanilla bug raised here: https://github.com/canonical-web-and-design/vanilla-framework/issues/2662

There's no defined design for an empty table in Vanilla, issue raised here: https://github.com/canonical-web-and-design/vanilla-framework/issues/2663

hatched:
I think we'll want this to be a bit more robust by allowing the consumer to define a component that's passed in for the empty state.

src/components/MainTable/MainTable.scss
```diff
@@ -0,0 +1,4 @@
+caption {
```
anthonydillon: Why is this style applied to MainTable? This should be a global style accompanied with an issue upstream.

src/components/MainTable/MainTable.js
```diff
@@ -82,60 +84,64 @@ const generateRows = (
...
+                {expandedContent}
+              </TableCell>
+            )}
+          </TableRow>
+        );
+      }
+    );
+    return <tbody>{rowItems}</tbody>;
+  } else {
+    return <caption>No data to display</caption>;
```
anthonydillon: Why caption and not just text?

barrymcgee: Raw text dropped inside a `<table>` is semantically meaningless and AFAIK invalid. As per the [table HTML spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table), I believe caption is the most appropriate choice here.

src/components/MainTable/MainTable.scss
barrymcgee: Issue raised upstream here: https://github.com/canonical-web-and-design/vanilla-framework/issues/2663

I didn't want to apply it globally as I see this as a local override until it has been agreed upstream in Vanilla?

anthonydillon: Thanks, could you add a comment of the issue so the styling can be removed when resolved. I'm allergic to custom styling.

src/components/MainTable/MainTable.js
huwshimi: I think this will need to be supplied from a prop.

huwshimi:
@anthonydillon this is already different from what we need for MAAS. I'm not sure a single case such as "No data to display" is very helpful. Why is there not data? Is it because it is loading? Was there an error? What do I need to do to add data?

barrymcgee:
I've added the ability here to supply the text via a prop. I think all the other questions around this will be addressed upstream in Vanilla itself and trickle-down here to the implementation. In the meantime, this PR will unblock us in the JAAS Dashboard.

src/components/Accordion/Accordion.js
```diff
@@ -22,7 +22,7 @@ const Accordion = ({ className, sections, ...props }) => {
       className={classNames(className, "p-accordion")}
       {...props}
       role="tablist"
-      aria-multiselect="true" //eslint-disable-line jsx-a11y/aria-props
+      ariaMultiselect="true" //eslint-disable-line jsx-a11y/aria-props
```
hatched: Did you mean to use `area-multiselectable` ? 

https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-multiselectable

src/components/Table/Table.js
```diff
@@ -17,6 +17,7 @@ const Table = ({
       "p-table--sortable": sortable,
       "p-table-expanding": expanding
     })}
+    style={{ captionSide: "bottom" }}
```
hatched: Cool style, I didn't know about this one!

Curious why this is here and not part of the `MainTable.scss`?

src/components/MainTable/MainTable.scss
hatched: aiui this will actually be applied globally as there is no `main-table` wrapper style. It's just defined here....which may make it challenging to debug later.

Can you add this wrapper class?

src/components/MainTable/MainTable.js
```diff
@@ -252,4 +263,9 @@ MainTable.propTypes = {
   sortFunction: PropTypes.func
 };
 
+// Specifies the default values for props:
+MainTable.defaultProps = {
+  emptyStateMsg: "No data to display"
+};
```
squidsoup: There's no need to use defaultProps with ES6 default values, better to set this on l151.

hatched:
@huwshimi could you please take a look at this?

src/components/MainTable/MainTable.js
```diff
@@ -81,6 +84,10 @@ const generateRows = (
   sortable,
   sortFunction
 ) => {
+  // If the table has no rows, return empty state message
+  if (Object.entries(rows).length === 0) {
```
huwshimi: This should also check that `emptyStateMsg` has been provided.

src/components/MainTable/MainTable.js
```diff
@@ -141,6 +148,7 @@ const generateRows = (
 const MainTable = ({
   defaultSort,
   defaultSortDirection,
+  emptyStateMsg = "No data to display",
```
huwshimi: I don't think this should have a default so that people can have no message if they want to.

src/components/MainTable/MainTable.js
```diff
@@ -219,6 +229,7 @@ const MainTable = ({
 MainTable.propTypes = {
   defaultSort: PropTypes.string,
   defaultSortDirection: PropTypes.oneOf(["ascending", "descending"]),
+  emptyStateMsg: PropTypes.string,
```
huwshimi: I think this should be `PropTypes.node` so that more complex content can be passed in as well. Maybe this should just be called `emptyState`. Also, might be good to add a doc string as well (like this: https://github.com/canonical-web-and-design/react-components/blob/master/src/components/Accordion/Accordion.js#L34), something like:

```js
/**
 * A state that will be shown when no rows are passed to the table.
 */
emptyState: PropTypes.node,
```

src/components/MainTable/MainTable.stories.mdx
```diff
@@ -331,3 +331,18 @@ This is a [React](https://reactjs.org/) component to support many table use case
     />
   </Story>
 </Preview>
+
+### Empty
```
huwshimi: Thanks for adding this!

anthonydillon:
@barrymcgee it looks like there are legitimate comments here from Huw. Could you address them please.

anthonydillon:
@huwshimi do you mind doing a review as your comments were last

src/components/MainTable/MainTable.js
```diff
@@ -219,6 +229,10 @@ const MainTable = ({
 MainTable.propTypes = {
   defaultSort: PropTypes.string,
   defaultSortDirection: PropTypes.oneOf(["ascending", "descending"]),
+  /**
+   * A state that will be shown when no rows are passed to the table.
+   */
+  emptyState: PropTypes.node,
```
anthonydillon: You're not displaying this node if it is provided.

barrymcgee: See line 89 

Screenshot 2020-01-13 at 14 19 56

PR #45, opened 2019-12-02
Fix Notification timeout function not being executed.

Caleb-Ellis:
## Done
- Fixed Notification timeout function not being executed, despite being provided timeout and close props.

## QA
- `yarn link` this project and `/node_modules/react` and link them in `maas-ui`.
- Go to <maas-ui-url>:8400/MAAS/r/account/prefs/details.
- Change your details and check that the notification closes after five seconds.

Fixes canonical-web-and-design/maas-ui#517

### DIFF
--- a/src/components/Notification/Notification.js
+++ b/src/components/Notification/Notification.js
@@ -21,7 +21,7 @@ const Notification = ({
   const timeoutId = useRef();
 
   useEffect(() => {
-    if (timeout && close && !timeoutId.current) {
+    if (timeout && close) {
       timeoutId.current = setTimeout(() => close(), timeout);
     }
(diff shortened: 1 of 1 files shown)


src/components/Notification/Notification.js
```diff
@@ -21,7 +21,7 @@ const Notification = ({
   const timeoutId = useRef();
 
   useEffect(() => {
-    if (timeout && close && !timeoutId.current) {
+    if (timeout && close) {
```
Caleb-Ellis: So I *think* the issue here was that the timeout was being created then immediately cleared (useEffect weirdness?), but since `timeoutId.current` was no longer falsy it didn't create a new one.

PR #75, opened 2020-03-06
SearchBox external state control

### LINKED ISSUE #76: Allow SearchBox component to be controlled by external state

huwshimi:
Done:
- Allow the search box to be controlled by external state.

QA:
- There's probably no easy way to QA this change without using the component. Make sure the component looks and functions as before.

Fixes #76

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dist
 
 .docker-project
 .yarn.*
+.dotrun.json
 
 npm-debug.log*
 yarn-debug.log*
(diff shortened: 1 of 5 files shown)


huwshimi:
@Caleb-Ellis the external state would be for when you want to manipulate the input value externally from the component so you need the input to be a controlled component (https://reactjs.org/docs/forms.html#controlled-components).

In MAAS we need to do this when the filters change the search input value.

PR #82, opened 2020-03-11
Prepare 0.4.0 release

huwshimi:


huwshimi:
@Caleb-Ellis that's right, the docs get built and stored in the repo, so most of the changes are just rebuilt files.

PR #84, opened 2020-03-12
Add initial cypress spec.

squidsoup:
## Done
Add initial integration tests with cypress.

## QA
Run:
```PORT=<PORT> yarn test-cypress```

You should see two tests for Accordion pass.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -54,7 +56,11 @@
     "prepublishOnly": "yarn clean && yarn install && yarn build",
     "serve": "yarn docs",
     "start": "yarn docs",
-    "test": "react-scripts test"
+    "test": "react-scripts test",
+    "test-cypress": "concurrently 'yarn run docs' 'yarn run cypress:test' -k -s first",
```
huwshimi: `concurrently` is missing as a dep.

squidsoup: > `concurrently` is missing as a dep.

oops, so it is, thanks. pushed.

PR #129, opened 2020-05-18
Add prop to pass the selected value to the parent for the Select 

solazio:
## Done

- Add prop to pass the selected value to the parent for the Select component

## QA 

- Go to `/?path=/docs/select--select` and see the new prop of the `Select` component
- See the state of the parent component is updated with the `value` of the selected option when the selection is changed

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -18,6 +18,7 @@ const Select = ({
   help,
   id,
   label,
+  onChange,
   options,
   required,
   stacked,
@@ -34,13 +35,12 @@ const Select = ({
       label={label}
       required={required}
       stacked={stacked}
-      success={success}
-    >
+      success={success}>
       <select
         className={classNames("p-form-validation__input", className)}
         id={id}
-        {...props}
-      >
+        onChange={evt => onChange(evt.target.value)}
+        {...props}>
         {generateOptions(options)}
       </select>
     </Field>
@@ -54,6 +54,7 @@ Select.propTypes = {
   help: PropTypes.string,
   id: PropTypes.string,
   label: PropTypes.string,
+  onChange: PropTypes.func,
   options: PropTypes.arrayOf(
     PropTypes.shape({
       label: PropTypes.string.isRequired,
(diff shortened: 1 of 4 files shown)


src/components/Select/Select.js
```diff
@@ -18,6 +18,7 @@ const Select = ({
   help,
   id,
   label,
+  onSelectionChange,
```
squidsoup, line 21: this is typically just called `onChange`

src/components/Select/Select.js
```diff
@@ -54,6 +56,11 @@ Select.propTypes = {
   help: PropTypes.string,
   id: PropTypes.string,
   label: PropTypes.string,
+  /**
+   * Optional function that is called when the selection is changed.
+   * The function gets the selected option as an argument.
+   */
+  onSelectionChange: PropTypes.func,
```
squidsoup, line 63: This comment is a bit redundant, propTypes that are `required` are explicitly marked as such.

solazio:
@squidsoup I will add the test and change the name. Should I remove the whole comment, or only the “optional” part? Thanks

squidsoup:
> @squidsoup I will add the test and change the name. Should I remove the whole comment, or only the “optional” part? Thanks

Cool thanks ovi. Up to you wrt the comment, but I think everyone should already know what an `onChange` event is, well hopefully :)

solazio:
@squidsoup can you please have another look at this

src/components/Select/Select.js
```diff
@@ -34,13 +35,12 @@ const Select = ({
...
       stacked={stacked}
-      success={success}
-    >
+      success={success}>
       <select
         className={classNames("p-form-validation__input", className)}
         id={id}
-        {...props}
-      >
+        onChange={evt => onChange(evt.target.value)}
```
Lukewh, line 42: can we just pass the evt through. also the onChange is optional, so we should only call it if it's defined.

```suggestion
        onChange={evt => onChange && onChange(evt)}
```

src/components/Select/Select.test.js
```diff
@@ -7,18 +7,34 @@ describe("Select ", () => {
   it("renders", () => {
     const wrapper = shallow(
       <Select
-        id="test-id"
+        id='test-id'
```
Lukewh, line 10: single quotes!? You having a giggle?

PR #152, opened 2020-06-16
Migrate Tabs component from maas-ui.

### LINKED ISSUE #150: Migrate Tabs component from MAAS

Caleb-Ellis:
## Done
- Migrated [Tabs component from maas-ui](https://github.com/canonical-web-and-design/maas-ui/tree/master/ui/src/app/base/components/Tabs) with some modifications to make it more generic
  - You can now give links a `component` prop in cases where you might want normal anchor links or react-router links
  - Removed the `noBorder` prop - that's not in Vanilla and should just be handled by `listClassName` anyway

## QA
- Run `yarn docs` and check that the canvas and docs for Tabs are good

Fixes #150

### DIFF
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -1,0 +1,82 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+
+const Tabs = ({ className, links, listClassName }) => {
+  return (
+    <nav className={classNames("p-tabs", className)}>
+      <ul className={classNames("p-tabs__list", listClassName)}>
+        {links.map((link) => {
+          const {
+            active,
+            className,
+            component,
+            label,
+            listItemClassName,
+            ...rest
+          } = link;
+          const Component = component || "a";
+          return (
+            <li
+              className={classNames("p-tabs__item", listItemClassName)}
+              key={label}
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,58 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+
+const Tabs = ({ links, listClassName, navClassName }) => {
+  return (
+    <nav className={classNames("p-tabs", navClassName)}>
```
huwshimi, line 7: I think because this is the wrapping element we could just have `className` instead of `navClassName`.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,58 @@
...
+              </Component>
+            </li>
+          );
+        })}
+      </ul>
+    </nav>
+  );
+};
+
+Tabs.propTypes = {
```
huwshimi, line 40: It'd be good to document these props, kinda like https://github.com/canonical-web-and-design/react-components/blob/master/src/components/Accordion/Accordion.js#L49

Caleb-Ellis: Done!

PR #154, opened 2020-06-17
Add `titleElement` prop to accordion to support headings in titles

bartaz:
Vanilla 2.13 added an option to use headings in accordion titles.
This PR adds `titleElement` prop to the `Accordion` (and `AccordionSection`) components to enable this.

Screenshot 2020-06-17 at 15 12 46

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


bartaz:
@huwshimi Sure, thanks.

I removed snapshot from Accordion test (and just verify if prop is passed down to AccordionSecition). In AccordionSection test I kept the snapshot (to make sure we don't mess up the HTML structure at some point), but also explicitly check for heading element with the class.

huwshimi:
@bartaz Looks good, thanks for updating the changelog too.

PR #157, opened 2020-06-22
Add basic structure of Search and Filter component

barrymcgee:
Run Storybook locally and view: http://localhost:60078/?path=/docs/searchandfilter--default-story

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.test.js
+++ b/src/components/SearchAndFilter/SearchAndFilter.test.js
@@ -1,0 +1,25 @@
+import { shallow, mount } from "enzyme";
+import React from "react";
+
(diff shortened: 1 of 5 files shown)


src/components/SearchAndFilter/SearchAndFilter.test.js
```diff
@@ -0,0 +1,16 @@
...
+
+import SearchAndFilter from "./SearchAndFilter";
+
+describe("Search and filter ", () => {
+  it("renders", () => {
+    const wrapper = shallow(<SearchAndFilter />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("hide the clear button when there is no value in search box", () => {
```
anthonydillon, line 12: Can you add a test for onchange function call?

PR #163, opened 2020-06-22
Added basic contextual menu

barrymcgee:
Storybook: http://localhost:61502/?path=/docs/contextualmenu--default-story


src/components/ContextualMenu/ContextualMenu.test.js
```diff
@@ -0,0 +1,18 @@
+import { shallow } from "enzyme";
+import React from "react";
+
+import ContextualMenu from "./ContextualMenu";
+
+describe("Contextual Menu ", () => {
+  it("renders instance but not visible", () => {
+    const wrapper = shallow(<ContextualMenu>Test content</ContextualMenu>);
+    expect(wrapper).toMatchSnapshot();
```
hatched, line 9: Snapshots in this case aren't the best idea, you should have an assertion for what you're actually testing.

squidsoup: Agree with @hatched, snapshot tests in isolation are not sufficient, but okay to have _both_ I think.

src/components/ContextualMenu/ContextualMenu.test.js
```diff
@@ -0,0 +1,22 @@
+import { shallow } from "enzyme";
+import React from "react";
+
+import ContextualMenu from "./ContextualMenu";
+
+describe("Contextual Menu ", () => {
+  it("renders instance but not visible", () => {
```
anthonydillon, line 7: Are you missing a snapshot test?

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -0,0 +1,34 @@
...
+
+import SearchBox from "../SearchBox";
+
+const SearchAndFilter = ({ externallyControlled = false, onChange }) => {
+  const [searchTerm, setSearchTerm] = useState("");
+
+  const searchOnChange = (searchTerm) => {
+    setSearchTerm(searchTerm);
+    if (externallyControlled) {
+      onChange(searchTerm);
```
anthonydillon, line 12: You should check onChange exists here otherwise you will get unexpected errors.

PR #170, opened 2020-06-23
Add basic filter panel to Search and Filter component

barrymcgee:
## Done 

Add basic filter panel to Search and Filter component

## QA

- Fire up Storybook: http://localhost:65491/?path=/story/searchandfilter--default-story
- Click in input to activate panel
- Press Escape to close
- Click outside input or panel to close

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1335

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.js
+++ b/src/components/SearchAndFilter/SearchAndFilter.js
@@ -1,10 +1,15 @@
-import React, { useState } from "react";
+import React, { useState, useEffect, useRef } from "react";
 import PropTypes from "prop-types";
 
 import SearchBox from "../SearchBox";
+import ContextualMenu from "../ContextualMenu";
+
+import "./SearchAndFilter.scss";
 
 const SearchAndFilter = ({ externallyControlled = false, onChange }) => {
   const [searchTerm, setSearchTerm] = useState("");
+  const [filterPanelVisible, setFilterPanelVisible] = useState(false);
+  const filterPanelRef = useRef();
 
   const searchOnChange = (searchTerm) => {
     setSearchTerm(searchTerm);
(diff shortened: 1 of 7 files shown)


src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -13,14 +18,56 @@ const SearchAndFilter = ({ externallyControlled = false, onChange }) => {
...
         externallyControlled={externallyControlled}
         placeholder="Search and filter"
         onChange={(searchTerm) => searchOnChange(searchTerm)}
+        onFocus={() => setFilterPanelVisible(true)}
         value={searchTerm}
       />
+      <div
+        className="search-and-filter__panel"
+        ref={filterPanelRef}
+        data-visible={filterPanelVisible}
```
anthonydillon, line 67: Why not use `aria-hidden` and make it accessible from the get go?

src/components/SearchAndFilter/SearchAndFilter.scss
```diff
@@ -0,0 +1,33 @@
...
+  .p-contextual-menu__dropdown {
+    top: 0;
+    left: 0;
+  }
+
+  &__panel {
+    padding: 0;
+    transition: opacity 0.25s;
+
+    &[data-visible="false"] {
```
anthonydillon, line 28: Same here you should target `aria-hidden`

src/components/SearchAndFilter/SearchAndFilter.test.js
```diff
@@ -22,4 +22,18 @@ describe("Search and filter ", () => {
     wrapper.find(".p-search-box__input").simulate("change");
     expect(mockOnChange).toHaveBeenCalledTimes(1);
   });
+
+  it("shows panel on focus", () => {
```
anthonydillon, line 26: Can you add a test for the click or is that difficult?

src/components/SearchAndFilter/SearchAndFilter.scss
```diff
@@ -0,0 +1,33 @@
+.search-and-filter {
+  position: relative;
```
anthonydillon, line 2: I don't see any internal position so I don't think we need this?

anthonydillon:
## QA 
- If you click on the field the panel is revealed but clicking on the field again closes it and it doesn't open again.
- If you click the field the panel is revealed but if you tab off you can't hit Esc to close the panel.

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: Good call 🙌🏻

src/components/SearchAndFilter/SearchAndFilter.test.js
barrymcgee: I don't think we need to test that the browser adds focus on click?

barrymcgee: Gotcha.. added

PR #171, opened 2020-06-25
Add filter panel sections

barrymcgee:
- Run Storybook
- http://localhost:65233/?path=/story/searchandfilter--external-state

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1327,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1329

### DIFF
--- a/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
+++ b/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
@@ -1,0 +1,20 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Chip from "../../Chip";
+
+import "./filter-panel-section.scss";
+
+const FilterPanelSection = ({ data }) => (
+  <div className="filter-panel-section">
+    <h3 className="filter-panel-section__heading">{data?.heading}</h3>
+    {data?.chips?.map((chip, i) => (
+      <Chip lead={chip.lead} value={chip.value} key={i} />
+    ))}
+  </div>
+);
+
+FilterPanelSection.propTypes = {
+  data: PropTypes.any,
+};
+
+export default FilterPanelSection;
--- a/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
+++ b/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
@@ -9,8 +9,12 @@ const FilterPanelSection = ({ data }) => (
     {data?.heading && (
       <h3 className="filter-panel-section__heading">{data.heading}</h3>
     )}
-    {data?.chips?.map((chip, i) => (
-      <Chip lead={chip.lead} value={chip.value} key={i} />
+    {data?.chips?.map((chip) => (
+      <Chip
+        lead={chip.lead}
+        value={chip.value}
+        key={`${chip.lead}+${chip.value}`}
+      />
     ))}
   </div>
 );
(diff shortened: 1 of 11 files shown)


anthonydillon:
After a quick chat this would be better as a sub-component.

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -0,0 +1,30 @@
...
+
+import "./filter-panel-section.scss";
+
+const FilterPanelSection = ({ data }) => (
+  <div className="filter-panel-section">
+    {data?.heading && (
+      <h3 className="filter-panel-section__heading">{data.heading}</h3>
+    )}
+    {data?.chips?.map((chip, i) => (
+      <Chip lead={chip.lead} value={chip.value} key={i} />
```
hatched, line 13: If the order of the chips in the `data` changes then `key` will become invalid and trigger a re-render. It'll be a bit more performant if the key is tied to the representation of the chip, so maybe...

```suggestion
      <Chip lead={chip.lead} value={chip.value} key={chip.lead+chip.value} />
```

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
```diff
@@ -0,0 +1,27 @@
+import { shallow, mount } from "enzyme";
+import React from "react";
+
+import FilterPanelSection from "./FilterPanelSection";
+
+const sampleData = {
+  heading: "Regions",
+  chips: [{ value: "us-east1" }, { value: "us-east2" }, { value: "us-east3" }],
```
hatched, line 8: Sample data doesn't contain `lead` nor do any of the tests.

src/components/SearchAndFilter/SearchAndFilter.test.js
```diff
@@ -3,6 +3,25 @@ import React from "react";
 
 import SearchAndFilter from "./SearchAndFilter";
 
+const sampleData = [
+  {
+    heading: "Clouds",
+    chips: [{ value: "Google" }, { value: "AWS" }, { value: "Azure" }],
```
hatched, line 9: missing `lead` in this data set as well.

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it("shows correct data passed as prop", () => {
+    const wrapper = mount(<FilterPanelSection data={sampleData} />);
+    expect(wrapper.find(".filter-panel-section").length).toEqual(1);
+    const section = wrapper.find(".filter-panel-section");
+    expect(section.find(".filter-panel-section__heading").text()).toEqual(
+      "Regions"
+    );
+    const firstChip = section.find(".p-chip").at(0);
```
anthonydillon, line 24: Can you find it via the key? Therefore the order will not matter?

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
```diff
@@ -0,0 +1,27 @@
...
+    expect(firstChip.text()).toEqual("us-east1");
```
anthonydillon, line 25: I would expect a count check on the chip elements.

src/components/SearchAndFilter/SearchAndFilter.test.js
anthonydillon: The Chip element itself has tests to cover lead and non-lead so testing that here is probably not needed.

PR #173, opened 2020-06-30
Add an overflow chip counter for filter panel

barrymcgee:
- Pull branch
- Run Storybook
- Resize window so that chips overflow
- Verify counter appears and disappears as expected

http://localhost:53721/?path=/story/searchandfilter--external-state

### DIFF
--- a/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
+++ b/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
@@ -8,12 +8,11 @@ const FilterPanelSection = ({ data }) => {
   const { chips, heading } = data;
   const [overflowCounter, setOverflowCounter] = useState(0);
   const chipWrapper = useRef(null);
-  const viewportWidth = window.innerWidth;
 
   // if the offsetTop is more than double height of a single chip, consider it
   // oveflowing
   const updateFlowCount = function () {
-    const chips = chipWrapper.current.querySelectorAll(".p-chip");
+    const chips = chipWrapper?.current.querySelectorAll(".p-chip");
     let overflowChips = 0;
     chips.forEach((chip) => {
       if (chip.offsetTop > chip.offsetHeight * 2) overflowChips++;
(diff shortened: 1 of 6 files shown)


src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -1,23 +1,56 @@
...
+  const chipWrapper = useRef(null);
+  const viewportWidth = window.innerWidth;
+
+  // if the offsetTop is more than double height of a single chip, consider it
+  // oveflowing
+  const updateFlowCount = function () {
+    const chips = chipWrapper.current.querySelectorAll(".p-chip");
+    let overflowChips = 0;
+    chips.forEach((chip) => {
+      if (chip.offsetTop > chip.offsetHeight * 2) overflowChips++;
```
hatched, line 19: I'm not a fan of the conditional without curlies, it's just asking for issues later (although granted prettier and linters mitigate that) It's just not really worth the 2 character savings :)

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -1,23 +1,56 @@
...
-  </div>
-);
+const FilterPanelSection = ({ data }) => {
+  const { chips, heading } = data;
+  const [overflowCounter, setOverflowCounter] = useState(0);
+  const chipWrapper = useRef(null);
+  const viewportWidth = window.innerWidth;
+
+  // if the offsetTop is more than double height of a single chip, consider it
+  // oveflowing
```
hatched, line 14: This is a smart approach :) nice!

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -1,23 +1,56 @@
...
+    setOverflowCounter(overflowChips);
+  };
+
+  useEffect(() => {
+    updateFlowCount();
+    window.addEventListener("resize", updateFlowCount);
+    return () => {
+      window.removeEventListener("resize", updateFlowCount);
+    };
+  }, [chipWrapper, viewportWidth]);
```
hatched, line 30: You're updating this on viewport width but it's possible that the width of this could change on its parent container resizing not just the viewport...

barrymcgee: Now using resizeObserver to watch size changes in the chips wrapper 👌🏻

PR #174, opened 2020-07-02
Show overflow chips when counter is clicked

barrymcgee:
-Run Storybook
- View http://localhost:49435/iframe.html?id=searchandfilter--external-state
- Decrease screen size so overflow counter is shown
- Click overflow counter
- All chips should be shown
- Refresh page
- Navigate with keyboard, check counter can be activated with keyboard

### DIFF
--- a/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
+++ b/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
@@ -7,6 +7,7 @@ import "./filter-panel-section.scss";
 const FilterPanelSection = ({ data }) => {
   const { chips, heading } = data;
   const [overflowCounter, setOverflowCounter] = useState(0);
+  const [expanded, setExpanded] = useState(false);
   const chipWrapper = useRef(null);
 
   // If the offsetTop is more than double height of a single chip, consider it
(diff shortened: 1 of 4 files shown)


src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -7,12 +7,13 @@ import "./filter-panel-section.scss";
...
+  const [expanded, setExpanded] = useState(false);
   const chipWrapper = useRef(null);
 
   // If the offsetTop is more than double height of a single chip, consider it
   // overflowing
   const updateFlowCount = function () {
-    const chips = chipWrapper?.current.querySelectorAll(".p-chip");
+    const chips = chipWrapper?.current?.querySelectorAll(".p-chip");
     let overflowChips = 0;
     chips.forEach((chip) => {
```
hatched, line 18: Because of the change on 16 there you'll want to add a check that `chips` exists. If 16 short circuits then you'll end up with `undefined.forEach` which will fail.

barrymcgee: Good catch, thanks

barrymcgee:
@hatch Thanks for QA notes - they'll will be swept up in next few PRs in the epic.

PR #181, opened 2020-07-15
Add Stylelint using VF config

barrymcgee:
Linting CI tests should pass


barrymcgee:
@hatched After much chasing, it seems that Storybook failure is a regression in Babel: https://github.com/storybookjs/storybook/issues/11439#issuecomment-654847250

hatched:
Looks good, everything seems to work as expected now.

PR #187, opened 2020-07-21
Add chips to search bar

barrymcgee:
- Run Storybook
- http://localhost:53092/iframe.html?id=searchandfilter--external-state&viewMode=story
- Click chips in dropdown panel
- Dismiss chips from search box
- Click enough chips to overflow

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1308,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1330

### DIFF
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -3,13 +3,15 @@ import PropTypes from "prop-types";
 
 import "./Chip.scss";
 
-const Chip = ({ value, lead = "", onDismiss }) => {
+const Chip = ({ value, lead = "", onDismiss, selected }) => {
   return (
-    <div className="p-chip">
-      {lead ? `${lead.toUpperCase()}: ` : null}
-      {value}
+    <div className="p-chip" aria-pressed={selected} role="button">
+      <span className="p-chip__lead">
+        {lead ? `${lead.toUpperCase()}: ` : null}
+      </span>
+      <span className="p-chip__value">{value}</span>
       {onDismiss ? (
-        <button className="p-chip__dismiss" onClick={onDismiss()}>
+        <button className="p-chip__dismiss" onClick={() => onDismiss()}>
           <i className="p-icon--close">Dismiss</i>
         </button>
       ) : null}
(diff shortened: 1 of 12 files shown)


src/components/Chip/Chip.js
```diff
@@ -3,13 +3,15 @@ import PropTypes from "prop-types";
 
 import "./Chip.scss";
 
-const Chip = ({ value, lead = "", onDismiss }) => {
+const Chip = ({ value, lead = "", onClick, onDismiss, selected }) => {
   return (
-    <div className="p-chip">
-      {lead ? `${lead.toUpperCase()}: ` : null}
-      {value}
+    <div className="p-chip" data-selected={selected}>
```
anthonydillon, line 8: I would suggest we use `aria-pressed` instead of a data attr. https://gomakethings.com/button-state-and-accessibility/

anthonydillon: But maybe that should not be used on a div in this case.

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -54,32 +60,107 @@ const SearchAndFilter = ({
...
+          externallyControlled={externallyControlled}
+          placeholder={searchData.length ? "Add filter" : "Search and filter"}
+          onChange={(searchTerm) => searchOnChange(searchTerm)}
+          onClick={() => setFilterPanelHidden(false)}
+          onFocus={() => setFilterPanelHidden(false)}
+          onSubmit={(e) => handleSubmit(e)}
+          value={searchTerm}
+        />
+        {overflowSearchTermCounter > 0 && (
+          <span
```
anthonydillon, line 140: Can you make this a button, please? This is currently not accessible

src/components/SearchAndFilter/SearchAndFilter.scss
```diff
@@ -25,4 +84,12 @@
       pointer-events: none;
     }
   }
+
+  &__selected-count {
+    color: $color-information;
+    cursor: pointer;
```
anthonydillon, line 90: You wont need this if its a button

src/components/Chip/Chip.js
barrymcgee: This is cool.. didn't know about it. It can also be used on a div by also adding `role="button"`

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: Similar here, I'll add a `role="button"` which will make it accessible without requiring a ream of styling overrides

PR #191, opened 2020-07-24
Migrate Accordion and AccordionSection to TypeScript

squidsoup:
## Done
* Migrate `Accordion` and `AccordionSection` to TypeScript.
* Add base types in `types`.

## QA
* Ensure storybook renders correctly.
* Build and pack, `yarn build; npm pack`, install tarball with `yarn add` in `maas-ui` to ensure `maas-ui` builds.


src/components/Accordion/Accordion.tsx
```diff
@@ -1,30 +1,48 @@
...
-      sectionKey={key}
+      sectionKey={key?.toString()}
       setExpanded={setExpanded}
       titleElement={titleElement}
       {...props}
     />
   ));
 
+type Props = {
+  sections: typeof AccordionSection[];
```
huwshimi, line 29: So I guess this would become `AccordionSectionProps & { key?: string | number }[]`.

I think typeof gives the type of the component, not the props:

Screen Shot 2020-07-28 at 3 29 04 pm

huwshimi:
Looks good now @squidsoup, thanks for those fixes.

PR #197, opened 2020-07-31
Allow a node to be passed to field help

huwshimi:
## Done:
- Update Field to support a React node to be supplied for the help prop.
- Update components that pass help to Field.

QA: None, tests should pass.

Fixes: https://github.com/canonical-web-and-design/react-components/issues/195.
Fixes: https://github.com/canonical-web-and-design/react-components/issues/67.

### DIFF
--- a/src/components/Field/Field.js
+++ b/src/components/Field/Field.js
@@ -21,12 +21,12 @@ const generateError = (error, caution, success) => {
   );
 };
 
-const generateLabel = (forId, required, label, stacked) => {
+const generateLabel = (forId, required, label, labelClassName, stacked) => {
   if (!label) {
     return;
   }
   const labelNode = (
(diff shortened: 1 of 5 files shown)


Caleb-Ellis:
@huwshimi do you think it'd be a good idea to change the validation props (`caution`, `error`, `success`) as well? It's definitely less likely to be anything other than a string, but I could see links or `<strong>`s or something being in there.

huwshimi:
@Caleb-Ellis yeah good idea, I'll do that.

huwshimi:
@Caleb-Ellis Done. I also noticed some wrapper class props didn't exist on all field types and while I was at it also passed the label class (https://github.com/canonical-web-and-design/react-components/issues/67).

PR #198, opened 2020-07-31
Build Slider component.

Caleb-Ellis:
## Done

- Migrated Slider component from maas-ui
- Added stories for default + with number input

## QA

- Run `yarn docs`
- Check that there is a Slider component that you can interact with.


squidsoup:
It appears I can't actually slide the slider component in storybook 😅

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,85 @@
+import React from "react";
+import type { ChangeEventHandler, HTMLProps, ReactNode } from "react";
+
+import Field from "../Field";
+
+export const FILLED_COLOR = "#0066CC";
+
+type Props = {
```
huwshimi, line 8: I think we probably still want to have prop-types as well for those not using typescript, as well as giving as much help as possible to people consuming the library.

squidsoup: > I think we probably still want to have prop-types as well for those not using typescript, as well as giving as much help as possible to people consuming the library.

Once we've moved off CRA, we should look at generating proptype from types using [babel-plugin-typescript-to-proptypes](https://github.com/milesj/babel-plugin-typescript-to-proptypes).

Caleb-Ellis: :+1:

PR #200, opened 2020-08-04
Add search with pattern matching to search and filter

barrymcgee:
- Pull branch
- Run Storybook
- Click http://localhost:56916/iframe.html?id=searchandfilter--external-state&viewMode=story
- Search for known and unknown chips

### DIFF
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -11,6 +11,7 @@ const Chip = ({
   onDismiss,
   selected,
   subString = "",
+  quoteValue,
 }) => {
   // When user tabs over chip and presses Enter or Space key, chip will trigger
   // click functionality
@@ -23,6 +24,7 @@ const Chip = ({
     }
   };
 
+  const chipValue = highlightSubString(value, subString).text;
   return (
     <div
       className="p-chip"
@@ -36,7 +38,7 @@ const Chip = ({
       <span
         className="p-chip__value"
         dangerouslySetInnerHTML={{
-          __html: highlightSubString(value, subString).text,
+          __html: quoteValue ? `'${chipValue}'` : chipValue,
         }}
       />
 
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -29,7 +29,7 @@ const Chip = ({
       className="p-chip"
       aria-pressed={selected}
       role="button"
-      tabIndex="0"
+      tabIndex={!onDismiss ? "0" : null}
       onClick={onClick}
       onKeyPress={(e) => onKeyPress(e)}
     >
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -56,6 +56,9 @@ Chip.propTypes = {
   value: PropTypes.string,
   lead: PropTypes.string,
   onDismiss: PropTypes.func,
+  onClick: PropTypes.func,
+  selected: PropTypes.bool,
+  subString: PropTypes.string,
 };
 
 export default Chip;
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -39,7 +39,7 @@ const Chip = ({
       <span
         className="p-chip__value"
         dangerouslySetInnerHTML={{
-          __html: hightlightSubString(value, subString),
+          __html: hightlightSubString(value, subString).text,
         }}
       />
 
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -14,10 +14,9 @@ const Chip = ({
 }) => {
   // When user tabs over chip and presses Enter or Space key, chip will trigger
   // click functionality
-  const onKeyPress = (e) => {
-    // charCode 32 = Space key
-    // charCode 13 = Enter key
-    if (e.charCode === 32 || e.charCode === 13) {
+  const onKeyDown = (e) => {
+    // The " " value is what is returned for the spacebar
+    if (e.key === " " || e.key === "Enter") {
       if (onClick instanceof Function) {
         onClick({ lead: lead, cloud: value });
       }
@@ -31,7 +30,7 @@ const Chip = ({
       role="button"
       tabIndex={!onDismiss ? "0" : null}
       onClick={onClick}
-      onKeyPress={(e) => onKeyPress(e)}
+      onKeyDown={(e) => onKeyDown(e)}
     >
       <span className="p-chip__lead">
         {lead ? `${lead.toUpperCase()}: ` : null}
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { hightlightSubString } from "../../utils";
+import { highlightSubString } from "../../utils";
 
 import "./Chip.scss";
 
@@ -38,7 +38,7 @@ const Chip = ({
       <span
         className="p-chip__value"
         dangerouslySetInnerHTML={{
-          __html: hightlightSubString(value, subString).text,
+          __html: highlightSubString(value, subString).text,
         }}
       />
 
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -1,15 +1,22 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { hightlightSubString } from "../../utils";
 
 import "./Chip.scss";
 
-const Chip = ({ value, lead = "", onDismiss, selected }) => {
+const Chip = ({ value, lead = "", onDismiss, selected, subString = "" }) => {
   return (
     <div className="p-chip" aria-pressed={selected} role="button">
       <span className="p-chip__lead">
         {lead ? `${lead.toUpperCase()}: ` : null}
       </span>
-      <span className="p-chip__value">{value}</span>
+      <span
+        className="p-chip__value"
+        dangerouslySetInnerHTML={{
+          __html: hightlightSubString(value, subString),
+        }}
+      />
+
       {onDismiss ? (
         <button className="p-chip__dismiss" onClick={() => onDismiss()}>
           <i className="p-icon--close">Dismiss</i>
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -32,9 +32,7 @@ const Chip = ({
       onClick={onClick}
       onKeyDown={(e) => onKeyDown(e)}
     >
-      <span className="p-chip__lead">
-        {lead ? `${lead.toUpperCase()}: ` : null}
-      </span>
+      {lead && <span className="p-chip__lead">{lead.toUpperCase()}</span>}
       <span
         className="p-chip__value"
         dangerouslySetInnerHTML={{
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -4,9 +4,35 @@ import { hightlightSubString } from "../../utils";
 
 import "./Chip.scss";
 
-const Chip = ({ value, lead = "", onDismiss, selected, subString = "" }) => {
+const Chip = ({
+  value,
+  lead = "",
+  onClick,
+  onDismiss,
+  selected,
+  subString = "",
+}) => {
+  // When user tabs over chip and presses Enter or Space key, chip will trigger
+  // click functionality
+  const onKeyPress = (e) => {
+    // charCode 32 = Space key
+    // charCode 13 = Enter key
+    if (e.charCode === 32 || e.charCode === 13) {
+      if (onClick instanceof Function) {
+        onClick({ lead: lead, cloud: value });
+      }
+    }
+  };
+
   return (
-    <div className="p-chip" aria-pressed={selected} role="button">
+    <div
+      className="p-chip"
+      aria-pressed={selected}
+      role="button"
+      tabIndex="0"
+      onClick={onClick}
+      onKeyPress={(e) => onKeyPress(e)}
+    >
       <span className="p-chip__lead">
         {lead ? `${lead.toUpperCase()}: ` : null}
       </span>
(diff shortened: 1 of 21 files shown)


squidsoup:
Hey @barrymcgee, it would be great if you would consider migrating this component to TypeScript while you're working on it. Happy to help if you need a hand.

barrymcgee:
@squidsoup Sure, I'll give it a bash. I'd prefer to land this and do as a follow-up though.

ziheliu214:
When typing in a character, in this example 'f':
1) the unrelated chips were shown too 'bar' 'baz'
image
2) after selecting one, the 'f' should disappear from the search field with placeholder text 'add filter' appears instead
image

Thanks!

ziheliu214:
Other comments (sorry if you haven't accomplished these but I'm commenting too far ahead!)

1) Clear all [X] button is needed as long as there is any content on the search field.
2) Click on the selected item from the drop-down again should de-select the item - the chip added in the field should be cleared.

hatched:
Couple QA issues on Firefox Android.

- [ ] The text isn't vertically centered in the pill 
- [ ] The hint text gets cut off when you have a couple pills
- [ ] Part of the hidden row is shown
- [ ] Clicking the +6 doesn't do anything

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -23,50 +29,92 @@ const FilterPanelSection = ({ data, addToSelected, searchData }) => {
     setOverflowCounter(overflowCount);
   };
 
+  // Check if search term characters matches any characters in panel heading
+  const searchTermInHeading = hightlightSubString(heading, searchTerm).includes(
+    "<strong>"
```
anthonydillon, line 34: This tightly couples the expected result of `hightlightSubString`. I think a simple `string.contains` or common function used by both places would be better here as you might change `strong` to `<i>` in the future and this would result in unexpected failures here and other places that use it.

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -23,50 +29,92 @@ const FilterPanelSection = ({ data, addToSelected, searchData }) => {
...
+  // Serialise chip values into string so it can be interrogated with subString
+  let chipValues = [];
+  Object.entries(chips).forEach((chipValue) => {
+    chipValues.push(chipValue[1].value);
+  });
+  // Search chips for character match with search term
+  const searchTermInChips = hightlightSubString(
+    chipValues.toString(),
+    searchTerm
+  ).includes("<strong>");
```
anthonydillon, line 46: Same as above

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -23,50 +29,92 @@ const FilterPanelSection = ({ data, addToSelected, searchData }) => {
...
-    </div>
+          )}
+          <div className="filter-panel-section__chips" ref={chipWrapper}>
+            {chips?.map((chip) => {
+              // If search term has been added to input, only matching chips
+              // should display
+              const searchTermInChip = hightlightSubString(
+                chip.value,
+                searchTerm
+              ).includes("<strong>");
```
anthonydillon, line 88: Same here

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -108,6 +131,13 @@ const SearchAndFilter = ({
     }
   }, [searchData]);
 
+  // Add search prompt value to search on Enter key (code: 32) press
+  const searchPromptKeyPress = (e) => {
+    if (e.charCode === 32) {
```
anthonydillon, line 136: Can you use `e.key === Enter`? Saves comment explaining the codes above.

src/utils.js
```diff
@@ -1 +1,17 @@
 export const IS_DEV = process.env.NODE_ENV === "development";
+
+/**
+ * Find substring and wrap in <strong /> tag
+ * @param {string} str - The string to search
+ * @param {string} subString - The substring to find
+ * @param {Obj} newStr - HTML fragment containing match wrapped in strong tag
```
anthonydillon, line 7: This should be docstringed as the return.

anthonydillon (dismissed):
Looks ggreat @barrymcgee. Few suggestions in the code.

anthonydillon:
## QA
- The [designs](https://app.zeplin.io/project/5ed19809e42d2c9b901fcd9c/screen/5ed4ed28f2b8f36fccfc5147) show that if a search matches part of the filter panel section title it should highlight that title (as the demo does) but also all chips in that section should be visible as the user will type cloud to see all clouds:
image

- Long chips overflow and have no way of closing (not a blocker)
image

- Having no space for the input is a critical issue:
image

- Jeff mentioned it but clicking the overflow on the panel does nothing

- I think we should add the focus styling to the input and help the user (not a blocker)
image

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: I did try this but originally `e.key` here returns an empty string and `e.keyCode` returns `0`? 🧐

src/components/Chip/Chip.js
```diff
@@ -1,15 +1,48 @@
...
+      <span
+        className="p-chip__value"
+        dangerouslySetInnerHTML={{
+          __html: hightlightSubString(value, subString).text,
+        }}
+      />
+
       {onDismiss ? (
         <button className="p-chip__dismiss" onClick={() => onDismiss()}>
           <i className="p-icon--close">Dismiss</i>
```
squidsoup, line 48: Please update `Chip`s propTypes.

src/components/Chip/Chip.js
```diff
@@ -1,15 +1,48 @@
...
+  onDismiss,
+  selected,
+  subString = "",
+}) => {
+  // When user tabs over chip and presses Enter or Space key, chip will trigger
+  // click functionality
+  const onKeyPress = (e) => {
+    // charCode 32 = Space key
+    // charCode 13 = Enter key
+    if (e.charCode === 32 || e.charCode === 13) {
```
squidsoup, line 20: Might be nice to defined these as consts, then you don't need the comments:

`if (e.charCode === SPACE_KEY || e.charCode === ENTER_KEY)`

is more readable.

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -23,50 +29,94 @@ const FilterPanelSection = ({ data, addToSelected, searchData }) => {
...
+              >
+                +{overflowCounter}
+              </span>
+            )}
+          </div>
+        </div>
+      )}
+    </>
   );
 };
```
squidsoup, line 122: Please update `FilterPanelSection`s propTypes.

src/components/Chip/Chip.js
```diff
@@ -1,15 +1,48 @@
...
+      if (onClick instanceof Function) {
```
squidsoup, line 21: `if (onClick)` is probably more idiomatic, any reason you need this?

src/utils.js
```diff
@@ -1 +1,20 @@
...
+export const hightlightSubString = (str, subString) => {
+  const caseInsensitiveRegex = new RegExp(subString, "gi");
+  const newStr = str.replace(
+    caseInsensitiveRegex,
+    (match) => `<strong>${match}</strong>`
+  );
+
+  return {
+    text: subString === "" ? str : newStr,
+    match: newStr === str ? false : true,
```
squidsoup, line 18: This is redundant, `newStr === str` will return a boolean value.

src/utils.js
```diff
@@ -1 +1,20 @@
 export const IS_DEV = process.env.NODE_ENV === "development";
+
+/**
+ * Find substring and wrap in <strong /> tag
+ * @param {string} str - The string to search
+ * @param {string} subString - The substring to find
+ * @return {Obj} newStr - Object with text and match bool
+ */
+export const hightlightSubString = (str, subString) => {
```
squidsoup, line 9: This could use some tests

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -153,13 +184,25 @@ const SearchAndFilter = ({
           aria-hidden={filterPanelHidden}
         >
           <ContextualMenu>
+            {searchTerm.length > 0 && (
+              <div
+                className="search-prompt"
+                onClick={() => handleSubmit()}
+                onKeyPress={(e) => searchPromptKeyPress(e)}
+                role="button"
+                tabIndex="0"
```
squidsoup, line 193: `tabIndex` is a number, not a string.

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -24,9 +20,6 @@ const SearchAndFilter = ({
```
squidsoup, line 20: Not in the diff, but if you could fix it please, `useRef` should never be called without an initial value. In both cases, you probably want `useRef(null)` here.

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -153,13 +184,25 @@ const SearchAndFilter = ({
           aria-hidden={filterPanelHidden}
         >
           <ContextualMenu>
+            {searchTerm.length > 0 && (
+              <div
+                className="search-prompt"
+                onClick={() => handleSubmit()}
+                onKeyPress={(e) => searchPromptKeyPress(e)}
```
squidsoup, line 191: [onKeyPress is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onkeypress)

src/components/SearchAndFilter/SearchAndFilter.js
squidsoup: Probably because the event is deprecated (see below).

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -23,50 +29,94 @@ const FilterPanelSection = ({ data, addToSelected, searchData }) => {
...
+  // Serialise chip values into string so it can be interrogated with subString
+  let chipValues = [];
+  Object.entries(chips).forEach((chipValue) => {
+    chipValues.push(chipValue[1].value);
+  });
+  // Search chips for character match with search term
+  const searchTermInChips = hightlightSubString(
+    chipValues.toString(),
+    searchTerm
+  ).match;
```
squidsoup, line 44: This seems a bit convoluted (`hightlightSubstring` has two jobs), why not `chipValues.includes(searchTerm)`?

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
```diff
@@ -23,50 +29,94 @@ const FilterPanelSection = ({ data, addToSelected, searchData }) => {
...
+                searchTerm
+              ).match;
+              const chipVisible = searchTermInChip || searchTerm === "";
+              return (
+                <span key={`${chip.lead}+${chip.value}`}>
+                  {chipVisible && (
+                    <Chip
+                      lead={chip.lead}
+                      value={chip.value}
+                      selected={searchData?.includes(chip) ? true : false}
```
squidsoup, line 98: `includes` returns a boolean, you don't need the ternary.

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: This is a html element?

src/components/Chip/Chip.js
barrymcgee: Fixed this by using `onKeyDown` event.. thanks.

src/components/Chip/Chip.js
barrymcgee: This was just to prevent a fatal error where someone might pass the onClick as an empty string or object.

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.js
barrymcgee: Because the search is fuzzy, array.includes must be an exact match to return true.

hatched:
QA Notes:

- [ ] Input is very large but clickable area is very small:
Screen Shot 2020-09-07 at 8 38 33 AM

- [ ] Are the pills supposed to have two colons?
Screen Shot 2020-09-07 at 8 40 17 AM

- [ ] It doesn't collapse down into a single row if the input is the only thing on the next line.
Screen Shot 2020-09-07 at 8 43 00 AM

- [ ] The count shows up but it doesn't collapse to a single row when clicking away.
Screen Shot 2020-09-07 at 8 47 37 AM

- [ ] When the input goes to two lines it expands its container, this won't be feasible in the applications as it'll push the whole header down.

- [ ] Should search values not have a "search" prefix?
Screen Shot 2020-09-07 at 8 46 44 AM

barrymcgee:
> Should search values not have a "search" prefix?

Not according to the designs.

ziheliu214:
Screenshot 2020-09-11 at 16 31 11
Were trying to use '_search keyword_' to represent its a string search @barrymcgee @hatched Happy to hear your ideas

barrymcgee:
> When the input goes to two lines it expands its container, this won't be feasible in the applications as it'll push the whole header down.

It's designed to be positioned absolutely.

Screenshot 2020-09-11 at 18 03 46

hatched:
Thanks for keeping up on this one!

QA Notes:
- [ ] Selecting a pill when it's expanded to multiple rows collapses the row
Screen Shot 2020-09-11 at 1 08 29 PM
After clicking the pill it collapses.
Screen Shot 2020-09-11 at 1 08 36 PM
Clicking on the input closes the drop down and re-expands the input so you can see the selected pill.
Screen Shot 2020-09-11 at 1 08 50 PM
- [ ] When the input has focus all the pills move up a couple pixels
Screen Shot 2020-09-11 at 1 09 12 PM
Screen Shot 2020-09-11 at 1 09 21 PM

hatched:
Thanks for keeping up on this one @barrymcgee 
As it's really close lets land this as-is as a beta implementation and address the remaining qa issues as follow-ups:
- https://github.com/canonical-web-and-design/react-components/issues/277
- https://github.com/canonical-web-and-design/react-components/issues/278
- https://github.com/canonical-web-and-design/react-components/issues/279

hatched (approved):

PR #204, opened 2020-08-05
refactor: Migrate pagination to typescript.

squidsoup:
Migrate Pagination, PaginationItem, PaginationButton to TypeScript.


src/components/PaginationItem/PaginationItem.tsx
```diff
@@ -1,8 +1,18 @@
 import classNames from "classnames";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { MouseEventHandler } from "react";
```
huwshimi, line 3: Should this be `import type { MouseEventHandler } from "react";` like above?

squidsoup: It should, thanks @huwshimi!

PR #243, opened 2020-08-14
Theme storybook

huwshimi:
## Done

- Improve theme to look more like Vanilla. Storybook only supports very minimal theming so I've changed everything I could to use Vanilla colours.
- Upgrade Storybook.
- Replace Knobs with Controls.

### QA steps

- Have a look around the demo, make sure things look a bit like Vanilla and nothing is broken.

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/6.
Fixes: https://github.com/canonical-web-and-design/react-components/issues/5.


huwshimi:
> Just one thing with the Slider controls - it seems to crash if I change any of the form help text/validation props (the ones that accept a ReactNode). Other than that though, LGTM!

Thanks, fixed.

PR #249, opened 2020-08-18
Only install Cypress binary for the GH action

huwshimi:
## Done

- Only install the Cypress binary for the GitHub action.

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/88.


squidsoup (approved):
Nice!

PR #260, opened 2020-09-04
Handle adjusting the tooltip so that it remains on screen

huwshimi:
## Done
- Added useFitsScreen, useListener, usePrevious and useThrottle hooks.
- Handle adjusting the tooltip so that it remains on screen.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Navigate to the Tooltip examples.
- Check that the follow mouse example follows the mouse.
- Resize your window to different sizes. The tooltip should adjust so that it remains on the screen (except where it can't).


src/components/Tooltip/Tooltip.tsx
```diff
@@ -30,18 +33,46 @@ type Position =
...
+   */
   message?: string;
+  /**
+   * Position of the tooltip relative to the element.
+   */
   position?: Position;
+  /**
+   * An optional class to apply to the tooltip message element.
+   */
+  tooltipClassName?: string;
```
hatched, line 64: Is this strictly necessary with the `className` prop above? As it should still be able to be targeted from that class?

I'm just seeing a lot of prop creep here so thinking of reasonable ways to shrink the api surface area.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -30,18 +33,46 @@ type Position =
   | "top-left"
   | "top-right";
 
-type Props = {
+export type Props = {
+  /**
+   * Whether the tooltip should adjust to fit in the screen.
+   */
+  autoAdjust?: boolean;
```
hatched, line 40: Instead of having this prop what about defaulting to auto adjusting if the `position` prop is not defined?

src/hooks/useFitsScreen.ts
```diff
@@ -0,0 +1,183 @@
...
+        const referenceBottom = referenceTop + referenceHeight;
+        const referenceRight = referenceLeft + referenceWidth;
+        const referenceCenterX = referenceLeft + referenceWidth / 2;
+        const referenceCenterY = referenceTop + referenceHeight / 2;
+        const screenLeft = htmlRef.current.scrollLeft;
+        const screenRight = window.innerWidth - htmlRef.current.scrollLeft;
+        const screenTop = 0;
+        const screenBottom = window.innerHeight;
+
+        callback({
```
hatched, line 131: There is a lot of repeated math here, can they be abstracted out into variables to make this easier to read (and compress better)?

src/hooks/useFitsScreen.ts
```diff
@@ -0,0 +1,183 @@
...
+ * @param targetNode The element to try and fit on the screen.
+ * @param referenceNode The element to use to position the target.
+ * @param callback The function to call when updating fitment info.
+ * @param spacer An additional space to leave between the target and reference.
+ * @param shouldCheck Whether the fitment info should be being checked.
+ * @param fromMouse Whether the target should be being positioned in relation
+ *                  to the mouse. In this case refernceNode will be used to
+ *                  listen for mouseover events.
+ */
+export const useFitsScreen = (
```
hatched, line 80: This name implies a boolean return value.... maybe `useFitsWhereWithin` 🤷‍♂️

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -35,3 +35,18 @@ An alternative use of tooltips is to provide information on a disabled actionabl
     {Template.bind({})}
   </Story>
 </Canvas>
+
+### Follow mouse
```
hatched, line 39: Can we add some examples of a constrained tooltip as this can't be easily QA'd without it?

hatched (changes requested):
Lots of good work here, thanks @huwshimi.

src/components/Tooltip/Tooltip.tsx
huwshimi: The actual tooltip gets rendered inside a portal which hoists it to the top level of the DOM (body) so it no longer exists inside the wrapping element. This class has been added so you can still target the message if needed.

src/components/Tooltip/Tooltip.tsx
huwshimi: I think you still want to be able to provide a default position for when it does fit on the screen (e.g. by default you might want it to appear above a button so that it's not covering other content).

src/hooks/useFitsScreen.ts
huwshimi: How about `useWindowFitment`?

hatched: Yeah that sounds good!

src/components/Tooltip/Tooltip.stories.mdx
huwshimi: I'm not quite sure what you mean sorry. You can resize your window to see this adjust, but I'm not quite sure what you mean by "constrained"...

src/hooks/useFitsScreen.ts
huwshimi: Done, every repeated calculation is now a variable.

src/components/Tooltip/Tooltip.stories.mdx
huwshimi: (Followup from convo) this only constrains by window edges, not in relation to other elements on the screen.

hatched (approved):

PR #263, opened 2020-09-04
Initial build of ModularTable component

bartaz:
## Done

Added new `ModularTable` component that is a first basic implementation of the Vanilla modular table.
This is a MVP implementation of static table rendered based on `columns` and `data` objects.
It uses [React Table](https://react-table.tanstack.com/) internally for table functionality and data manipulation.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check ModularTable in Story book
- Make sure it renders correctly and is properly documented


src/components/ModularTable/ModularTable.tsx
```diff
@@ -0,0 +1,65 @@
...
+// TODO:
+// import PropTypes from "prop-types";
+import { useTable } from "react-table";
+
+import Table from "../Table";
+import TableRow from "../TableRow";
+import TableHeader from "../TableHeader";
+import TableCell from "../TableCell";
+
+function ModularTable({ data, columns }) {
```
bartaz, line 11: Any suggestions for a better name are welcome.

This is meant to be a default "Table" component. `Table` (and `TableCell`, etc) is already taken by basic presentational components, `MainTable` is also taken as current "full featured" table.
It's going to be some time until "modular table" replaces all functionality of current `MainTable`, but I think at some point `MainTable` will become deprecated when new component provides enough features.

So, long story short - any suggestions for a name better than `ModularTable`?
`TableComponent`, `TableController`, `ReactTable` (this ties it a bit to internal react-table implementation, but this may be more confusing than helpful)...

hatched: `ModularTable` sounds fine to me :) 

In the future when its to replace `MainTable` We can discuss if it can take over that namespace with a major version bump.

Caleb-Ellis: +1 to `ModularTable`. I think we're all on the same page with what's meant by that.

src/react-table-config.d.ts
```diff
@@ -0,0 +1,128 @@
+// check @types/react-table README for details on this config file:
```
squidsoup, line 1: @bartaz is there a reason we can't just use `yarn install @types/react-table`? This seems a little unusual.

src/components/ModularTable/ModularTable.tsx
bartaz: `ModularTable` it is then.

src/react-table-config.d.ts
bartaz: @squidsoup we do, this config is 'part' of react-table types (it imports and extends types from @types/react-table).

I agree it's really strange and confusing (especially when I don't really know TS very well).
The react-table API (the shape of `data` and `columns` input objects) is very extensible. Depending on the features you use, you will add additional fields to the data and columns objects. So, to my understanding, @types/react-table defines the shape of these as whatever the defaults are in react-table, but you can use this config to turn on/off any features you use to make sure you only allow props that you need.

In our case, I want to pass `className` as a column prop. This is not something that is in react-table default props, so the native types don't support it and TS complains when sees `className` used.
So I used this config to add `className` to the `ColumnInstance` interface.

I don't know if that's the best way to do it, or can we somehow to these custom props somewhere else in the component itself?

The way this config file works is mystery to me. It seems to import a lot of things that are not used in the file, but if I don't import, the react-table module stops working, so I'm kind of trying to figure out what is needed here, what is not.

Personally I agree that this is quite confusing and unusual and would like to avoid using this file if we can find another way to extend types from @types/react-table.

bartaz: Managed to clean up this config to a reasonable minimum.

bartaz:
@squidsoup @huwshimi I get emails from Percy about build failed, but it's not visible here in PR. How is Percy set up? It's not a check on PRs that has to pass?

bartaz:
@squidsoup @huwshimi I feel like this is close to getting ready for review/merging. Can you have a look at what could be missing? Should I be adding more comments in the code for documentation? What would be the minimal amount of tests needed?

src/components/ModularTable/ModularTable.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+### Props
+
+<Props of={ModularTable} />
+
+### Default
+
+<Preview>
+  <Story name="Default">
+    <ModularTable
+      columns={React.useMemo(
```
hatched, line 19: Is the `useMemo` strictly necessary when working with the `react-table`?

- If it is, then I'd prefer it to be abstracted away into the table so we can just provide it with data from immer/redux.
- If it isn't then can you remove it :D

src/components/ModularTable/ModularTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+              >
+                {column.render("Header")}
+              </TableHeader>
+            ))}
+          </TableRow>
+        ))}
+      </thead>
+      <tbody {...getTableBodyProps()}>
+        {rows.map((row) => {
+          prepareRow(row);
```
hatched, line 49: This is interesting, I had to look up why it was necessary (they have great docs btw...)

Could you add a comment here just to say what this is and that this will need to be updated if we're doing pagination?

hatched (approved):
Thanks @bartaz this is looking like a great start.

src/components/ModularTable/ModularTable.tsx
bartaz: Sure, will do.

src/components/ModularTable/ModularTable.stories.mdx
bartaz: It is necessary. I have no idea how they can detect it or why it has any difference (other than simply performance of re-renders), but react-table will not render if data is not memoized.
They clearly state in the API docs which params need to be memoized (https://react-table.tanstack.com/docs/api/useTable).

But I'm not sure if this should be abstracted and done internally by the component itself. Component will not know how this data is build and when should it be refreshed. I'm also not sure what would happen if objects are memoized twice - once by parent component, and second time 'abstracted' inside the table.

bartaz: Just to be sure I verified again, and for some reason now I can't make it fail if I don't memoize the data or columns.
So it seems that not memoizing may not 'break' react-table.
But still, their docs clearly state these arguments as requiring memoization as without it on each re-render table would recalculate a lot of logic. So I think we should keep memoization as requirement.

But I still think memoization should be done wherever the data is prepared for the table (because this component will know for example when to recalculate it (second param to `useMemo` hook)).

hatched: Ahhh ok, It probably uses a ==='s internally to check if it's the same pointer.

Because of this requirement I think that we need to be very clear in our documentation that it's based off of `react-table` and that this value needs to be memoized as that's not a very common problem (especially if it doesn't output an error in the console.)

huwshimi:
> @squidsoup @huwshimi I get emails from Percy about build failed, but it's not visible here in PR. How is Percy set up? It's not a check on PRs that has to pass?

It looks like it must only run on the main repo. You can see it in the checks for this PR: https://github.com/canonical-web-and-design/react-components/pull/273

squidsoup (approved):
Looks good @bartaz, particularly now with the types tidied. Looking forward to using this!

PR #266, opened 2020-09-06
refactor: migrate table components to typescript

squidsoup:
## Done
Migrate TableCell, TableHeader and TableRow to typescript.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Ensure components render as expected in storybook.

## Fixes

Fixes: #230 #231 #232 #233


src/components/Table/Table.tsx
```diff
@@ -1,15 +1,23 @@
...
 
 const Table = ({
   children,
   className,
-  expanding,
-  responsive,
-  sortable,
+  expanding = false,
+  responsive = false,
+  sortable = false,
```
hatched, line 18: With these last three being listed as optional we should probably also document their defaults as the other two do not have a default value.

src/components/TableCell/TableCell.tsx
```diff
@@ -1,6 +1,14 @@
...
 import PropTypes from "prop-types";
-import React from "react";
+import React, { HTMLProps, ReactNode } from "react";
+
+type Props = {
+  children?: ReactNode;
+  className?: string;
+  expanding?: boolean;
+  hidden?: boolean;
+  role?: string;
```
hatched, line 10: Is there an enum for roles in this context?

Also, see comment above about documenting default value for optional prop.

src/components/TableHeader/TableHeader.tsx
```diff
@@ -0,0 +1,25 @@
+import PropTypes from "prop-types";
+import React, { HTMLProps, ReactNode } from "react";
+
+type Props = {
+  children: ReactNode;
+  sort?: "none" | "ascending" | "descending";
```
hatched, line 6: See above about documenting default prop value.

hatched (approved):
Looks good with a few small requests.

src/components/Table/Table.tsx
squidsoup: @hatched thanks for the review! when you say "document their default", can you explain what you had in mind? We don't use jsdoc in this project, and these components don't actually have storybooks (although arguably should).

hatched: Yeah I didn't mean jsdoc specifically, I was more thinking about typedoc. But yeah if you could add a docblock to that effect that would be great.

src/components/TableCell/TableCell.tsx
squidsoup: I don't think there is in any context, react defines `role` in `HTMLAttributes` as `role?: string`.

src/components/Table/Table.tsx
squidsoup: @hatched I'm going to hold off on adding jsdoc here, as there are couple of different approaches I think we should discuss in dev catchup first:

1. add a js/tsdoc block to the component
2. add jsdoc blocks directly to the Props type, setting defaults with `@default`, and generating proptypes from these (material-ui does this)

bartaz:
@squidsoup I talked about the `role`s and accessibility with @lyubomir-popov.
What we found out is that the roles we have in Vanilla examples are incorrect (for the types of tables that we build).
Role `grid` and other from that family are used when creating spreadsheet-like interactive grids that you can navigate with keyboard. Role `table` is used for tables, but it's not needed when using HTML table elements (only when building virtual table with divs).

So in out case `role` attribute is completely not needed on table elements at the moment and should not be set to any value by default.

Before we get to the point when we really need it for something (we build virtual table or real grid) I would remove it from components.

huwshimi:
@bartaz Vanilla uses `role` to target elements in a few cases. At the very least we'll need to add the role to table header components when sortable. See:

https://github.com/canonical-web-and-design/vanilla-framework/blob/7f482e77822249a4dc9d6fce0f7667e264702a31/scss/_patterns_table-sortable.scss#L24

bartaz:
@huwshimi Ah, yes, I forgot about that. We need to fix it in Vanilla first then.

Anyway, for this PR I wouldn't worry about typing `role` too much.

PR #275, opened 2020-09-15
Migrate the contextual menu from maas-ui

huwshimi:
## Done

- Migrate the button component to TypeScript.
- Migrate the contextual menu from maas-ui.
- Add auto window adjustment.
- Migrate to TypeScript.
- Update SearchAndFilter to use the modified contextual menu.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open storybook and check that the contextual menu displays.
- Adjust the window size and see that the menu adjust to fit (you'll need to make your window very narrow).
- Check that the external state SearchAndFilter demo still works (/?path=/story/searchandfilter--external-state).

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/215.
Fixes: https://github.com/canonical-web-and-design/react-components/issues/149.


src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -239,31 +207,48 @@ const SearchAndFilter = ({
...
+            visible={!filterPanelHidden}
+            dropdownClassName="search-and-filter__dropdown"
+          >
+            <CSSTransition
+              appear={true}
+              in={true}
+              timeout={200}
+              classNames="search-and-filter__contextual-menu"
+            >
+              <>
```
bartaz, line 225: Is the `<>` fragment needed? They all seem to be wrapped in single parent component.

src/components/SearchAndFilter/SearchAndFilter.scss
```diff
@@ -111,4 +97,32 @@
...
   }
+
+  &__contextual-menu {
+    display: block;
+
+    .p-contextual-menu__dropdown {
+      padding: 1rem;
+    }
+
+    &-enter,
```
bartaz, line 108: These class names seem to be controlled by `CSSTransition` component. They are not consistent with BEM and Vanilla, so it would be worth noting that in a comment here, because quite likely if/when migrating this to Vanilla CSS we would rename them (and break the CSSTransition functionality).

I don't know what to do with such styles in a long run. Technicaly we want to move everything to Vanilla CSS, but also we don't want to put class names dependent on arbitrary react transition library into Vanilla CSS.
Is there any way to configure these class names so they can be renamed in Vanilla "BEM" consistent way?

I'm fine with leaving them as they are for now here, but it should be commented in the code where do they come from and that their names can't be changed because external library controls them.

src/components/SearchAndFilter/SearchAndFilter.js
huwshimi: `<CSSTransition>` only allows one child. In this case we have the search term and results as the two children which need to be wrapped.

src/components/SearchAndFilter/SearchAndFilter.scss
huwshimi: We can supply all the classnames in BEM format: http://reactcommunity.org/react-transition-group/css-transition#CSSTransition-prop-classNames.

I'll update this to use some BEM classes and we can adjust them as needed when we move it to Vanilla.

bartaz: Great, thanks.

src/hooks/usePrevious.ts
```diff
@@ -3,10 +3,11 @@ import { useEffect, useRef } from "react";
 /**
  * A hook to handle storing the previous value of anything.
  * @param value The value to watch.
+ * @param setInitial Whether the it should initialise with the current value.
```
bartaz, line 6: Seems like `the` is not needed here (or there is some other grammar error).

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -0,0 +1,340 @@
...
+  positionNode?: HTMLElement;
+  toggleAppearance?: string;
+  toggleClassName?: string;
+  toggleDisabled?: boolean;
+  toggleLabel?: string;
+  toggleLabelFirst?: boolean;
+  visible?: boolean;
+};
+/**
+ * Get the node to use for positioning the menu.
```
squidsoup, line 34: In tsdoc the param is separated with the description by a `-`

squidsoup (approved):
Looks good huw, just noticed a couple of small things.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,74 @@
...
+  appearance?: string;
+  children?: ReactNode;
+  className?: string;
+  dense?: boolean;
+  disabled?: boolean;
+  element?: ElementType;
+  hasIcon?: boolean;
+  inline?: boolean;
+  onClick?: (
+    evt: MouseEvent | React.MouseEvent | MouseEventHandler<HTMLButtonElement>
```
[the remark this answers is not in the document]
huwshimi: I've updated this to `React.MouseEvent` as that seems to be the only type it will actually accept. I've also updated the Button type to extend the base element (`} & HTMLAttributes<HTMLElement>;`) as this component could be a `button`, `anchor` or any other element that needs to appear as a button.

PR #280, opened 2020-09-16
Add returnSearchData function to default story

barrymcgee:
## Done

Add returnSearchData function to default story

### Storybook

```shell
yarn start
```

- View `/?path=/story/searchandfilter--default-story`
- Should load without error

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.js
+++ b/src/components/SearchAndFilter/SearchAndFilter.js
@@ -9,11 +9,7 @@ import { overflowingChipsCount } from "./shared";
 
 import "./SearchAndFilter.scss";
 
-const SearchAndFilter = ({
-  externallyControlled = false,
-  filterPanelData,
-  returnSearchData,
-}) => {
+const SearchAndFilter = ({ filterPanelData, returnSearchData }) => {
   const [searchData, setSearchData] = useState([]);
   const [searchTerm, setSearchTerm] = useState("");
   const [filterPanelHidden, setFilterPanelHidden] = useState(true);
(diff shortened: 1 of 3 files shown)


src/components/SearchAndFilter/SearchAndFilter.stories.mdx
```diff
@@ -4,7 +4,16 @@ import SearchAndFilter from "./SearchAndFilter";
 
 <Meta title="SearchAndFilter" component={SearchAndFilter} />
 
-export const Template = (args) => <SearchAndFilter {...args} />;
+export const Template = (args) => (
+  <SearchAndFilter
+    {...args}
+    returnSearchData={(searchData) => {
```
bartaz, line 10: If the `returnSearchData` prop is required, it should be marked required in the component.
If it's not required, component should render without it.

barrymcgee: Good point, I've now cleaned that up.

bartaz:
Not sure if we have any control over it, but the example in storybook appears in small scrollable frame:

Screenshot 2020-09-17 at 16 26 52

bartaz (approved):
LGTM, thanks!

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -212,7 +208,7 @@ const SearchAndFilter = ({
         })}
         <SearchBox
           autocomplete="off"
-          externallyControlled={externallyControlled}
+          externallyControlled={true}
```
squidsoup, line 211: Small thing, but the convention for `true` props, is just to list the prop, e.g.

`<Foo aTrueProp />`

PR #283, opened 2020-09-21
Scope searchbox amends to search and filter

### LINKED ISSUE #281: SearchBox broken in 0.10.0
The updated SearchBox has some styling issues:

Screen Shot 2020-09-21 at 9 33 32 am

Also the change to using forwardRef has made it so tests can't do `wrapper.find("SearchBox")` any more (returns nothing).

Not sure if something can be done to make that work again. I wonder if there's something here that might help:
https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

barrymcgee:
## Done

Scope searchbox amends to search and filter

### Storybook

```shell
yarn start
```

### QA

- View Storybook
- View SearchBox component and verify everything looks correct

Fixes #281


src/components/SearchAndFilter/SearchAndFilter.scss
```diff
@@ -111,4 +111,32 @@
     top: 3px;
     z-index: 999;
   }
+
+  .p-search-box {
```
bartaz, line 115: That's probably something to update later (maybe when we upstream this to Vanilla CSS), but it seems that if so much of this needs to be overridden, this shouldn't really be using `p-search-box` class at all and be something like `p-search-and-filer__search`.

But now that it's scoped this can wait I guess not to introduce to many changes to the component.

bartaz (approved):
LGTM, thanks.

Search box now renders correctly.

PR #287, opened 2020-09-23
Set the component name for SearchBox.

huwshimi:
## Done

- Set the display name for SearchBox so that it does not appear as "ForwardRef".

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

No QA.

## Fixes

Fixes: #281


squidsoup (approved):
Nice find!

PR #290, opened 2020-09-23
Modular table icons and empty state

bartaz:
## Done

- Added table icons support to modular table
- Added empty message support to modular table
- Improved modular table documentation in storybook

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check ModularTable in Storybook

## Fixes

Fixes: https://github.com/canonical-web-and-design/vanilla-squad/issues/899


bartaz:
@huwshimi @squidsoup Do you know why stylelint would check snapshot files and fail?

https://github.com/canonical-web-and-design/react-components/pull/290/checks?check_run_id=1155135437

Screenshot 2020-09-23 at 16 16 20

src/components/ModularTable/ModularTable.tsx
```diff
@@ -53,14 +59,23 @@ function ModularTable({
...
                         className: cell.column.className,
                       },
+                      {
+                        className: hasColumnIcon
+                          ? "p-table__cell--icon-placeholder"
+                          : "",
+                      },
                     ])}
                   >
+                    {iconName && <Icon name={iconName} />}
```
bartaz, line 78: @squidsoup 

I tried the approach with defining two separate props and separate types in TS like this:

Screenshot 2020-09-24 at 12 01 18

It worked on its own, correctly suggested types for name, etc.

But it felt short in this line, because having separate props in `Icon` component, means any components that use the `Icon` need to make this distinction as well, which becomes problematic if we want them to easily and consistently support both built-in and custom icons.

So I hope it's good enough as is for now.

huwshimi: Another approach would be to provide a typescript enum for the default icons.

```ts
export enum ICONS {
  plus = "plus,
  minus = "minus",
  ...
}
```

Which would be used like this:

```jsx
import { Icon, ICONS } from "@canonical/react-components"
 
<Icon name={ICONS.plus} />
<Icon name="custom-icon" />
```

bartaz: Great suggestion @huwshimi, thanks!

squidsoup (approved):
A couple of trivial bits for your consideration left, but looks great, thank you @bartaz!

PR #306, opened 2020-10-15
Fix u-hide override affecting expanding table headers

Caleb-Ellis:
## Done

- Namespaced `ModularTable.scss` so that the `u-hide` override would not affect expanding tables.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Run `yarn docs`
- Navigate to the MainTable > Expanding story
- Check that the table headers are correctly aligned with the content

### DIFF
--- a/src/components/ModularTable/ModularTable.tsx
+++ b/src/components/ModularTable/ModularTable.tsx
@@ -35,7 +35,7 @@ function ModularTable({
   const showEmpty: boolean = emptyMsg && (!rows || rows.length === 0);
 
   return (
-    <Table {...getTableProps()}>
+    <Table {...getTableProps({ className: "p-modular-table" })}>
       <thead>
         {headerGroups.map((headerGroup) => (
           <TableRow {...headerGroup.getHeaderGroupProps()}>
--- a/src/components/ModularTable/ModularTable.scss
+++ b/src/components/ModularTable/ModularTable.scss
@@ -5,29 +5,31 @@
 // Vanilla issue: https://github.com/canonical-web-and-design/vanilla-framework/issues/3337
 
 // stylelint-disable selector-no-qualifying-type
-td.u-hide,
-th.u-hide {
-  display: table-cell !important;
-  width: 0 !important;
+.p-modular-table {
+  td.u-hide,
+  th.u-hide {
+    display: table-cell !important;
+    width: 0 !important;
 
-  &--small {
-    @media screen and (max-width: $breakpoint-medium) {
-      display: table-cell !important;
-      width: 0 !important;
+    &--small {
+      @media screen and (max-width: $breakpoint-medium) {
+        display: table-cell !important;
+        width: 0 !important;
+      }
     }
-  }
 
-  &--medium {
-    @media (min-width: $breakpoint-medium) and (max-width: $breakpoint-large) {
-      display: table-cell !important;
-      width: 0 !important;
+    &--medium {
+      @media (min-width: $breakpoint-medium) and (max-width: $breakpoint-large) {
+        display: table-cell !important;
+        width: 0 !important;
+      }
     }
-  }
 
-  &--large {
-    @media screen and (min-width: $breakpoint-large) {
-      display: table-cell !important;
-      width: 0 !important;
+    &--large {
+      @media screen and (min-width: $breakpoint-large) {
+        display: table-cell !important;
+        width: 0 !important;
+      }
     }
   }
 }


src/components/ModularTable/ModularTable.tsx
```diff
@@ -107,6 +108,7 @@ function ModularTable({
 }
 
 ModularTable.propTypes = {
+  className: PropTypes.string,
```
huwshimi, line 111: I don't think this change is quite right, passing `className` to a ModularTable doesn't seem to apply the class.

src/components/ModularTable/ModularTable.tsx
```diff
@@ -35,7 +36,7 @@ function ModularTable({
   const showEmpty: boolean = emptyMsg && (!rows || rows.length === 0);
 
   return (
-    <Table {...getTableProps()}>
+    <Table className="p-modular-table" {...getTableProps()}>
```
huwshimi, line 39: Will that classname prop get overridden by `getTableProps()`?

huwshimi (changes requested):

src/components/ModularTable/ModularTable.tsx
Caleb-Ellis: Ah oops, brainfart. Added a hardcoded className and thought "yeah that's totally a prop".

I think I'll leave out being able to pass a `className` prop and just have the hardcoded wrapper className.

src/components/ModularTable/ModularTable.tsx
Caleb-Ellis: As far as I can tell, `getTableProps()` only returns `{ role: "table" }`. Still, after looking at the docs I think the correct way of handling this is adding the additional prop directly into the `getTableProps()` function.

huwshimi (approved):

bartaz:
@Caleb-Ellis @huwshimi The `td.u-hide` workaround was meant for ALL the tables, because it messes up how colspan works. 

I guess the issue with expanding table is that expanding table is using flex layout instead of table layout.
So it's the expanding table that is exception (and should be treated differently), not modular table.

bartaz:
@Caleb-Ellis We have an issue in Vanilla already to upsteam the u-hide fix for tables. Can you provide the details about what it causes in expending table there, so we can take that into account? https://github.com/canonical-web-and-design/vanilla-framework/issues/3337

bartaz:
And it seems it's another regression that should have been caught by Percy.

Caleb-Ellis:
> @Caleb-Ellis @huwshimi The `td.u-hide` workaround was meant for ALL the tables, because it messes up how colspan works.

Ah okay, I thought they were styles specific to the ModularTable component that were "leaking". In any case I don't think `ModularTable.scss` is the right place for those styles to live.

> I guess the issue with expanding table is that expanding table is using flex layout instead of table layout. So it's the expanding table that is exception (and should be treated differently), not modular table.

That's fair, I agree.

bartaz:
> In any case I don't think ModularTable.scss is the right place for those styles to live.

True. They should be in Table.scss probably.

@Caleb-Ellis I looked into expanding table, and it seems that the issue is that we have to hide last cell in the header (for "validation" reasons). It's weird approach in my opinion, but I guess we are not in a point where we can easily change it.

So I think that we should have `.p-table-expanding .u-hide` defined specifically for expanding table, or even better not use `u-hide` utility as part of required pattern style, but have pattern specific class `p-table-expanding__cell-placeholder` (or something like this) to make sure it's just for expanding table. I guess `.p-table-expanding thead td:last-child { display: none }` would technically work as well, but may be a bit more fiddly and error prone.

Caleb-Ellis:
I think coupling `u-hide` (or any utility class) to a component is odd, including the original override in `ModularTable.scss`. In my mind a utility does one thing and one thing only and isn't affected by where it sits in the DOM. So I think a `p-table-expanding__cell-placeholder` is the best option, and perhaps we need a separate utility for shrinking things to `width: 0`?

bartaz:
@Caleb-Ellis I agree, that utils should do one thing only, so ideally we should have separate `u-hide--table-cell` maybe?

The reason I went with table scoped u-hide was to make sure it works with existing `u-hide--small` etc classes, that are used quite a lot for responsive hiding of columns. So scoping u-hide for table made sure the change propagates without changing the class name. But in the end I guess it would be safer to have new util and update where needed in products.

huwshimi:
@Caleb-Ellis would you mind reverting this change?

PR #307, opened 2020-10-15
Prepare 0.13.1 release

huwshimi:
## Done

- Updated changelog, version and docs for 0.13.1.


bartaz (changes requested):
@caldav @huwshimi I commented on #306, because for me it doesn't seem to be a correct fix.

We should approach this differently, so unless you really need this workaround released (because it's blocking you), we should provide a different fix. Fix that will affect expanding table rather than modular table.

Caleb-Ellis:
@bartaz We have a MAAS release very soon so I'm keen to get this out sooner rather than later. Now that I know the styles were meant to apply to *all* tables I agree it's not the correct fix, but as it stands react-components v0.13.0 breaks expanding tables.

Unless namespacing the `th.u-hide` override has broken something else, I suggest we merge and release this with the intention of removing `ModularTable.scss` once those changes have been upstreamed to Vanilla.

huwshimi:
Closing this, will wait for Vanilla to implement the fix.

PR #313, opened 2020-10-20
Add classes to the wrapping tooltip elements.

huwshimi:
## Done

- Added classnames to the wrapping tooltip elements.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- View the Tooltip page in Storybook.
- Inspect an element that displays a tooltip. You should see classes applied to the wrapping elements.

## Fixes

Needed for: https://github.com/canonical-web-and-design/maas-ui/issues/1784.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated expanding MainTable to use aria-hidden instead of .u-hide utility class
 - Updated Vanilla framework to version 2.19.2
-- Added class names to the wrapping elements for the Tooltip component.
+- Added `positionElementClassName` prop to allow class names to be passed to the position element for the Tooltip component.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated expanding MainTable to use aria-hidden instead of .u-hide utility class
 - Updated Vanilla framework to version 2.19.2
+- Added class names to the wrapping elements for the Tooltip component.
 
 ### Removed
 
--- a/src/components/Tooltip/Tooltip.stories.mdx
+++ b/src/components/Tooltip/Tooltip.stories.mdx
@@ -50,3 +50,22 @@ An alternative use of tooltips is to provide information on a disabled actionabl
     {Template.bind({})}
   </Story>
 </Canvas>
+
+### Targeting elements
+
+The tooltip elements are displayed inside a [Portal](https://reactjs.org/docs/portals.html) so that the tooltip is not constrained by its current position in the DOM.
+Sometimes however, you need to target the tooltip elements. This can be achieved by passing class names to the tooltip elements. You can also pass classes to the elements that trigger and position the tooltip.
+
+<Canvas>
+  <Story
+    name="Targeting elements"
+    args={{
+      className: "custom-wrapping-class",
+      positionElementClassName: "custom-position-class",
+      tooltipClassName: "custom-tooltip-class",
+      message: "Tooltip message to display.",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -215,13 +215,17 @@ const Tooltip = ({
     <>
       {message ? (
         <span
-          className={className}
+          className={classNames("p-tooltip--wrapper", className)}
           onBlur={closePortal}
           onFocus={openPortal}
           onMouseOut={closePortal}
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- BUGFIX: Updated MainTable so that the table headers aren't squashed when `expanding` is `true`.
```
bartaz, line 14: This seems to be from the other PR.

bartaz:
How does adding a class name fix the bug mentioned in the description? There doesn't seem to be any style attached to it.
Also, as I don't fully understand the issue - is the issue with Vanilla tooltip pattern (so should be fixed in Vanilla) or is it solely with React component implementation?

The other problem I see is that we are introducing class name that is not coming from Vanilla (although it pretends to be).
`p-tooltip--wrapper` and `p-tooltip--wrapper-inner` seem to use Vanilla naming convention so people may expect them coming from Vanilla (while they don't). Additionally, convention of `--wrapper` is meant for modifiers of the pattern, like `p-tooltip--top-left` and they should be applied on a root element not on some element outside the pattern. 

I guess we should think about some process and convention for doing such fixes, and if there really is a need to adding classes in react that are not from Vanilla they should somehow be distinguishable. Something like `rc-tooltip-wrapper`?
But we should also have as little of them as possible.

CHANGELOG.md
huwshimi: Yeah I just added it in this PR while I was updating the changelog.

huwshimi:
> How does adding a class name fix the bug mentioned in the description? There doesn't seem to be any style attached to it.
> Also, as I don't fully understand the issue - is the issue with Vanilla tooltip pattern (so should be fixed in Vanilla) or is it solely with React component implementation?

This doesn't fix the issue by itself, but allows us to target the elements and add specific styles in maas-ui. It's not an issue with Vanilla, just an issue with how we position the elements in react-components to allow for auto-adjusting the tooltips.

> 
> The other problem I see is that we are introducing class name that is not coming from Vanilla (although it pretends to be).
> `p-tooltip--wrapper` and `p-tooltip--wrapper-inner` seem to use Vanilla naming convention so people may expect them coming from Vanilla (while they don't). Additionally, convention of `--wrapper` is meant for modifiers of the pattern, like `p-tooltip--top-left` and they should be applied on a root element not on some element outside the pattern.

Ah yes, oops, it should have been `__wrapper`.

> 
> I guess we should think about some process and convention for doing such fixes, and if there really is a need to adding classes in react that are not from Vanilla they should somehow be distinguishable. Something like `rc-tooltip-wrapper`?
> But we should also have as little of them as possible.

Sure. I think there will always be a few extra things to add functionality that Vanilla doesn't support by default. In this case it's to handle the automatic positioning (which Vanilla will never be able to support).

In think `rc-tooltip-wrapper` is OK, but maybe the format should be `rc-[component name in react-components]` and then use the normal `__wrapper` and `--modifier` rules. Something like `rc-tooltip__wrapper` or `rc-contextualmenu--dark`.

If that seems OK, I'll update this PR.

bartaz:
> This doesn't fix the issue by itself, but allows us to target the elements and add specific styles in maas-ui.

I feel bit worried about this approach. It seems like we are adding a class name to a library, but a library itself is not styling it at all, it's just for external users to style it. I don't think we should do that. We should only have class names that are styled by us.

If a class name is needed to be styled outside (by maas-ui for example) the component should have a prop to add classname to itself. So maas-ui would use the Tooltip component, set a classname on it and style this classname. This way maas-ui has complete control over this classname and the styling.

huwshimi:
> > This doesn't fix the issue by itself, but allows us to target the elements and add specific styles in maas-ui.
> 
> I feel bit worried about this approach. It seems like we are adding a class name to a library, but a library itself is not styling it at all, it's just for external users to style it. I don't think we should do that. We should only have class names that are styled by us.
> 
> If a class name is needed to be styled outside (by maas-ui for example) the component should have a prop to add classname to itself. So maas-ui would use the Tooltip component, set a classname on it and style this classname. This way maas-ui has complete control over this classname and the styling.

That's a much better idea. In fact, you can already supply a class for the wrapper element, I just need to add a class prop for the inner element.

huwshimi:
@bartaz this has been updated.

CHANGELOG.md
bartaz: But we don't have this bugfix in yet? The one that was merged was reverted, wasn't it?

huwshimi: Oh yes, sorry, you're totally right! Removed.

bartaz (approved):
LGTM, thanks!

bartaz:
Maybe would be worth adding a storybook example that uses new props?

huwshimi:
> Maybe would be worth adding a storybook example that uses new props?

Thanks, done!

PR #324, opened 2020-11-02
fix: allow null assignment to tooltip message

squidsoup:
## Done
As the Tooltip component allows you to control conditional rendering of the tooltip with the message prop, we should allow this to be explicitly assigned null.

This is useful when you are setting `message` in a ternary, e.g.

```ts
<Tooltip message={foo? "please render me": null} />
```

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- n/a

## Fixes

Fixes: # .

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -38,7 +38,7 @@ export type Props = {
   className?: string;
   children: ReactNode;
   followMouse?: boolean;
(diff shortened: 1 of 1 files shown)


Caleb-Ellis (approved):
LGTM!

PR #326, opened 2020-11-03
Remove ContextualMenu / CSSTransition

### LINKED ISSUE #315: [Search and filter] Panel should stay open at all times when clicking within component

### LINKED ISSUE #320: CSSTransitons in ContextulMenu are not compatible with Strict Mode React projects
This console error is thrown when including SearchAndFilter in a clean sandbox

Screenshot 2020-10-29 at 14 54 15

barrymcgee:
## Done

- Decouple ContextualMenu / CSSTransition from Search And Filter
- Ensure panel stays open when clicking inside component

## Why

- ContextualMenu has grown far too complex to be used in this context. It also uses inline styles which make it very difficult to override when nested inside SearchAndFilter.
- CSSTransiton adds many levels of complexity and is also not compatible downstream when consumed by React projects in Strict mode.

### Storybook

```shell
yarn start
```

Fixes #320, Fixes #315 

- Go to search and filter pattern
- Ensure it displays and works as expected

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.js
+++ b/src/components/SearchAndFilter/SearchAndFilter.js
@@ -1,9 +1,7 @@
-import { CSSTransition } from "react-transition-group";
 import React, { useState, useEffect, useRef } from "react";
 import PropTypes from "prop-types";
 
 import SearchBox from "../SearchBox";
-import ContextualMenu from "../ContextualMenu";
 import FilterPanelSection from "./FilterPanelSection";
 import Chip from "../Chip";
 import { overflowingChipsCount } from "./shared";
@@ -49,6 +47,39 @@ const SearchAndFilter = ({ filterPanelData, returnSearchData }) => {
     };
   }, [searchContainerActive]);
 
+  // This useEffect sets up listeners so the panel will close if user clicks
+  // anywhere else on the page or hits the escape key
+  useEffect(() => {
+    const closePanel = () => {
+      setFilterPanelHidden(true);
+    };
+
+    const mouseDown = (e) => {
+      // Check if click is outside of filter panel
+      if (!searchAndFilterRef?.current?.contains(e.target)) {
+        // If so, close the panel
+        closePanel();
+      }
+    };
+
+    const keyDown = (e) => {
+      if (e.code === "Escape") {
+        // Close panel if Esc keydown detected
+        closePanel();
+      }
+    };
+
+    // Add listener on document to capture click events
+    document.addEventListener("mousedown", mouseDown);
+    // Add listener on document to capture keydown events
+    document.addEventListener("keydown", keyDown);
+    // return function to be called when unmounted
+    return () => {
+      document.removeEventListener("mousedown", mouseDown);
+      document.removeEventListener("keydown", keyDown);
+    };
+  }, []);
+
   // Add passed chip to the searchData array
   const toggleSelected = (chip) => {
     const currentSelected = [...searchData];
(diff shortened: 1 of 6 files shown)


src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -1,9 +1,7 @@
-import { CSSTransition } from "react-transition-group";
```
bartaz, line 1: It it used elsewhere? Maybe it can be removed from dependencies?

src/components/SearchAndFilter/SearchAndFilter.scss
```diff
@@ -68,7 +62,14 @@
   }
 
   &__panel {
-    padding: 0;
+    opacity: 1;
+    padding: 0.5rem 1rem 0;
+    position: absolute;
+    border-radius: $border-radius;
```
bartaz, line 68: I think linter fails here as it expects alphabetical order.

bartaz:
Not very visible on gif, but there seems to be a weird effect that chips disapear first, then the rest of dropdown:

search-fade

bartaz:
Not sure if it's related to this PR, but when you type something to search and filter, and start to delete, at some point it throws errors:

Screenshot 2020-11-03 at 15 48 43

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -49,6 +47,39 @@ const SearchAndFilter = ({ filterPanelData, returnSearchData }) => {
...
+    // Add listener on document to capture click events
+    document.addEventListener("mousedown", mouseDown);
+    // Add listener on document to capture keydown events
+    document.addEventListener("keydown", keyDown);
+    // return function to be called when unmounted
+    return () => {
+      document.removeEventListener("mousedown", mouseDown);
+      document.removeEventListener("keydown", keyDown);
+    };
+  }, [searchTerm]);
```
bartaz, line 81: Out of curiosity, why it has to be recreated when `searchTerm` changes? Effect doesn't seem to use searchTerm.

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -49,6 +47,39 @@ const SearchAndFilter = ({ filterPanelData, returnSearchData }) => {
...
+    };
+
+    const keyDown = (e) => {
+      if (e.code === "Escape") {
+        // Close panel if Esc keydown detected
+        closePanel();
+      }
+    };
+
+    // Add listener on document to capture click events
```
hatched, line 72: You don't really need comments that say the same thing as the line of code, same for the comments above. :)

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: I assumed it was as the justification was including it in the first place was to have consistency.. but apparently not. Removing.

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: This looks like a bug, have now removed that dep.

barrymcgee:
@bartaz Thanks, good catches. Should all be resolved now.

hatched (approved):

src/components/SearchAndFilter/SearchAndFilter.js
huwshimi: Sorry I might not have been very clear. This was only added to support the transitions in SearchAndFilter (it needed a way to only run the animations on first render inside the portal).

PR #334, opened 2020-11-16
Search and filter tweaks

barrymcgee:
## Done

- Search and filter tweaks and bug fixes
- Added instructions on how to develop with project locally

### Storybook

```shell
yarn start
```

- View Search and Filter component in Storybook
- Verify that existingSearchData prop displays properly
- Check instructions in README on local development
- Try to run project locally in React project

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,3 +39,63 @@ You might want to:
 - Read abut [developing components](https://github.com/canonical-web-and-design/react-components/blob/master/HACKING.md).
 - Find out how to [publish to NPM](https://github.com/canonical-web-and-design/react-components/blob/master/PUBLISH-NPM-PACKAGE.md).
 - Know how to [publish the docs](https://github.com/canonical-web-and-design/react-components/blob/master/PUBLISHING-DOCS.md) to GitHub Pages.
+
+## Developing locally using this repository
+
+You may wish to link this library directly to your projects while developing locally.
+
+You can do this by cloning this repo to your local workspace;
+
+```shell
+git clone https://github.com/canonical-web-and-design/react-components
+```
+
+If you then drop into that folder and run;
+
+```shell
+yarn link
+```
+
+...this will add the project to a local yarn registry.
+
+Switching back to the project you are developing, run;
+
+```shell
+yarn link @canonical/react-components
+yarn install
+```
+
+...to pull the linked project from the local registry. If you now run `yarn build-watch` in your `react-components` folder, your project should pick up any changes on refresh or hot module reload.
+
+### Known issue: Invalid hook call (Issue [#327](https://github.com/canonical-web-and-design/react-components/issues/327))
+
+If you have followed the instructions above and see an "Invalid hooks call" in your local project, it's probably because your project is now trying to run two different versions of React.
+
+To work around this, you'll also have to do the following to link `react` and `react-dom` in this project to the local registry;
+
+```
+cd react-components
+yarn install
+cd node_modules/react
+yarn link
+cd ../../node_modules/react-dom
+yarn link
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -91,8 +92,9 @@
   },
   "scripts": {
     "build": "rm -rf dist && NODE_ENV=production babel src --out-dir dist --copy-files --extensions '.js,.jsx,.ts,.tsx'; yarn build-declaration",
+    "build-local": "NODE_ENV=production babel src --out-dir dist --copy-files --extensions '.js,.jsx,.tsx'",
     "build-declaration": "tsc",
-    "build-watch": "yarn build --watch",
+    "build-watch": "watch 'yarn run build-local' ./src",
```
bartaz, line 97: Why using `watch` here? Isn't `--watch` flag for babel enough?

It seems that before TS was introduced `yarn build --watch` was working, because it was passing `--watch` to babel. Now `tsc` is last step of build, so `--watch` is being passed to TS instead.
If the babel build is extracted to `build-local`, maybe `yarn build-local --watch` would work?

bartaz: But I'm not really sure how watch should work here if we have babel and TS steps in build.
@squidsoup Any thoughts on that Kit?

src/components/SearchAndFilter/SearchAndFilter.test.js
```diff
@@ -248,4 +248,24 @@ describe("Search and filter", () => {
       '<span class="p-chip__value"><strong>Google</strong></span>'
     );
   });
+
+  it("Existing search data displays correctly", () => {
```
hatched, line 252: Should we also have a test for multiple elements of existing search data?

README.md
```diff
@@ -39,3 +39,63 @@ You might want to:
...
+cd YOUR_PROJECT
+yarn link react
+yarn link react-dom
+```
+
+**Note:** When you're finished working locally - don't forget to go back and unlink;
+
+```
+cd react-components
+yarn unlink
```
hatched, line 92: You don't really _need_ to unlink from the source repository as they will just sit around available if anything else wants to link to them later.

README.md
```diff
@@ -39,3 +39,63 @@ You might want to:
...
+cd node_modules/react
+yarn unlink
+cd ../../node_modules/react-dom
+yarn unlink
+cd YOUR_PROJECT
+yarn unlink @canonical/react-components
+yarn link react
+yarn link react-dom
```
hatched, line 100: These links are interesting, I have always had to do the --force install afterwards, does this just 'link' to the locally installed version again?

README.md
```diff
@@ -39,3 +39,63 @@ You might want to:
...
+You can do this by cloning this repo to your local workspace;
+
+```shell
+git clone https://github.com/canonical-web-and-design/react-components
+```
+
+If you then drop into that folder and run;
+
+```shell
+yarn link
```
huwshimi, line 56: There's also a script to help with this: https://github.com/canonical-web-and-design/react-components/blob/master/package.json#L99 it also links react like you do in the next section.

There's an unlink-package script to reverse it: https://github.com/canonical-web-and-design/react-components/blob/master/package.json#L109

barrymcgee: Thanks - didn't spot these..have now added them to the updated README

README.md
barrymcgee: The force install is to overwrite the links when you're done.

Screenshot 2020-11-17 at 14 47 38

package.json
barrymcgee: Good catch.. don't need the watch package afterall

hatched (approved):

PR #355, opened 2020-12-08
Improve docs for SearchAndFilter

barrymcgee:
## Done

- Improve Storybook docs for SearchAndFilter

### Storybook

```shell
yarn start
```

View Search And Filter variations in Storybook

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.js
+++ b/src/components/SearchAndFilter/SearchAndFilter.js
@@ -9,7 +9,7 @@ import { overflowingChipsCount, isChipInArray } from "./shared";
 import "./SearchAndFilter.scss";
 
 const SearchAndFilter = ({
-  filterPanelData,
+  filterPanelData = [],
   returnSearchData,
   existingSearchData = [],
 }) => {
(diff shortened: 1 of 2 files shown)


barrymcgee:
@huwshimi @squidsoup I can't seem to be my ArgsTable here to infer my props - any idea why? The proptypes are properly set up in the component AFAIK...

Screenshot 2020-12-08 at 11 33 32

cristinadresch:
@huwshimi could you review this one please

src/components/SearchAndFilter/SearchAndFilter.js
```diff
@@ -9,7 +9,7 @@ import { overflowingChipsCount, isChipInArray } from "./shared";
 import "./SearchAndFilter.scss";
 
 const SearchAndFilter = ({
-  filterPanelData,
+  filterPanelData = [],
```
huwshimi, line 12: The prop type for this has `isRequired` so there are a bunch of prop warnings for this and `returnSearchData` on the storybook page.

huwshimi (approved):
Just need to fix up the prop type warnings and this should be good.

I had a quick look at the ArgsTable issue and it seems to be caused by something internal to the component. If I remove a bunch of the `useEffect` blocks and callbacks it appears again. might be worth seeing if there's some leakage somewhere that's causing a bunch of unintended rerenders or something. You might be able to try adding `useCallback` and our custom `useListener` hook to see if that improves things.

src/components/SearchAndFilter/SearchAndFilter.js
barrymcgee: Good catch, thanks. Now fixed.

PR #363, opened 2021-01-12
Remove custom styles from Spinner component

bartaz:
## Done

- Removes custom styles from Spinner component
- Removes unnecessary (?) inline property

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check the Spinner storybook if it renders correctly and has all needed options
  - https://react-components-363.demos.haus/?path=/docs/spinner--default-story
- Ideally, `yarn link` this feature branch to external project that uses Spinner and check if there are no regressions

## Fixes

Fixes: https://github.com/canonical-web-and-design/vanilla-framework/issues/3145


bartaz:
@Caleb-Ellis Following the discussion on Spinner styles from https://github.com/canonical-web-and-design/vanilla-framework/issues/3145 I made this PR.

It removes all styling from Spinner component. This resulted also in removing 'inline' variant (that was adding additional margin).
Could you have a look if this is fine as it is - will it cover all needed cases in MAAS? Or is the `inline` version still needed for some cases? If so, could you give us some screenshots of the Spinner in the context, so we can either add some styling needed or find another solution?

Caleb-Ellis (approved):
LGTM! We do use `inline` in a couple of places in MAAS but it's literally only used to nudge the spinner a tiny bit to the right.

Thanks @bartaz!

Issue #368, opened 2021-01-18
Remove SummaryButton styles


sowasred2012:
This is a follow up to https://github.com/canonical-web-and-design/vanilla-framework/issues/3344

The SummaryButton has a tiny amount of custom CSS applied, but we can achieve the same thing by removing them and applying the updated (as of [Vanilla 2.22](https://github.com/canonical-web-and-design/vanilla-framework/releases/tag/v2.22.0)) `is-inline` class to the `button` element.

PR #369, opened 2021-01-18
Remove .p-summary__text scss

sowasred2012:
## Done

Removed unnecessary custom styling for the SummaryButton component, used Vanilla's `is-inline` button utility to achieve the same effect.

### Storybook

Run:

```shell
yarn start
```

Then visit http://localhost:8403:/?path=/story/summarybutton--default-story

### QA steps

- See that the SummaryButton component looks good across all screen widths

## Fixes

Fixes: #368  .

## Screenshots:

Before:
Screenshot from 2021-01-18 15-00-15

After:
Screenshot from 2021-01-18 14-57-59

### DIFF
--- a/src/components/SummaryButton/SummaryButton.scss
+++ b/src/components/SummaryButton/SummaryButton.scss
@@ -1,6 +1,0 @@
-@import "~vanilla-framework/scss/settings";
-
-.p-summary__text {
-  display: inline-block;
-  margin-right: $sph-inner--small;
-}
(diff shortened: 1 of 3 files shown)


anthonydillon (approved):
Awesome thanks :+1:

sowasred2012:
@anthonydillon sorry, could you take another quick look - realised we don't want `is-inline` on the button if there's no text (otherwise there's unnecessary left margin)

anthonydillon:
Yea saw that, makes sense. Good catch

PR #381, opened 2021-01-27
Initial implementation of CodeSnippet component

### LINKED ISSUE #373: Build CodeSnippet component
The code snippet pattern was just introduced in Vanilla (https://vanillaframework.io/docs/base/code#code-snippet), we should have a react component to match.

bartaz:
## Done

- React implementation of new Vanilla Code snippet component

Fixes #373

### QA steps

- Open demo: https://react-components-381.demos.haus/?path=/docs/codesnippet--default-story
- Make sure all the examples work as expected
- Make sure documentation and examples are clear on how to use the component

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```


src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+CodeSnippetBlock.props = {
+  code: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  numbered: PropTypes.bool,
+  icon: PropTypes.oneOf(["linuxPrompt", "windowsPrompt", "url"]),
+  isWrapped: PropTypes.bool,
+};
+
+CodeSnippetBlock.defaultProps = {
```
hatched, line 59: Lets not use `defaultProps` but instead define them in the function signature so that it's clear that defaults exist when reading the code.

bartaz: Ah, sure, and storybook picks those up as well for the generated docs. Good catch, thanks.

sowasred2012:
LGTM - just one thing, syntax highlighting on the code example for the dropdowns is off, it's very difficult make some of it out:
Screenshot from 2021-02-05 12-10-30

Perhaps it's a broader problem, outside the scope of this PR though.

sowasred2012 (approved):

bartaz:
@hatched @huwshimi Could you check if the code is up to our React standards? One thing I'm not sure about, when the component is internally built of several smaller ones, should we just keep them together in one file, or better split them into own files (still under /CodeSnippet folder)?

hatched:
> when the component is internally built of several smaller ones, should we just keep them together in one file, or better split them into own files (still under /CodeSnippet folder)?

Your file here seems to be fine for me.

I typically deal with this on a case by case basis. If the child components are small and won't ever be used outside of the parent then I will keep in them in the same file. At some point those files will grow beyond a point where they feel comfortable to work on and I'll split it out.

I do however order them by importance from top to bottom; So your top one would be the parent that's exported and all children would be below.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    // wrap code lines in spans (and preserve the whitespace)
+    const lines = code.split(/\r?\n/);
+    numberedCode = lines.map((line, i) => (
+      <React.Fragment key={`p-code-snippet__line-${i}`}>
+        <span className="p-code-snippet__line">{line}</span>
+        {"\n"}
+      </React.Fragment>
+    ));
+  } else if (icon) {
```
hatched, line 65: Do we really want to special case `icon` like this? Won't this preclude this use case...
```
$ cd foo
$ ./run something
```

A more generic approach might be beneficial here. Where you run any code provided through the spanification above and add the appropriate icon to the start of the line.

src/components/CodeSnippet/CodeSnippet.stories.mdx
```diff
@@ -0,0 +1,144 @@
...
+    return <CodeSnippet blocks={[
+       {
+         code: code[lang],
+         dropdowns: [
+           { options: [
+              { value: "js", label: "JS" },
+              { value: "css", label: "CSS" },
+              { value: "html", label: "HTML", selected: true },
+            ],
+            onChange: (event) => { setLang(event.target.value) }
```
hatched, line 136: I feel like this switching should be a feature of the `CodeSnippet` component itself. Maybe for a follow-up.

hatched (changes requested):
Good work with this initial implementation @bartaz It works well and the code is easy to follow. 👍 

Sorry I have made a lot of comments. It's entirely possible some of the choices you made below were intentional and I don't have the context, if so just say so and ignore 😄 

Feel free to ping me if you'd like to discuss any of the points in detail.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+  children?: ReactNode;
+  blocks: Array<CodeSnippetBlockProps>;
+};
+
+type CodeSnippetBlockProps = {
+  code: string;
+  title?: string;
+  numbered?: boolean;
+  icon?: "linuxPrompt" | "windowsPrompt" | "url";
+  isWrapped?: boolean;
```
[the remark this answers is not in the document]
bartaz: Yes, good call, `wrapLines` is better.

src/components/CodeSnippet/CodeSnippet.stories.mdx
bartaz: This is just a simple example of how dropdown in code snippet can be used. But there may be other uses. One dropdown may affect multiple code blocks (input and output), or (with next version of Vanilla) there may be multiple dropdowns on a single code block, they may or may not affect each other. It would be really hard to implement all the possible interactions in CodeSnippet itself. Especially, that we currently don't use these dropdowns in products, so it's hard to say what real uses will be.

So I wanted to focus on creating stateless 'visual' component, that lets you render whatever code snippet you need, but leave the implementation of stateful controller to products.

But I agree, that in future we maybe should have a stateful implementation of code snippet with dropdowns functionality, once some consistent use case appears across products.

src/components/CodeSnippet/CodeSnippet.tsx
bartaz: In terms of adding an icon to a code block it was designed to have one icon, not an icon per block.
So technically (at least in a way it was implemented in Vanilla SCSS) it is a separate "special case" modifier on a block.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  if (numbered) {
+    className += "--numbered";
+
+    // wrap code lines in spans (and preserve the whitespace)
+    const lines = code.split(/\r?\n/);
+    numberedCode = lines.map((line, i) => (
+      <React.Fragment key={`p-code-snippet__line-${i}`}>
+        <span className="p-code-snippet__line">{line}</span>
+        {"\n"}
```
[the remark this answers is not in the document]
bartaz: I think using display block on lines doesn't work well with `pre` formatting. Because both blocks create new lines, and new lines in `pre` are preserved.
So, what we are trying to do here is wrap each line in span, but preserve the whitespaces (new lines).

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+import type { ReactNode } from "react";
+
+export type DropdownOptionPros = {
+  label: string;
+  value: string | number;
+  selected?: boolean;
+  disabled?: boolean;
+};
+
+export type CodeSnippetDropdownProps = {
```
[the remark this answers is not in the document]
bartaz: Fair point, I guess it was just copy/paste - I was basing it on implementation in some other component.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+  title,
+  numbered = false,
+  icon = null,
+  isWrapped = false,
+  dropdowns,
+}: CodeSnippetBlockProps): JSX.Element => {
+  let className = "p-code-snippet__block";
+  let numberedCode;
+
+  if (numbered) {
```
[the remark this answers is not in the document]
bartaz: At the first thought I didn't want to mix icons with numbers in one prop (because numbers aren't really icons, and we have only one icon per block and number per line), but I agree that because numbers and icons are exclusive it kinda makes sense to put them on same prop.

So, I thought that maybe renaming it to `appearance` (one of "numbered" "linuxPrompt" "windowsPrompt" "url") makes sense.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+import type { ReactNode } from "react";
+
+export type DropdownOptionPros = {
```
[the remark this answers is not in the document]
bartaz: LOL :) yes

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+};
+
+export const CodeSnippetBlock = ({
+  code,
+  title,
+  numbered = false,
+  icon = null,
+  isWrapped = false,
+  dropdowns,
+}: CodeSnippetBlockProps): JSX.Element => {
```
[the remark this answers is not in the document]
bartaz: I just used what was used elsewhere, I don't fully understand the difference.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+          {...props}
+        ></CodeSnippetBlock>
+      ))}
+    </div>
+  );
+};
+
+CodeSnippet.propTypes = {
+  className: PropTypes.string,
+  blocks: PropTypes.array,
```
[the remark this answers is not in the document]
bartaz: Good catch. Oh I wish the props were autogenerated from type definition...

src/components/CodeSnippet/CodeSnippet.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { shallow } from "enzyme";
```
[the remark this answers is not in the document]
bartaz: Yep, good call, I've seen `shallow` used in tests in other components mostly, but it is pretty useless for anything other than checking "does it render a thing? yes it renders a thing"...

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+export type DropdownOptionPros = {
+  label: string;
+  value: string | number;
+  selected?: boolean;
+  disabled?: boolean;
+};
+
+export type CodeSnippetDropdownProps = {
+  onChange?: (evt: React.SyntheticEvent) => void;
```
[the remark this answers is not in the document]
bartaz: Good catch, it should be required.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,143 @@
...
+  dropdowns: PropTypes.array,
+};
+
+export type Props = {
+  className?: string;
+  children?: ReactNode;
+  blocks: Array<CodeSnippetBlockProps>;
+};
+
+type CodeSnippetBlockProps = {
```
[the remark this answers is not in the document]
bartaz: Done

bartaz:
@hatched Addressed most of your points, and commented on others. Ready for another look, thanks!

hatched (approved):

src/components/CodeSnippet/CodeSnippet.tsx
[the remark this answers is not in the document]
huwshimi: We had a discussion about this morning and decided to stick with `JSX.Element`

src/components/CodeSnippet/CodeSnippet.tsx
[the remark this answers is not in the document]
huwshimi: We've had quite a few cases where we end up wrapping a react-components component in maas-ui and want to pass the props to the inner component. We've had to go back and export the props in react-components so that we don't have to redefine them in maas-ui, so I think the pattern should be to always export the type for the props on react-components.

huwshimi (commented):
> > when the component is internally built of several smaller ones, should we just keep them together in one file, or better split them into own files (still under /CodeSnippet folder)?
> 
> Your file here seems to be fine for me.
> 
> I typically deal with this on a case by case basis. If the child components are small and won't ever be used outside of the parent then I will keep in them in the same file. At some point those files will grow beyond a point where they feel comfortable to work on and I'll split it out.
> 
> I do however order them by importance from top to bottom; So your top one would be the parent that's exported and all children would be below.

We do the opposite and only ever have one component per file. A component ends up having types, prop types, a component function and a set of tests (at a minimum, it can also have generate/util functions, css etc.) and it seems hard for it not to be messy to include multiple sets of these in a file/dir.

Maybe one difference between how maas-ui and jaas-dashboard do things is that jaas-dashboard tests from the top component where as maas-ui has tests for each component.

It'd be good to standardise on one approach at some point.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,145 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+import type { HTMLProps } from "react";
+
+export type Props = {
+  className?: string;
+  blocks: Array<CodeSnippetBlockProps>;
```
huwshimi, line 8: We tend to use the CodeSnippetBlockProps[] syntax for typing arrays.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,145 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React from "react";
+import type { HTMLProps } from "react";
+
+export type Props = {
+  className?: string;
```
huwshimi, line 7: It's a good practice to allow `null` (`className?: string | null`;) for optional props in case the value is derived from another value or from a function return.

src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+CodeSnippet.propTypes = {
+  blocks: PropTypes.array.isRequired,
+  className: PropTypes.string,
+};
+
+type CodeSnippetBlockProps = {
+  code: string;
+  title?: string;
+  appearance?: "numbered" | "linuxPrompt" | "windowsPrompt" | "url";
```
huwshimi, line 32: These specific strings get used throughout the component, so instead of have to type the exact string you could use an enum here:

```typescript
export enum CodeSnippetAppearance {
  LINUX_PROMPT = "linuxPrompt",
  NUMBERED = "numbered",
  URL = "url",
  WINDOWS_PROMPT = "windowsPrompt",
}
```

Appearance would be defined as `appearance?: CodeSnippetAppearance`.

Below you would then do things like `appearance === CodeSnippetAppearance.LINUX_PROMPT`

huwshimi (approved):
Looking good! A few suggestions for you.

src/components/CodeSnippet/CodeSnippet.tsx
bartaz: Right, I thought about enums, but when I looked for example in react-components repo I've seen list list of strings used. But that's a good suggestion. And now I see that Icons use enums as well.

src/components/CodeSnippet/CodeSnippet.tsx
bartaz: Oh, I thought that making a prop optional does exactly that by default.

src/components/CodeSnippet/CodeSnippet.tsx
bartaz: Ah, sure, that's nicer, forgot about that syntax.

src/components/CodeSnippet/CodeSnippet.tsx
bartaz: I checked if you can assign `null` if prop is optional, and it seems to be working fine without explicitly adding `null` to type:

Screenshot 2021-02-09 at 08 07 20
Screenshot 2021-02-09 at 08 06 59
Screenshot 2021-02-09 at 08 06 45
Screenshot 2021-02-09 at 08 07 11

PR #382, opened 2021-01-29
Converted Modal to TS & Added optional header, close button, button row.

hatched:
## Done

- Converted Modal to TypeScript.
- Made the `title` optional.
- Made the close button optional.
- Added an optional `buttons` row.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.


src/components/Modal/Modal.stories.mdx
```diff
@@ -2,17 +2,7 @@ import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs/blocks";
 
 import Modal from "./Modal";
 
-<Meta
-  title="Modal"
-  component={Modal}
-  argTypes={{
-    children: {
-      control: {
-        disable: true,
```
huwshimi, line 11: This is required as you can't edit the children in the demo. We should also disable `buttons` as they also can't be edited.

src/components/Modal/Modal.stories.mdx
```diff
@@ -38,22 +28,15 @@ The modal component can be used to overlay an area of the screen which can conta
     args={{
       close: () => {},
       title: "Confirm delete",
+      buttons: [
+        <button onClick={() => {}}>Cancel</button>,
```
huwshimi, line 32: These need a unique 'key' prop.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
+import { nanoid } from "nanoid";
+import PropTypes from "prop-types";
+import React, { ReactElement, useRef } from "react";
+
+import "./modal.scss";
+
+type Props = {
+  buttons?: ReactElement[];
+  children?: ReactElement;
```
huwshimi, line 9: I think you want `ReactNode` here as it's more permissive (anything that can be rendered by React).

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
+import { nanoid } from "nanoid";
+import PropTypes from "prop-types";
+import React, { ReactElement, useRef } from "react";
+
+import "./modal.scss";
+
+type Props = {
+  buttons?: ReactElement[];
```
huwshimi, line 8: Do we want this to be limited to Buttons?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
...
+  children?: ReactElement;
+  close?: () => void;
+  title?: string;
```
huwshimi, line 11: It could be nice to have this as ReactNode in case someone wants to display something else in this position other than just a string (e.g. an icon or a link). I tend to try and be as flexible as possible with these types.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
+import { nanoid } from "nanoid";
+import PropTypes from "prop-types";
+import React, { ReactElement, useRef } from "react";
+
+import "./modal.scss";
+
+type Props = {
+  buttons?: ReactElement[];
+  children?: ReactElement;
+  close?: () => void;
```
huwshimi, line 10: For optional props I tend to add `| null` so that if props are defined by other code (e.g. the result of a function) then they may be `null` and then we have to define the prop as `close={onClose || undefined}` to satisfy the type.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <div className="p-modal" onClick={close}>
+      <div
+        className="p-modal__dialog"
+        role="dialog"
+        aria-labelledby={titleId.current}
+        aria-describedby={descriptionId.current}
+        onClick={(evt) => evt.stopPropagation()}
+      >
+        {title && (
```
huwshimi, line 31: Here `title` should be turned into a boolean (`!!title` or `Boolean(title)`) or should use a ternary otherwise for some falsy values `title` will be rendered (e.g. if `title` was `0` then it will display "0").

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
...
+          <header className="p-modal__header">
+            <h2 className="p-modal__title" id={titleId.current}>
+              {title}
+            </h2>
+            {close && (
```
huwshimi, line 36: Same thing with falsy values here.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
...
+                aria-label="Close active modal"
+                onClick={close}
+              >
+                Close
+              </button>
+            )}
+          </header>
+        )}
+        <div id={descriptionId.current}>{children}</div>
+        {buttons && <div className="p-modal__button-row">{buttons}</div>}
```
huwshimi, line 48: Falsy issue here too.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
...
+      </div>
+    </div>
+  );
+};
+
+Modal.propTypes = {
+  buttons: PropTypes.array,
+  children: PropTypes.node.isRequired,
```
huwshimi, line 56: This has `isRequired` but in the type it is defined as optional.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
...
+        )}
+        <div id={descriptionId.current}>{children}</div>
+        {buttons && <div className="p-modal__button-row">{buttons}</div>}
+      </div>
+    </div>
+  );
+};
+
+Modal.propTypes = {
+  buttons: PropTypes.array,
```
huwshimi, line 55: It would be nice for those not using typescript if this was an `arrayOf`.

src/components/Modal/modal.scss
```diff
@@ -0,0 +1,7 @@
+@import "~vanilla-framework/scss/settings";
+
+.p-modal__button-row {
+  border-top: 1px solid $color-mid-light;
+  padding-top: 1rem;
```
huwshimi, line 5: This could use `$spv-outer--small-scaleable`.

src/components/Modal/modal.scss
```diff
@@ -0,0 +1,7 @@
+@import "~vanilla-framework/scss/settings";
+
```
huwshimi, line 2: Could you also add this file to `src/index.scss` for those that want to include all the css?

huwshimi (approved):
Thanks Jeff! I've added a bunch of suggestions based on what we've learnt when turning these components into TypeScript.

src/components/Modal/Modal.stories.mdx
hatched: Ahh ok, I wanted to give people the ability to edit it in the demo. I can see now that it uses some weird `mdx` component structure instead of real react heh.

src/components/Modal/Modal.tsx
hatched: Good call, I suppose technically it can be any react element.

src/components/Modal/Modal.tsx
hatched: I've updated this prop to be `buttonRow` and the `PropTypes.node` so that you can pass whatever you want into this area.

src/components/Modal/Modal.tsx
hatched: updated

src/components/Modal/modal.scss
hatched: Thanks, I'm not very up to date on these :)

hatched:
Thanks @huwshimi this is ready for another look...might as well do another QA because the call signature of the component has changed slightly.

huwshimi:
@hatched looks like Storybook is having issues with the mdx file:

`Unexpected error while loading ./components/Modal/Modal.stories.mdx: Error: Module build failed (from ./node_modules/@mdx-js/loader/index.js):
SyntaxError: Unexpected token (17:0)`

hatched:
Looks like it required trailing commas, that's been added now....oops

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,61 @@
+import { nanoid } from "nanoid";
+import PropTypes from "prop-types";
+import React, { ReactElement, ReactNode, useRef } from "react";
+
+import "./modal.scss";
+
+type Props = {
+  buttonRow?: ReactNode | ReactNode[] | null;
```
huwshimi, line 8: If we're not limiting the button type then maybe this should just be `ReactNode | null;` and change the buttons in the example to use a fragment instead of passing an array (and having to worry about unique keys).

huwshimi (approved):
Thanks @hatched one last suggestion.

src/components/Modal/Modal.tsx
hatched: good call

PR #385, opened 2021-02-02
Use node-12 in docker to fix demos

bartaz:
## Done

Fixes demos by using `node-12` instead of `node`

## QA

Make sure demo is running: https://react-components-385.demos.haus/


anthonydillon (approved):
:+1:

Issue #404, opened 2021-03-18
Remove Search & Filter custom styling


bartaz:
Vanilla 2.26 upstreams styling for search and filter component, so custom styling should be removed from react components.

The Vanilla search and filter uses different class names, so the component needs to be updated.

The biggest difference is the internal use of search box. React component uses `SearchBox` component directly, while Vanilla styling is using it with updated class names (specific to search and filter component).

PR #424, opened 2021-04-01
Remove custom SearchAndFilter styling

bartaz:
## Done

- Remove custom styling from SearchAndFilter, use class names provided by Vanilla instead

Fixes: #404

### QA steps

- Check the demo of SearchAndFilter component:
  - https://react-components-424.demos.haus/?path=/story/search-and-filter--default-story
- Make sure it works and looks as expected
  - try different screen widths
  - try all the functionality: clicking on chips, typing search, overflowing chips, etc

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```


sowasred2012:
I checked this on live and it's happening there - so maybe better addressing separately, but there's a weird interaction when there's an overflow:
  - Fill the search bar with enough chips to make the overflow count appear "+3" etc
  - Click the count to make it expand and reveal the other chips and the text input
  - Type something in the text input, see the "X" appear in the top right of the component (rather than with the text input)
  - Click out of the component to make it lose focus and contract to a single line
  - The "X" remains in place, and obscures the count
  - You have to click the "X" to reveal it, but it's not clear what clicking the "X" actually did until you then click the revealed count and see that your text input is gone
search

sowasred2012 (approved):
LGTM - look & functionality seems identical

PR #427, opened 2021-04-09
Add release drafter config

bartaz:
## Done

- Adds release drafter config
- Adds release drafter GH workflow

### QA steps

- Unfortunately, needs to be merged to master to run, so no way to QA at this point (apart from comparing with configs from vanilla)


sowasred2012 (approved):
Looks the same as we have on Vanilla, so :+1:

PR #428, opened 2021-04-09
Add automatic publish to npm workflow

bartaz:
## Done

- Adds a GH workflow to publish to npm when GH release is published

## QA

No way to QA it other than trying it out with the next release once this is merged. Just a code review (comparing with Vanilla) needed.

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: 12
+          node-version: 14
       - run: yarn install
       - run: yarn build
       - run: yarn lint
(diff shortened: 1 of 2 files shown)


sowasred2012 (approved):
LGTM

.github/workflows/publish-on-release.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  publish-npm:
+    name: Publish to NPM
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 12
```
hatched, line 15: Node 14 is the current LTS.

bartaz: Oh, good catch, thanks. Seems we are still using 12 in other places, but good moment to try the upgrade.

PR #439, opened 2021-04-14
Update accordion to new HTML structure

### LINKED ISSUE #403: Update accordion component to 2.26
Vanilla 2.26 introduces updated accordion markup. React component should be updated to accommodate that.

"We deprecated the previous accordion tab patterns, `.p-accordion__tab` and `.p-accordion__tab--with-title` `.p-accordion__title`, in favour of a more accessible pattern. Please use `.p-accordion__heading .p-accordion__tab`.'

bartaz:
## Done

- Itemised list of what was changed by this PR.

Fixes #403 

### QA steps

- Review and test accordion examples:
  - https://react-components-439.demos.haus/?path=/docs/accordion--default-story
  - make sure they work as expected
  - make sure they use new HTML structure, as per https://vanillaframework.io/docs/patterns/accordion

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### DIFF
--- a/src/components/AccordionSection/AccordionSection.tsx
+++ b/src/components/AccordionSection/AccordionSection.tsx
@@ -85,7 +85,7 @@ AccordionSection.propTypes = {
   setExpanded: PropTypes.func,
   title: PropTypes.string,
   titleElement: PropTypes.oneOf(["h2", "h3", "h4", "h5", "h6"]),
-  headingLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+  headingLevel: PropTypes.oneOf([2, 3, 4, 5, 6]),
 };
 
 export default AccordionSection;
--- a/src/components/AccordionSection/AccordionSection.tsx
+++ b/src/components/AccordionSection/AccordionSection.tsx
@@ -15,6 +15,7 @@ export type Props = {
   setExpanded?: (key: string | null, title: string | null) => void;
   title?: string;
   titleElement?: AccordionHeadings;
+  headingLevel?: number;
 };
 
 const AccordionSection = ({
@@ -25,6 +26,7 @@ const AccordionSection = ({
   setExpanded,
   title,
   titleElement,
+  headingLevel = 3,
 }: Props): JSX.Element => {
   const sectionId = useRef(nanoid());
   const key = sectionKey || sectionId.current;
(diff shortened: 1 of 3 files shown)


src/components/AccordionSection/AccordionSection.tsx
```diff
@@ -29,34 +29,29 @@ const AccordionSection = ({
...
-        role="tab"
-        type="button"
-      >
-        {hasTitleElement ? (
-          <Title className="p-accordion__title">{title}</Title>
-        ) : (
-          title
-        )}
-      </button>
+      <Title role="heading" className="p-accordion__heading">
```
sowasred2012, line 36: We should include the [`aria-level` attribute](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/examples/patterns/accordion/default.html#L10) here too, otherwise the heading has no semantic place in the document heirarchy when using a `div` rather than a heading element. 

I see this component currently accepts an optional `titleElement` prop, maybe we need to require a `headingLevel` prop too? I'm not sure - we wouldn't need it (or `role="heading"`, thinking about it) if `titleElement`  _was_ passed.

sowasred2012 (changes requested):

src/components/AccordionSection/AccordionSection.tsx
bartaz: I think the title element (assuming it is an `h*` doesn't need aria-level, probably even shouldn't have it). So it's only needed for divs.

I guess we can add a prop with some default value.

src/components/AccordionSection/AccordionSection.tsx
```diff
@@ -84,6 +85,7 @@ AccordionSection.propTypes = {
   setExpanded: PropTypes.func,
   title: PropTypes.string,
   titleElement: PropTypes.oneOf(["h2", "h3", "h4", "h5", "h6"]),
+  headingLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
```
sowasred2012, line 88: The `titleElement` prop on the line above doesn't accept `h1`, and it _probably_ is reasonable that an accordion title wouldn't be the most important heading on the page
```suggestion
  headingLevel: PropTypes.oneOf([2, 3, 4, 5, 6]),
```

sowasred2012 (approved):
LGTM - just one suggestion

src/components/AccordionSection/AccordionSection.tsx
bartaz: Good catch, thanks.

Issue #259, opened 2020-09-01
Update checkbox and radio buttons


huwshimi:
Update to the latest Vanilla structure for radio buttons and checkboxes.

This could mean a new component (or two) instead of part of Input as it is now:

https://vanillaframework.io/docs/base/forms#radio-button

https://vanillaframework.io/docs/base/forms#checkbox

bartaz:
The scope should include:

- adding react equvalents of `p-checkbox` and `p-radio`
- add `is-table-header` option to checkboxes to align them in table headings

Open questions:

- should we (and can we) remove the checkbox/radio types from Input component

Issue #418, opened 2021-03-31
Add an example with indeterminate checkboxes


bartaz:
Vanilla provides styling for checkbox elements with indeterminate state. We should have an example of how to achieve that with React component.

PR #441, opened 2021-04-15
Add Checkbox and Radio components

sowasred2012:
## Done

- Added Checkbox and Radio components, as well as a `CheckableInput` wrapper component

## QA

- Visit [Checkbox docs](https://react-components-441.demos.haus/?path=/story/checkboxinput--default-story)
- See that each state looks good
- Visit [Radio docs](https://react-components-441.demos.haus/?path=/docs/radioinput--default-story)
- See that each state looks good
- See that all tests pass

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

## Fixes

Fixes: #259 & #418  .


src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import React from "react";
+
+import Radio from "./Radio";
+
+describe("Radio ", () => {
+  it("renders", () => {
+    const wrapper = shallow(
+      <Radio label="Test radio" name="Radio test"></Radio>
+    );
+    expect(wrapper).toMatchSnapshot();
```
hatched, line 11: We try to avoid using snapshots because it doesn't really test for a specific functionality. It also means that if you update anything in the component (or child components) this test will fail and require you to update and validate the new snapshot.

Might I propose updating the name of this test to something more specific like.... "Renders a radio input" and then your test can check for the input and any content you expect to be there directly.

src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,35 @@
+import PropTypes from "prop-types";
+import React, { HTMLProps } from "react";
+
+import CheckableInput from "../CheckableInput";
+
+type Props = {
+  label: string;
+  disabled?: boolean;
+  required?: boolean;
+} & HTMLProps<HTMLElement>;
```
hatched, line 10: Should `HTMLElement` be `HTMLInputElement` instead?

src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Radio = ({
```
hatched, line 12: Radio is a bit too generic, `RadioInput`

src/components/CheckableInput/CheckableInput.tsx
```diff
@@ -0,0 +1,52 @@
+import classNames from "classnames";
+import { TSFixMe } from "index";
+import { nanoid } from "nanoid";
+import PropTypes from "prop-types";
+import React, { useEffect, useRef, HTMLProps } from "react";
+
+type Props = {
+  type: string;
```
hatched, line 8: Can you define the valid types here 
```suggestion
  type: "checkbox" | "radio";
```

src/components/CheckableInput/CheckableInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+    inputRef.current.indeterminate = indeterminate;
+  }, [indeterminate]);
+
+  return (
+    <label className={classNames(`p-${type}`, { "is-required": required })}>
+      <input
+        ref={inputRef}
+        type={type}
+        aria-labelledby={inputId.current}
+        className={classNames(`p-${type}__input`)}
```
hatched, line 35: you're not really using `classNames` here :) 
```suggestion
        className={`p-${type}__input`}
```

src/components/CheckableInput/CheckableInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+  // setting this type as any for now, otherwise it causes linting errors
+  const inputRef: TSFixMe = useRef(null);
+
+  useEffect(() => {
+    inputRef.current.indeterminate = indeterminate;
+  }, [indeterminate]);
+
+  return (
+    <label className={classNames(`p-${type}`, { "is-required": required })}>
+      <input
```
hatched, line 31: How can I get the value of this input out? You might want to look at using a `forwardRef`.

How would I pass my own "input" element into here? Say something like using a `Formik` input? Without being able to do that it limits the functionality of this component.

src/components/CheckableInput/CheckableInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+  required = false,
+  indeterminate = false,
+  ...checkboxProps
+}: Props): JSX.Element => {
+  const inputId = useRef(nanoid());
+  // setting this type as any for now, otherwise it causes linting errors
+  const inputRef: TSFixMe = useRef(null);
+
+  useEffect(() => {
+    inputRef.current.indeterminate = indeterminate;
```
hatched, line 26: Can a radio button be indeterminate? If not sure this component stack makes much sense.

hatched (changes requested):
Thanks @sowasred2012 I've added a number of comments that can be applied to both components (even though they are only written once). My biggest concern with this is the ability to get information out of the input and supplying my own input for use with `Formik` or `react-hook-form`.

src/components/CheckableInput/CheckableInput.tsx
bartaz: @hatched We were loosely looking at our current `Input` component when implementing this.

For the value, it just works as a standard input - you pass a `value`, `onChange` or whatever props the input needs to create a controlled input, something along these lines?

```js
const [checked, setChecked] = useState(false);

return <Checkbox checked={checked} onChange={() => setChecked(!checked)} />
```

It's a good call about `Formik`, we would need to have a look at how it works. Does our current `Input` component support it, or you implement your own components for that?

bartaz: @hatched Having a quick look at [Formik docs](https://formik.org/docs/overview#the-gist), it seems like they are passing props to inputs, which seems to be supported by our current `Checkbox` API, or am I missing something?

Do you have any example of what you mean by passing you own "input" element?

bartaz: Ah, just found that in next example:

Screenshot 2021-04-16 at 07 09 50


@hatched You have any suggestions what is the best way to pass such custom components into other components, via props, children?

This "looks nice" from JSX perspective, but is likely too open to any other child elements:
```js
<Checkbox>
   <Field type="checkbox" />
</Checkbox>
```

This seems to look a bit strange, not sure if it would even work or is a good practice in React:

```js
<Checkbox
   input={<Field type="checkbox" />}
/>
```

huwshimi: This is how we use the react-component inputs with Formik in MAAS: https://github.com/canonical-web-and-design/maas-ui/blob/master/ui/src/app/base/components/FormikField/FormikField.tsx. At first glance it looks like the new component will continue work.

bartaz: Thanks @huwshimi, that's helpful.

src/components/CheckableInput/CheckableInput.tsx
sowasred2012: Thanks for the review, and all the pointers - this is one of my first encounters with React and TypeScript, so it's very useful. I'm gonna resolve some of these comments without comment while implementing the suggestions - just imagine me face palming and saying "of course..." :laughing:

src/components/CheckableInput/CheckableInput.tsx
sowasred2012: Just talked this through with @bartaz - I see what you mean, but we don't define `indeterminate` as a prop on `RadioInput`, so it couldn't pass it if it wanted to, and I'm not sure how we could lift that `useEffect` block up into the `CheckboxInput` component.

sowasred2012:
@hatched @bartaz I've updated with your feedback, and we've taken @huwshimi's advice regarding Formik (basically, it seems like this component would work with it as is) - would you mind taking another look?

src/components/CheckableInput/CheckableInput.tsx
hatched: Thanks for the deep dive into this! I see now that there is `{...checkboxProps}` on line 38 that will remedy the original issue I had.

hatched (approved):
Thanks for sticking with this 👍 looks great.

hatched:
QA -1 only because there are no examples in storybook for the `CheckableInput` component. Which I presume is the one you'd prefer people to use over the individual ones :)

bartaz:
> QA -1 only because there are no examples in storybook for the `CheckableInput` component. Which I presume is the one you'd prefer people to use over the individual ones :)

@hatched I think that `CheckboxInput` and `RadioInput` are the preferred components here, as they explicitly say what they are and should be easier to understand by developer using them. `CheckableInput` here is just an internal way to abstract common functionality.

sowasred2012:
> QA -1 only because there are no examples in storybook for the `CheckableInput` component. Which I presume is the one you'd prefer people to use over the individual ones :)

I took influence from the `Field`/`Input` example - `Field` doesn't have a storybook example, but is used by all of the `Input`s, so I thought we could do a similar thing here?

hatched:
Alright no problem - If we aren't actually supposed to use that component then I don't see any problem with it as it is....an undocumented API is an unused API ;)

Issue #415, opened 2021-03-31
Add dark option to ContextualMenu


bartaz:
We introduced dart variant of ContextualMenu, we should make it available in React component.

bartaz:
After the discussion in the review of #442 we decided not to implement this at this point.

The recommended way to change the theme in Vanilla is via SCSS settings.

PR #442, opened 2021-04-16
Add dark theme option to the ContextualMenu component

bartaz:
## Done

- Add `dark` prop to ContextualMenu to turn on dark theme

### QA steps

- Check dark example for contextual menu, make sure it renders as expected:
  - https://react-components-442.demos.haus/?path=/docs/contextualmenu--default-story#dark

## Fixes

Fixes: #415 

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### DIFF
--- a/src/components/ContextualMenu/ContextualMenu.tsx
+++ b/src/components/ContextualMenu/ContextualMenu.tsx
@@ -21,6 +21,7 @@ export type Props<L> = {
   closeOnEsc?: boolean;
   closeOnOutsideClick?: boolean;
   constrainPanelWidth?: boolean;
+  dark?: boolean;
   dropdownClassName?: string;
   hasToggleIcon?: boolean;
   links?: MenuLink<L>[];
@@ -95,6 +96,7 @@ const ContextualMenu = <L,>({
   closeOnEsc = true,
   closeOnOutsideClick = true,
   constrainPanelWidth,
+  dark = false,
   dropdownClassName,
   hasToggleIcon,
   links,
@@ -142,6 +144,7 @@ const ContextualMenu = <L,>({
   const labelNode = toggleLabel ? <span>{toggleLabel}</span> : null;
   const wrapperClass = classNames(className, "p-contextual-menu", {
     [`p-contextual-menu--${adjustedPosition}`]: adjustedPosition !== "right",
+    "is-dark": dark,
   });
 
   // Update the coordinates of the wrapper once it mounts to the dom. This uses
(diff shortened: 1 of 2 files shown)


sowasred2012 (changes requested):
I can't see what the reason for this is yet, given the props passed are the same as those passed in the default example, but on Firefox (it doesn't happen in Chrome, bizarrely), the ContextualMenu is initially rendered out of place and moves back to the correct place when resizing the browser:
dark

bartaz:
> I can't see what the reason for this is yet, given the props passed are the same as those passed in the default example

@sowasred2012 Interesting find. Menu seems to be rendered via react portal, so maybe the position calculation goes off when loading or something. I'll try to rewrite the example to provide some element to toggle the menu, maybe that will help.

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -142,6 +144,7 @@ const ContextualMenu = <L,>({
   const labelNode = toggleLabel ? <span>{toggleLabel}</span> : null;
   const wrapperClass = classNames(className, "p-contextual-menu", {
     [`p-contextual-menu--${adjustedPosition}`]: adjustedPosition !== "right",
+    "is-dark": dark,
```
hatched, line 147: Does Vanilla support putting `is-dark` on the body tag and have it cascade all the way down? 

If not, while this may be a one-off just to have a dark dropdown and not a 'darkmode' component, if we have to add a prop to every component to turn it into darkmode that's not going to be very scalable.

bartaz:
@sowasred2012 This seems to be a storybook issue that I can't find  a good workaround for. Practically we would need to render contextual menu hidden and make users click to open it to position correctly. But I'd rather render it visible, so it can be screenshot by percy. I'll open a separate issue to investigate it.

src/components/ContextualMenu/ContextualMenu.tsx
bartaz: I added it because I've seen it being a prop on other component. `Strip` it seems has it.

At this point themes in Vanilla are set on component level. It seems to be a design decision that not every component is light or dark.
For example, on most our sites, top navigation uses dark theme while rest of the page is light theme. We generally set this on component level, either as settings in SCSS or via `is-dark` class name.

But, your comment made me think, and I guess that you may be right that probably we don't want to directly reimplement the same approach here in react components.

If any react project wants to change the theme of components they can (and should) do it via scss settings for whole project. Let's not provide it as a props on individual components until we really think that's the way to go forward.

bartaz:
After the discussion in the review we decided not to impement this at this point.

The recommended way to change the theme in Vanilla is via SCSS settings.

PR #503, opened 2021-07-09
Migrate ActionButton to typescript

huwshimi:
## Done

- Migrate ActionButton to typescript.

## Fixes

Fixes: #208.


src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  }, [showSuccess]);
+
+  const buttonClasses = classNames(
+    className,
+    "p-action-button",
+    `p-button--${appearance}`,
+    {
+      "is-loading": showLoader,
+      "is-success": showSuccess,
+      "is-disabled": disabled,
```
Caleb-Ellis, line 92: I think we need to add the `is-processing` class here too so that the opacity doesn't change

Caleb-Ellis (approved):
LGTM! Just one comment about some percy diffs

src/components/ActionButton/ActionButton.tsx
huwshimi: Thanks, they passed after I fixed that.

PR #505, opened 2021-07-09
Migrate the Chip component to typescript

huwshimi:
## Done

- Migrate the Chip component to typescript.

## Fixes

Fixes: #212.


src/components/Chip/Chip.tsx
```diff
@@ -1,19 +1,32 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { highlightSubString } from "../../utils";
+import type { KeyboardEvent, MouseEvent } from "react";
+
+export type Props = {
+  lead?: string;
+  onClick: (
```
Caleb-Ellis, line 8: I think this might be optional

Caleb-Ellis (approved):
LGTM, at least as far as this PR is concerned. But the `onClick` type is a bit weird. Where does `cloud` come from?

src/components/Chip/Chip.tsx
```diff
@@ -1,19 +1,32 @@
...
+    event: MouseEvent<HTMLDivElement> | { lead: string; cloud: string }
```
bartaz, line 9: This event type seems weird. The lead/cloud thing seems project specific (I guess it's just a leftover from JAAS). And from implementation it seems that onClick works differently when clicking (returns event) and using keyboard (returns arbitrary object).

Probably out of scope for this migration PR, but we should have a separate issue to address this.

huwshimi: I've filed an issue https://github.com/canonical-web-and-design/react-components/issues/507.

src/components/Chip/Chip.tsx
huwshimi: Yeah I think you're right, I've updated it.

bartaz:
I think tests (snapshot) needs to be updated.

PR #506, opened 2021-07-09
Migrate Col to typescript

huwshimi:
## Done

- Migrate Col to typescript.

## Fixes

Fixes: #214.


Caleb-Ellis (approved):
LGTM!

src/components/Col/Col.tsx
```diff
@@ -1,6 +1,8 @@
 import classNames from "classnames";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { ElementType, ReactNode } from "react";
+
+export type ColSize = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
 
 export const colSizes = [
```
bartaz, line 7: @huwshimi Not sure where it is used, but this seems to list both numbers and strings. Is that expected?

huwshimi: Yeah I think that was to make it easier to not make mistakes, but now that it's typescript I think it's probably easier to just have numbers and typescript will tell you the correct type.

huwshimi: I'll update this to be numbers as well so that it's consistent.

huwshimi:
The update to number is causing issues so I'll have to update this next week.

Issue #477, opened 2021-06-08
`highlightSubString` util needs to be more resilient


hatched:
related: https://github.com/canonical-web-and-design/jaas-dashboard/issues/1013

If the `highlightSubString` util is called with invalid values then it throws and breaks the application. It should be more resilient as it's is typically a downstream change that's not easy to fix on the client side.

bartaz:
@hatched Is this related to a particular component? I'm not sure whats the context of `highlightSubString` util.

hatched:
Sorry @bartaz I was actually actively working on this fix.

PR #514, opened 2021-07-12
Graceful highlight

hatched:
## Done

A user reported that the search and filter component was causing the Juju Dashboard to crash because it was receiving invalid data. This makes the source of the problem, the `highlightSubString` function more resilient to bad data in an attempt to solve the issue.

### QA steps

- Use the search and filter UI, it should continue to work as normal

## Fixes

Fixes: #477

### DIFF
--- a/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
+++ b/src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
@@ -10,16 +10,12 @@ const sampleData = {
 
 describe("Filter panel section", () => {
   it("renders", () => {
-    const wrapper = shallow(
-      <FilterPanelSection data={sampleData} searchTerm="" />
-    );
+    const wrapper = shallow(<FilterPanelSection data={sampleData} />);
     expect(wrapper).toMatchSnapshot();
   });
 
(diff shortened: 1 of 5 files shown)


src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
```diff
@@ -10,12 +10,16 @@ const sampleData = {
 
 describe("Filter panel section", () => {
   it("renders", () => {
-    const wrapper = shallow(<FilterPanelSection data={sampleData} />);
+    const wrapper = shallow(
+      <FilterPanelSection data={sampleData} searchTerm="" />
```
anthonydillon, line 14: Why is `searchTerm` a required prop here? Can it not be set to a default value of an empty string?

anthonydillon (changes requested):
Question about the added prop to test.

src/components/SearchAndFilter/FilterPanelSection/FilterPanelSection.test.js
hatched: Updated :)

anthonydillon (approved):
LGTM :+1:

PR #515, opened 2021-07-13
Update Notification component to latest structure

Caleb-Ellis:
## Done

- Updated Notificaiton component to latest structure
- Migrated Notification component to TypeScript

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check that the stories for the Notification cover all the new functionality as outlined in [the design](https://app.zeplin.io/project/5fac06865cfff51271b72cc6/screen/60353ac1f130133007d40217) (left hand side and modal)

## Fixes

Fixes canonical-web-and-design/vanilla-framework#3818
Fixes: #224.


bartaz:
Demo: https://react-components-515.demos.haus/?path=/story/notification--information

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
...
+   */
+  inline?: boolean;
+  /**
+   * The function to run when dismissing/closing the notification.
+   */
+  onDismiss?: () => void;
+  /**
+   * The severity of the notification.
+   */
+  severity?: "caution" | "information" | "negative" | "positive";
```
bartaz, line 52: Could this be a TypeScript enum?

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
...
+  actions?: {
+    appearance?: ButtonProps["appearance"];
+    className?: string;
+    label: string;
+    onClick: () => void;
+  }[];
+  /**
+   * Whether the notification should not show a border.
+   */
+  appearance?: "borderless" | "default" | "modal" | "raised";
```
bartaz, line 32: Could this be a TypeScript enum?

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React, { useEffect, useRef } from "react";
+import type { HTMLProps, ReactNode } from "react";
+
+import Button, { ButtonAppearance } from "../Button";
+import type { Props as ButtonProps } from "../Button";
+
+export const notificationTypes = {
```
bartaz, line 9: It's repeated in the TS Props definition. Maybe this can be a TS enum?

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * The props for the Notification component.
+ */
+export type Props = {
+  /**
+   * The list of actions that the notification can perform.
+   */
+  actions?: {
+    appearance?: ButtonProps["appearance"];
```
bartaz, line 24: As discussed in stand-up, I guess we should allow for as little as possible at the beginning.

So let's limit that to `label` and `onClick` only for now. Ideally (if TS allows for that) we should only allow max 2 elements in actions array.

bartaz: Restricting size of array seems to be possible with TS tuples: https://stackoverflow.com/questions/65495285/typescript-restrict-maximum-array-length

So I guess something along those lines may work:


```ts

type NotificationAction = {
   label: string;
   onClick: () => void;
}

export type Props = {
   actions?: readonly [NotificationAction?, NotificationAction?];
   /* ... rest of the props */
}
```

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
...
+              data-test="notification-timestamp"
+            >
+              {timestamp}
+            </span>
+          )}
+          {hasActions ? (
+            <div className="p-notification__actions">
+              {actions.map((action, i) => (
+                <Button
+                  appearance={action.appearance || ButtonAppearance.LINK}
```
bartaz, line 135: Make them always `LINK` buttons for now.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
...
+      })}
+      {...props}
+    >
+      <div className="p-notification__content">
+        {title && (
+          <h5 className="p-notification__title" data-test="notification-title">
+            {title}
+          </h5>
+        )}
+        <p className="p-notification__message">{children}</p>
```
bartaz, line 109: The `inline` variant seems to depend on a space being part of the title, without it React just merges the elements together:

Screenshot 2021-07-13 at 12 53 07

We should include the space `{' '}` between title and message to make sure space doesn't need to be part of content.

src/components/Notification/Notification.stories.mdx
```diff
@@ -26,15 +37,17 @@ Notifications are used to display global information, our notification variants
 
 <ArgsTable of={Notification} />
```
bartaz, line 39: I think it would be worth documenting somewhere on this page that we deprecated old props and introduced new ones.

So we could list which props were renamed: `type` -> `severity`, `close` -> `onDismiss`, `status` -> `title`.

Or maybe we could add this to Props table (to comments?):

Screenshot 2021-07-13 at 13 01 28

But I guess it may be easily missed when placed in there.

src/components/Notification/Notification.tsx
Caleb-Ellis: It could be yes, but I chose not to so that you don't have to also import the enum if you want to use the Notification component in a TypeScript project. i.e. with an enum it would have to be `severity={NotificationSeverity.CAUTION}` and could not be `severity="caution"` even though they evaluate to the same string. The React Typescript cheatsheet actually [recommends avoiding enums](https://react-typescript-cheatsheet.netlify.app/docs/basic/troubleshooting/types/#enum-types), surprisingly.

Anyway I like how enums mean you only define the string in one place, so as a compromise I've created defined constants (`NotificationAppearance` and `NotificationSeverity`) that you can import if you want, otherwise the string will work fine too.

src/components/Notification/Notification.tsx
Caleb-Ellis: As aboce

src/components/Notification/Notification.tsx
Caleb-Ellis: Replaced with `NotificationSeverity` and used in TS Props and PropTypes

src/components/Notification/Notification.tsx
bartaz: Thanks for clarifications.

src/components/Notification/Notification.tsx
```diff
@@ -4,32 +4,41 @@ import React, { useEffect, useRef } from "react";
...
 
-export const notificationTypes = {
+import type { Values } from "types";
+
+export const NotificationAppearance = {
+  BORDERLESS: "borderless",
+  DEFAULT: "default",
+  MODAL: "modal",
+  RAISED: "raised",
+} as const;
```
bartaz, line 15: What does `export const ... as const` do? I didn't know this notation.

src/components/Notification/Notification.tsx
```diff
@@ -4,32 +4,41 @@ import React, { useEffect, useRef } from "react";
...
+
+export const NotificationSeverity = {
   CAUTION: "caution",
   INFORMATION: "information",
   NEGATIVE: "negative",
   POSITIVE: "positive",
+} as const;
+
+type NotificationAction = {
+  className?: string;
```
bartaz, line 25: Is the className needed? It practically allows to do anything you want with the button, and I'm not sure we want that.

src/components/Notification/Notification.tsx
Caleb-Ellis: The `as const` part tells the TS compiler that the object is completely static, otherwise the type inference for the values is `string` instead of very specific strings ("modal", "raised", etc.) . It's the way to get `enum`-like behaviour without having to use `enum`s.

src/components/Notification/Notification.tsx
Caleb-Ellis: No I don't think it's needed. I left it in because there have been a few occasions where being able to change the class of an internal element of a component has been handy, but this seems like another case of "only add it in if we need to".

src/components/Notification/Notification.tsx
bartaz: Thanks @Caleb-Ellis, it's a bit weird to read/write in code, but I like the concept of having strict list of strings, but also being able to pass a string as a value, not just enum.

I'll let you discuss with @huwshimi if that's the direction we want to go with values like this.

src/components/Notification/Notification.tsx
bartaz: Yes, I would think so. I don't think that we want people to mess up with notification buttons. At least for now.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,181 @@
...
+import PropTypes from "prop-types";
+import React, { useEffect, useRef } from "react";
+import type { HTMLProps, ReactNode } from "react";
+
+import Button, { ButtonAppearance } from "../Button";
+
+import type { Values } from "types";
+
+export const NotificationAppearance = {
+  BORDERLESS: "borderless",
```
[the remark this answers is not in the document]
bartaz: I published [v2.33.0-rc2](https://www.npmjs.com/package/vanilla-framework/v/2.33.0-rc2), so you can update to that to get spacing updates from Lyubo and removed variants.

src/types/index.ts
```diff
@@ -1,2 +1,3 @@
 export type Headings = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
 export type SortDirection = "none" | "ascending" | "descending";
+export type ValuesOf<T> = T[keyof T];
```
huwshimi, line 3: Maybe `ValueOf ` because it's used to get any one of the values (same as `keyof` and `typeof`).

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  useEffect(() => {
+    if (timeout && onDismiss) {
+      timeoutId.current = setTimeout(() => onDismiss(), timeout);
+    }
+    return () => clearTimeout(timeoutId.current);
+  }, [onDismiss, timeout]);
+
+  if (IS_DEV && (close || status || type)) {
+    console.warn(
```
bartaz, line 109: Nice touch!

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,197 @@
...
+      )}
+    </div>
+  );
+};
+
+Notification.propTypes = {
+  actions: PropTypes.arrayOf(
+    PropTypes.shape({
+      appearance: PropTypes.string,
+      className: PropTypes.string,
```
bartaz, line 178: We removed `appearance` and `className` from actions.

src/components/Notification/Notification.stories.mdx
```diff
@@ -43,40 +77,189 @@ Notifications are used to display global information, our notification variants
...
+      severity: "information",
+      timestamp: "1h ago",
+      title: "Title",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+### Deprecated
```
bartaz, line 253: I'd suggest moving this up, just under the generated props list.

Caleb-Ellis: :+1:

src/components/Notification/Notification.tsx
Caleb-Ellis: :+1:

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * The notification message content.
+   */
+  children?: ReactNode;
+  /**
+   * Optional class(es) to apply to the parent notification element.
+   */
+  className?: string;
+  /**
+   * Deprecated. Use `onDismiss` instead.
```
bartaz, line 44:
```suggestion
   * **Deprecated**. Use `onDismiss` instead.
```

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * A relevant timestamp for the notification, e.g. when it was created.
+   */
+  timestamp?: ReactNode;
+  /**
+   * The title of the notification.
+   */
+  title?: ReactNode;
+  /**
+   * Deprecated. Use `severity` instead.
```
bartaz, line 76:
```suggestion
   * **Deprecated**. Use `severity` instead.
```

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * The function to run when dismissing/closing the notification.
+   */
+  onDismiss?: () => void;
+  /**
+   * The severity of the notification.
+   */
+  severity?: ValueOf<typeof NotificationSeverity>;
+  /**
+   * Deprecated. Use `title` instead.
```
bartaz, line 60:
```suggestion
   * **Deprecated**. Use `title` instead.
```

bartaz (commented):
Some suggestions for props descriptions.

bartaz (approved):
LGTM, thanks!

PR #520, opened 2021-07-15
Migrate Form to TypeScript.

huwshimi:
## Done

- Migrate Form to TypeScript.

## Fixes

Fixes: #217.


src/components/Form/Form.tsx
```diff
@@ -0,0 +1,55 @@
...
+  inline: boolean;
+  stacked?: false;
+};
+
+export type StackedForm = BaseForm & {
+  stacked: boolean;
+  inline?: false;
+};
+
+export type Props = (OrderedForm | InlineForm | StackedForm) &
```
Caleb-Ellis, line 26: This seems a bit hard to parse. I wonder if it would be better to update the props if there are actually three mutually exclusive variants, e.g. `appearance: "default" | "inline" | "stacked" and do the same deprecation thing as with notifications.

huwshimi: Yeah I reckon you're right. It's the same thing (only worse) with the list component: https://github.com/canonical-web-and-design/react-components/pull/522.

I suspect as you say we should split these up into individual components, but I'm inclined to want to get these landed as TypeScript files first so that we know if we're breaking things when we split them up (and what the different types should be when we do so).

Caleb-Ellis: That's fair!

Caleb-Ellis (approved):

PR #526, opened 2021-07-15
Migrate Strip to TypeScript

huwshimi:
## Done

- Migrate Strip to TypeScript.

## Fixes

Fixes: #229.


src/components/Strip/Strip.tsx
```diff
@@ -1,26 +1,44 @@
...
+  colSize?: ColSize;
+  dark?: boolean;
+  deep?: boolean;
+  element?: ElementType;
+  includeCol?: boolean;
+  light?: boolean;
+  rowClassName?: string;
+  shallow?: boolean;
+  type?: string;
+} & HTMLProps<HTMLElement>;
```
Caleb-Ellis, line 24: Forgot about this in the other PRs but we should add descriptions for each prop

huwshimi: Yeah good idea, might have to go back over the other components. Also might be worth trying to generate the prop types once they're all migrated: https://github.com/canonical-web-and-design/react-components/issues/201.

huwshimi: I've pushed these.

Caleb-Ellis (approved):
LGTM!

Issue #312, opened 2020-10-19
Update to new JSX transform


huwshimi:
See the details here:

https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

huwshimi:
I made [an attempt at this](https://github.com/canonical-web-and-design/react-components/pull/535), but needs to be revisited when Storybook has better support for the new JSX transform.

bartaz:
This may not be blocked by Storybook anymore, but I'm not sure how much it's worth investing directly now.

We are starting to work on a possible new architecture for Design System (both styling and React components), so this could be considered to be revisited in the new architecture.

PR #535, opened 2021-07-16
Update jsx transform

huwshimi:
## Done

- Update to the new JSX transform.

## QA

N/A

## Fixes

Fixes: #312.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,13 @@
 {
   "presets": [
     "@babel/preset-env",
-    "@babel/preset-react",
-    "@babel/preset-typescript"
+    "@babel/preset-typescript",
+    [
+      "@babel/preset-react",
+      {
+        "runtime": "automatic"
+      }
+    ]
   ],
   "plugins": [
     "@babel/plugin-proposal-class-properties",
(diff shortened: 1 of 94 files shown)


huwshimi:
Well I just don't know what's going on here. Some config somewhere must be wrong as it's loading duplicates of React.

`Module parse failed: Identifier 'React' has already been declared`

bartaz:
Yes, that's weird, it seems to fail on `build-docs` step, so it seems like storybook building is not working.
Fails on demo jenkins as well: https://jenkins.canonical.com/webteam/view/demos/job/start-demo/9492/console

bartaz:
Found a couple of JSX transform related bug reports in Storybook, nothing directly about this kind of error, but maybe some config recommendations would work for us:

https://github.com/storybookjs/storybook/issues/13421
https://github.com/storybookjs/storybook/issues/12952
https://github.com/storybookjs/storybook/issues/12881

cristinadresch:
@huwshimi will you have time to look at it this iteration?

huwshimi:
It looks like there might be issues with the new JSX transform and Storybook so I'll close this for now and we can try again in the future once Storybook supports it.

PR #542, opened 2021-07-21
Improve the MainTable types

huwshimi:
## Done

- Improve the MainTable types to correct some issues that were found when using with maas-ui.


sowasred2012 (approved):
LGTM

PR #552, opened 2021-07-28
Use consistent type for className prop

### LINKED ISSUE #551: className prop is typed inconsistently
In some places we have `className?: string;` and other places we have `className?: string | null`. The former causes strict mode errors in instances where we have a prop like `className={boolean === true ? "class" : null}` so we should consistently use the latter option.

Caleb-Ellis:
## Done

- Added `ClassName` type and used across all components with a `className` prop.

## QA

- N/A

## Fixes

Fixes #551


sowasred2012 (approved):
LGTM

PR #554, opened 2021-07-28
Export types from the index

### LINKED ISSUE #553: Types could be exported in a nicer way
When importing react-components types in a project you have to do something like:
``` typescript
import type { Props as ComponentProps } from "@canonical/react-components/dist/components/Component/Component";
```

It'd be nicer if you could do something like this instead:
``` typescript
import type { ComponentProps } from "@canonical/react-components";
```

Caleb-Ellis:
## Done

- Export types form the index so they can be imported nicely

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- N/A, unless you want to build and install the tarball and check that you can import types in a project like:
  ``` typescript
  import type { ButtonProps } from "@canonical/react-components";
  ```

## Fixes

Fixes #553

### DIFF
--- a/src/components/Accordion/index.ts
+++ b/src/components/Accordion/index.ts
@@ -1,1 +1,2 @@
 export { default } from "./Accordion";
+export type { Props as AccordionProps } from "./Accordion";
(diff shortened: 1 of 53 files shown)


sowasred2012 (approved):
LGTM

PR #558, opened 2021-07-29
Fix more TypeScript bugs

Caleb-Ellis:
## Done

- Fixed some more TypeScript issues that I've encountered while trying to upgrade to v0.19.0

### QA steps

- N/A

### DIFF
--- a/src/components/ActionButton/ActionButton.tsx
+++ b/src/components/ActionButton/ActionButton.tsx
@@ -119,6 +119,7 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
+      disabled={disabled}
       ref={ref}
       style={
         height && width
--- a/src/components/Form/Form.tsx
+++ b/src/components/Form/Form.tsx
@@ -4,7 +4,7 @@ import type { HTMLProps, ReactNode } from "react";
 
 import type { ClassName } from "types";
 
-export type BaseForm = {
+export type Props = {
   /**
    * The content of the form.
    */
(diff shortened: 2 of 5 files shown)


src/components/ActionButton/ActionButton.tsx
```diff
@@ -119,6 +119,7 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
+      disabled={disabled}
```
Caleb-Ellis, line 122: This needs to be explicitly defined because it's an explicitly defined `ActionButton` prop, i.e. it's not spread in `...buttonProps`

src/components/Form/Form.tsx
```diff
@@ -4,7 +4,7 @@ import type { HTMLProps, ReactNode } from "react";
 
 import type { ClassName } from "types";
 
-export type BaseForm = {
```
Caleb-Ellis, line 7: I've removed all the complicated logic for determining the conditional `Props` type because the generated `propTypes` were wrong. I think this is fine because we shouldn't really be having these weird union types anyway - they should be separate components. For now it'll be possible to use an "incompatible" combination of props but at least there won't be 70000 console errors.

src/components/List/List.tsx
```diff
@@ -1,88 +1,40 @@
...
 
 export type SteppedListItem = {
   content: ReactNode;
   title: ReactNode;
   titleElement?: Headings;
-} & HTMLAttributes<HTMLLIElement>;
+} & HTMLProps<HTMLLIElement>;
 
-export type BaseProps = {
+export type Props = {
```
Caleb-Ellis, line 19: Same as above

src/components/Notification/Notification.tsx
```diff
@@ -26,7 +26,7 @@ export type Props = {
   /**
    * A list of up to two actions that the notification can perform.
    */
-  actions?: readonly [NotificationAction?, NotificationAction?];
+  actions?: NotificationAction[];
```
Caleb-Ellis, line 29: Restricting the number of actions to 2 was causing some issues in places where we wanted to push to an existing array, e.g.
``` typescript
const actions: NotificationProps["actions"] = [action];
if (condition) {
  actions.push(extraAction); // "push" does not exist on [NotificationAction?, NotificationAction?]
}
```

Similarly, with concat:
``` typescript
let actions: NotificationProps["actions"] = [action];
if (condition) {
  actions = actions.concat([extraAction]); // source could have length greater than 2
}
```

So to keep it simple I've removed the length restriction.

sowasred2012 (approved):
LGTM

PR #559, opened 2021-07-30
Add generic type for spreading props

Caleb-Ellis:
## Done

- Added `PropsWithSpread` type that ensures that defined component props are never overwritten by a subsequent prop spread
- Added descriptions to some base types

### QA steps

- Pull this branch and open `SearchBox.tsx` in VSCode
- Hover over the props in the component function argument and check that the `onChange`, `onSubmit` and `value` props are typed according to the `Props` definition.

### DIFF
--- a/src/components/ActionButton/ActionButton.tsx
+++ b/src/components/ActionButton/ActionButton.tsx
@@ -2,44 +2,48 @@ import classNames from "classnames";
 import React, { useEffect, useRef, useState } from "react";
 import type { ButtonHTMLAttributes, ReactNode } from "react";
 
+import type { ButtonProps } from "../Button";
+import { ButtonAppearance } from "../Button";
 import Icon from "../Icon";
-import { ButtonAppearance } from "../Button/Button";
 
-import type { ClassName, ValueOf } from "types";
+import type { ClassName, PropsWithSpread } from "types";
 
 export const LOADER_MIN_DURATION = 400; // minimium duration (ms) loader displays
 export const SUCCESS_DURATION = 2000; // duration (ms) success tick is displayed
 
-export type Props = {
-  /**
-   * The appearance of the button.
-   */
-  appearance?: ValueOf<typeof ButtonAppearance>;
-  /**
-   * The content of the button.
-   */
-  children?: ReactNode;
-  /**
-   * Optional class(es) to pass to the button element.
-   */
-  className?: ClassName;
-  /**
-   * Whether the button should be disabled.
-   */
-  disabled?: boolean;
-  /**
-   * Whether the button should display inline.
-   */
-  inline?: boolean;
-  /**
-   * Whether the button should be in the loading state.
-   */
-  loading?: boolean;
-  /**
-   * Whether the button should be in the success state.
-   */
-  success?: boolean;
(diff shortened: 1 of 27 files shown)


src/components/ActionButton/ActionButton.tsx
```diff
@@ -2,44 +2,48 @@ import classNames from "classnames";
...
-   * Whether the button should be in the success state.
-   */
-  success?: boolean;
-} & ButtonHTMLAttributes<HTMLButtonElement>;
+export type Props = PropsWithSpread<
+  {
+    /**
+     * The appearance of the button.
+     */
+    appearance?: ButtonProps["appearance"];
```
Caleb-Ellis, line 19: Just quickly changed this over to be the same as `ButtonProps` instead of redefining it here. That way there's only one place where we have to care about what appearances a button can take,

src/components/CheckableInput/CheckableInput.tsx
```diff
@@ -3,20 +3,25 @@ import { nanoid } from "nanoid";
...
+    /**
+     * The label for the input element.
+     */
+    label: ReactNode;
+    /**
+     * Whether the input element should display in indeterminate state.
+     */
+    indeterminate?: boolean;
+  },
+  Omit<HTMLProps<HTMLInputElement>, "type">
```
huwshimi, line 23: Does this still need to explicitly Omit type? If so we should maybe leave a comment as to why.

src/components/MainTable/MainTable.tsx
```diff
@@ -10,99 +10,111 @@ import TableHeader from "../TableHeader";
...
-  content?: ReactNode;
-} & Omit<TableCellProps, "children" | "content">;
+export type MainTableCell = PropsWithSpread<
+  {
+    /**
+     * The content of the table cell.
+     */
+    content?: ReactNode;
+  },
+  Omit<TableCellProps, "children">
```
huwshimi, line 38: Same here, if this still needs to be explicitly omitted then a comment might help.

src/types/index.ts
```diff
@@ -12,5 +28,14 @@ export type SubComponentProps<P> = Partial<P> & {
   // https://github.com/microsoft/TypeScript/issues/28960
   [prop: string]: unknown;
 };
+/**
+ * This type is simply an alias for the 'any' type and should be used sparingly,
+ * if at all.
+ */
 export type TSFixMe = any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
huwshimi, line 35: I actually have a feeling we don't use that anywhere anymore.

huwshimi (approved):
Nice! A couple of notes about leaving comments where we still have to use omit.

src/components/CheckableInput/CheckableInput.tsx
Caleb-Ellis: Good call!

src/types/index.ts
Caleb-Ellis: Yeah we don't actually, but I feel like we should keep it just to be safe. Might be handy for onboarding people unfamiliar with TS for example.

Issue #570, opened 2021-08-09
[Select] Options should accept null


huwshimi:
The optional [`Options` prop](https://github.com/canonical-web-and-design/react-components/blob/master/src/components/Select/Select.tsx#L56) should accept `null`.

hatched:
Why? :)

When making issues like this we should try to answer at least 2 of the W's, What and Why, to encourage discussion (if anyone feels it necessary).

huwshimi:
Ah right sorry, this is a constantly occurring issue, so I presumed (foolishly) that most people would have come across it.

When generating props (like for the options) often methods or selectors will return `null` when there are no values, however, while the prop is optional it accepts the values or undefined, but not `null` so you end up with an error:

```bash
Type 'Option[] | null' is not assignable to type 'Option[] | undefined'.
Type 'null' is not assignable to type 'Option[] | undefined'.
```

PR #575, opened 2021-08-09
Allow Select to accept null for optional props.

huwshimi:
## Done

- Allow Select to accept null for optional props.

## QA

N/A

## Fixes

Fixes: #570.


Caleb-Ellis (approved):
LGTM!

Issue #571, opened 2021-08-09
[Textarea] autoComplete prop is not accepted


huwshimi:
The Textarea component does not allow [the `autoComplete` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-autocomplete).

<Textarea autoComplete="off"></Textarea>

```
Property 'autoComplete' does not exist on type 'IntrinsicAttributes & { component?:
(({ caution, className, error, grow, help, id, label, labelClassName, onKeyUp, required,
stacked, style, success, takeFocus, wrapperClassName, ...props }: Props) => Element) |
undefined; name: string; value?: string | number | readonly string[] | undefined; } & {
caution?: ReactNo...'.
```

PR #576, opened 2021-08-09
Allow input attributes on Textarea

huwshimi:
## Done

- Allow input attributes on Textarea.

## QA

N/A

## Fixes

Fixes: #571.

### DIFF
--- a/src/components/Textarea/Textarea.tsx
+++ b/src/components/Textarea/Textarea.tsx
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import React, { useEffect, useRef } from "react";
-import type { InputHTMLAttributes, ReactNode } from "react";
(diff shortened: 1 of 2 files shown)


src/components/Textarea/Textarea.tsx
```diff
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import React, { useEffect, useRef } from "react";
-import type { HTMLAttributes, ReactNode } from "react";
+import type { InputHTMLAttributes, ReactNode } from "react";
```
Caleb-Ellis, line 3: Looks like there's also `TextareaHTMLAttributes` which seems like it would make more sense?

huwshimi: Oh nice thanks, I even looked!

Caleb-Ellis (approved):
LGTM!

PR #579, opened 2021-08-17
Allow tooltip messages that contain HTML

huwshimi:
## Done

- Allow tooltip messages that contain HTML.

## QA

N/A


bartaz (approved):
LGTM

Issue #566, opened 2021-08-05
SearchBox can't be used inside forms


Caleb-Ellis:
The wrapping component of a `SearchBox` is currently a `<form>` which cannot be nested inside another form. We should just use a generic `<div>` for this component so it can be used anywhere.

PR #582, opened 2021-08-24
Remove form from searchbox

sowasred2012:
## Done

- Removed the form wrapping the `SearchBox` component, and replaced it with a div.
- Also removed the `onSubmit` prop, since it seems like there's no reason to have one without a form, and added a `submittable` prop in its place, that renders the submit button if it's ever needed.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Visit ?path=/docs/searchbox--external-state#submittable
- See that it looks good, and the docs make sense

## Fixes

Fixes: #566  .


src/components/SearchBox/SearchBox.stories.mdx
```diff
@@ -10,7 +10,7 @@ export const Template = (args) => <SearchBox {...args} />;
 
 This is a [React](https://reactjs.org/) component for the Vanilla [SearchBox](https://docs.vanillaframework.io/patterns/search-box/).
 
-Search boxes enable search functionality on a page and are typically used in a navigation bar.
+Search boxes enable search functionality on a page and are typically used in a navigation bar. SearchBoxes should be wrapped within a form element that handles its own submit action.
```
hatched, line 13:
```suggestion
Search boxes enable search functionality on a page and are typically used in a navigation bar. SearchBox elements should be wrapped within a form element that handles its own submit action.
```

src/components/SearchBox/SearchBox.test.tsx
```diff
@@ -5,18 +5,23 @@ import SearchBox from "./SearchBox";
 
 describe("SearchBox ", () => {
   it("renders", () => {
-    const wrapper = shallow(<SearchBox />);
+    const wrapper = shallow(<SearchBox onSearch={jest.fn()} />);
```
hatched, line 8: Can you add a test to check that `onSearch` is actually called when you expect it to?

src/components/SearchBox/SearchBox.test.tsx
```diff
@@ -5,18 +5,23 @@ import SearchBox from "./SearchBox";
...
     expect(wrapper.find(".p-search-box__reset").exists()).toBe(true);
   });
 
   it("can externally control the value", () => {
     const wrapper = shallow(
-      <SearchBox externallyControlled onChange={jest.fn()} value="admin" />
+      <SearchBox
+        externallyControlled
+        onSearch={jest.fn()}
+        onChange={jest.fn()}
```
hatched, line 22: There doesn't appear to be a test that checks that this method is called when we expect it to.

hatched (changes requested):
Looks good thanks, I just have a couple requests for some more tests.

src/components/SearchBox/SearchBox.tsx
```diff
@@ -1,119 +1,106 @@
...
         </button>
-      </form>
-    );
-  }
-);
+      )}
+      <button
+        className="p-search-box__button"
+        disabled={disabled}
+        type="submit"
```
bartaz, line 96: Does it need to be `submit` button? It's not in the form now, and it may cause cases when this button submits forms it's not supposed to submit (when used as part of some other bigger form).

src/components/SearchBox/SearchBox.tsx
```diff
@@ -1,119 +1,105 @@
...
+   */
+  externallyControlled?: boolean;
+  /**
+   * A function that will be called when the input value changes.
+   */
+  onChange?: (inputValue: string) => void;
+  /**
+   * A function that is called when the user clicks the search icon
+   */
+  onSearch: () => void;
```
bartaz, line 32: `onSearch` is marked as required here, is this expected?

If so, most of examples in storybook don't have it defined.

sowasred2012: I've made `onSearch` prop optional, I thought it should be required, but maybe it's not necessary.

Issue #189, opened 2020-07-23
ActionButton should extend Button


squidsoup:
Currently `ActionButton` and `Button` are completely distinct components. It probably makes sense however for `ActionButton` to extend `Button`.

bartaz:
Can be revisited during migration to new architecture.

PR #589, opened 2021-09-08
Extend Button to ActionButton

carkod:
## Done

Extend Button to ActionButton

### Storybook

Check rendered `ActionButton` https://react-components-589.demos.haus/?path=/story/actionbutton--default-story

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/189

### DIFF
--- a/src/components/ActionButton/ActionButton.stories.mdx
+++ b/src/components/ActionButton/ActionButton.stories.mdx
@@ -10,9 +10,15 @@ import ActionButton from "./ActionButton";
       control: {
         type: "text",
       },
+      defaultValue: "Click me!",
     },
+    appearance: {
+      defaultValue: "positive"
+    }
   }}
 />
+
+export const Template = (args) => <ActionButton {...args} />;
 
 ### ActionButton
 
@@ -27,9 +33,7 @@ ActionButton accepts the props from [Button](?path=/docs/button--base) in additi
 ### Default
 
 <Canvas>
-  <Story name="Default">
-    <ActionButton appearance="positive">Click me!</ActionButton>
-  </Story>
+  <Story name="Base">{Template.bind({})}</Story>
 </Canvas>
 
 ### Loading
--- a/src/components/ActionButton/ActionButton.tsx
+++ b/src/components/ActionButton/ActionButton.tsx
@@ -3,7 +3,7 @@ import React, { useEffect, useRef, useState } from "react";
 import type { ButtonHTMLAttributes, ReactNode } from "react";
 
 import type { ButtonProps } from "../Button";
-import { ButtonAppearance } from "../Button";
+import Button, { ButtonAppearance } from "../Button";
 import Icon from "../Icon";
 
 import type { ClassName, PropsWithSpread } from "types";
@@ -121,7 +121,7 @@ const ActionButton = ({
   // forwardRef which is not currently supported by components that use
   // typescript generics.
   return (
-    <button
+    <Button
       className={buttonClasses}
       disabled={disabled}
       ref={ref}
@@ -144,7 +144,7 @@ const ActionButton = ({
       ) : (
         children
       )}
-    </button>
+    </Button>
   );
 };
 
--- a/src/components/ActionButton/__snapshots__/ActionButton.test.tsx.snap
+++ b/src/components/ActionButton/__snapshots__/ActionButton.test.tsx.snap
@@ -4,20 +4,25 @@ exports[`ActionButton matches loading snapshot 1`] = `
 <ActionButton
   loading={true}
 >
-  <button
+  <Button
     className="p-action-button p-button--neutral is-processing"
     disabled={false}
   >
-    <Icon
-      className="u-animation--spin"
-      light={false}
-      name="spinner"
+    <button
+      className="p-button--neutral p-action-button p-button--neutral is-processing"
+      disabled={false}
     >
-      <i
-        className="u-animation--spin p-icon--spinner"
-      />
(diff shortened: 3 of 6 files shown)


hatched:
I understand this is still a WIP but can you create an issue which explains why you're making this change and attach it to this PR plz.

hatched:
You can ignore my previous comment, I totally missed the link 😆

carkod:
> You can ignore my previous comment, I totally missed the link

Yeah sorry, Cristina just connected the issue

carkod:
The `ActionButton` is using refs to get the width and height of the element, and this is not working because we need to forwardRef to the `Button` parent component, but the `forwardRef` is not working with the custom type, and I am not sure how to add this to the type template or maybe it can't be added.

bartaz:
@Caleb-Ellis @huwshimi Carlos started working on this, and we also just noticed a comment in code:

https://github.com/canonical-web-and-design/react-components/blob/81315bd8a74799ad62030c2933cf8f86df953330/src/components/ActionButton/ActionButton.tsx#L119-L122

Does this mean this issue is blocked? Can `Button` component be updated in a way that will allow using `forwardRef` with it?

Caleb-Ellis:
It seems like there are some options for using generics with `forwardRef` e.g. https://fettblog.eu/typescript-react-generic-forward-refs/. I'd wait for @huwshimi's response though since he did the `ActionButton` TypeScript conversion and might have some more insight.

My two cents are to leave this for now - I imagine it'll be quite a large structural change to get this to work (maybe not in terms of LOC but in terms of our approach to generics in react-components).

carkod:
> It seems like there are some options for using generics with `forwardRef` e.g. https://fettblog.eu/typescript-react-generic-forward-refs/. I'd wait for @huwshimi's response though since he did the `ActionButton` TypeScript conversion and might have some more insight.
> 
> My two cents are to leave this for now - I imagine it'll be quite a large structural change to get this to work (maybe not in terms of LOC but in terms of our approach to generics in react-components).
> 
> EDIT: Sorry, I should say - don't feel the need to get this done, but if you can come up with a nice solution I'd be happy to see it merged. :) Basically, the PR should solve the following:
> 
> 1. Have `ActionButton` props be a strict superset of `Button` props, including the generic `P`.
> 2. Be able to pass a `ref` from `ActionButton` down to `Button` while still maintaining the type safety when using `P`.

Oh that was the first article I read when I was trying to add `forwardRef` to `Button`. But in the article they suggest:
```
const Button = React.forwardRef((props, ref) => (
  <button type="button" {...props} ref={ref}>
    {props.children}
  </button>
));
```
Which in our case I can't do:

sea-of-red

Not sure, I suppose the custom generic type (`export type Props<P = null>..`) needs to be extended or changed somehow?

hatched:
Sometimes it's worth taking a step back and re-evaluating the complexity of the feature. 
  - Should `Button` ever be anything other than a `button`? 
  - Is the generic necessary?

Often times it's not worth over complicating one component when two would do just fine.

Caleb-Ellis:
> Sometimes it's worth taking a step back and re-evaluating the complexity of the feature.
> 
> * Should `Button` ever be anything other than a `button`?
> * Is the generic necessary?
> 
> Often times it's not worth over complicating one component when two would do just fine.

I was strongly considering removing the generic as well, but stopped short of recommending it. In MAAS there is literally only one place where we use Button as an anchor tag, and we don’t even use a generic there for type-safety.

Still, I think there is some use for Button to be dynamic like that. But I can’t imagine any instance in which we want ActionButton to not be a button, so it might actually make sense for them to be entirely separate components. Maybe we could just split out the common props and share those?

bartaz:
> Still, I think there is some use for Button to be dynamic like that. But I can’t imagine any instance in which we want ActionButton to not be a button, so it might actually make sense for them to be entirely separate components. Maybe we could just split out the common props and share those?

Yes, I guess the main purpose behind making ActionButton reuse the button is to make sure they share the same props. Namely - ActionButton should allow all the styling available on the Button (all the color variants, dense, small, whatnot). The problem is that if we keep them separate, anytime we change something in Vanilla button we need to remember to change it in two separate places.

huwshimi:
I think the Button was made to take a generic so that we could support navigation events with a Button. For instance the ContextualMenu component can [take a generic for the button](https://github.com/canonical-web-and-design/react-components/blob/master/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx#L14) so that we can pass [react-router Link components](https://reactrouter.com/web/api/Link).

As far as adding a ref to the Button goes I'd be tempted to use a custom ref prop e.g. `innerRef` or something, like the [second option](https://fettblog.eu/typescript-react-generic-forward-refs/#option-2%3A-create-a-custom-ref-%2F-the-wrapper-component) in the article linked to above. While this introduces a non-standard prop for refs it is easy to work with and does not require significant changes to the Button component. That prop can also be easily deprecated in the future if a built-in way of handling generics ever becomes available.

huwshimi:
Just to add to this because I happened to see the same pattern in use in Formik today, they use the `innerRef` pattern for their components: https://formik.org/docs/api/field#innerref.

ClementChaumel:
Hey @carkod and maybe @huwshimi Do you have any context on this?
Is this still relevant, should it be investigated or should we just close it? Thanks!

carkod:
> #189

Well, I guess the context is this https://github.com/canonical/react-components/issues/189. But if it's not required anymore, then I guess we can close it

PR #592, opened 2021-09-09
Add social icons to Icon component

carkod:
## Done

Add social icons to Icon component

## QA

Check on Storybook that social icons appear https://react-components-592.demos.haus?path=/docs/icon--default-story
Check percy does not show any broken elements.

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/422

### DIFF
--- a/src/components/Icon/Icon.stories.mdx
+++ b/src/components/Icon/Icon.stories.mdx
@@ -6,6 +6,8 @@ import "./Icon.stories.scss";
 
 <Meta title="Icon" component={Icon} />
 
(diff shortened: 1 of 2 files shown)


carkod:
So I fixed a couple of issues now, I realized that the args need to be added per example, so that it doesn't apply to all of them.
Also there is some icon size differences, but it happens in Vanilla framework too, social icons are larger than usual icons.

bartaz (approved):
LGTM, thanks!

Issue #588, opened 2021-09-06
Password toggle component


sowasred2012:
Snapcraft stores admin has a need for the [password toggle pattern](https://vanillaframework.io/docs/examples/patterns/forms/password-toggle) as a component.

PR #600, opened 2021-09-20
Password toggle

sowasred2012:
Co-authored-by: Peter French <peterfrench94@gmail.com>

## Done

- Added PasswordToggle component

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Visit http://localhost:8403/?path=/docs/passwordtoggle--default-story
- See that the examples and the docs make sense
- See that the tests pass

## Fixes

Fixes: #588  .

### DIFF
--- a/src/components/PasswordToggle/PasswordToggle.stories.mdx
+++ b/src/components/PasswordToggle/PasswordToggle.stories.mdx
@@ -8,6 +8,8 @@ import PasswordToggle from "./PasswordToggle";
 
 This is a [React](https://reactjs.org/) component for the Vanilla [Password Toggle](https://docs.vanillaframework.io/base/forms/#password-toggle).
 
+It can be used when an input needs to obscure its value, whilst giving the user a way to reveal it if needed.
+
 ### Props
 
 <ArgsTable of={PasswordToggle} />
--- a/src/components/PasswordToggle/PasswordToggle.stories.mdx
+++ b/src/components/PasswordToggle/PasswordToggle.stories.mdx
@@ -1,0 +1,98 @@
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs/blocks";
+
+import PasswordToggle from "./PasswordToggle";
+
+<Meta title="PasswordToggle" component={PasswordToggle} />
+
+### PasswordToggle
+
+This is a [React](https://reactjs.org/) component for the Vanilla [Password Toggle](https://docs.vanillaframework.io/base/forms/#password-toggle).
+
+### Props
+
+<ArgsTable of={PasswordToggle} />
+
+### Default
+
+<Canvas>
+  <Story name="Default">
+    <PasswordToggle id="examplePassword1" label="Password" />
+  </Story>
+</Canvas>
+
+### Read only
+
+<Canvas>
+  <Story name="Read only">
+    <PasswordToggle
+      id="examplePassword2"
+      label="Password"
+      readOnly
+      value="password"
+    />
+  </Story>
+</Canvas>
+
+### Error
+
+<Canvas>
+  <Story name="Error">
+    <PasswordToggle
+      error="This field is required."
+      id="examplePassword3"
+      label="Password"
+      value="password"
+    />
+  </Story>
+</Canvas>
+
+### Success
+
+<Canvas>
+  <Story name="Success">
+    <PasswordToggle
+      success="Strong password"
+      id="examplePassword3"
+      label="Password"
+      value="password"
+    />
+  </Story>
+</Canvas>
+
+### Caution
+
+<Canvas>
+  <Story name="Caution">
+    <PasswordToggle
+      caution="Weak password"
+      id="examplePassword4"
+      label="Password"
+      value="password"
+    />
+  </Story>
+</Canvas>
+
+### Help
+
+<Canvas>
+  <Story name="Help">
+    <PasswordToggle
+      help="Enter a secure password"
+      id="examplePassword5"
+      label="Password"
+    />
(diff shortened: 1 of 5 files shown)


src/components/PasswordToggle/PasswordToggle.tsx
```diff
@@ -0,0 +1,135 @@
...
+     */
+    required?: boolean;
+    /**
+     * The content for success validation.
+     */
+    success?: ReactNode;
+    /**
+     * Whether to focus on the input on initial render.
+     */
+    takeFocus?: boolean;
```
hatched, line 55: I'm not sure about this... 

Supporting a `forwardRef` might be more appropriate. It would enable this functionality by using a more standard `inputRef.focus()` from the parent as well as any other ref-based interactions.

src/components/PasswordToggle/PasswordToggle.tsx
```diff
@@ -0,0 +1,135 @@
...
+  label,
+  readOnly,
+  required,
+  success,
+  takeFocus,
+  wrapperClassName,
+  ...inputProps
+}: Props): JSX.Element => {
+  const [fieldType, setFieldType] = useState("password");
+  const isPassword = fieldType === "password";
```
hatched, line 79: It looks like the value of `isPassword` is more important than `fieldType` itself (used 3 vs 1). I don't feel strongly about this but it feels like maybe you should instead use the `isPassword` value as the canonical value and then determine the field type from that. (imho it would read a bit nicer)

src/components/PasswordToggle/PasswordToggle.stories.mdx
```diff
@@ -0,0 +1,100 @@
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs/blocks";
```
hatched, line 1: Great list of examples 👍

src/components/PasswordToggle/PasswordToggle.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    expect(className.includes("extra-class")).toBe(true);
+  });
+
+  it("can take focus on first render", () => {
+    const container = document.createElement("div");
+    document.body.appendChild(container);
+    const wrapper = mount(<PasswordToggle id="test-id" takeFocus />, {
+      attachTo: container,
+    });
+    expect(wrapper.find("input").getDOMNode()).toBe(document.activeElement);
```
hatched, line 27: nice

src/components/PasswordToggle/PasswordToggle.tsx
```diff
@@ -0,0 +1,135 @@
...
+  error,
+  help,
+  id,
+  label,
+  readOnly,
+  required,
+  success,
+  takeFocus,
+  wrapperClassName,
+  ...inputProps
```
hatched, line 76: We've had issues with this spread in the past, can you add a test that input props get passed to the input itself.

src/components/PasswordToggle/PasswordToggle.test.tsx
```diff
@@ -0,0 +1,36 @@
+import { mount, shallow } from "enzyme";
+import React from "react";
+
+import PasswordToggle from "./PasswordToggle";
+
+describe("PasswordToggle ", () => {
+  it("renders", () => {
```
hatched, line 7: I'm -0 on these types of tests. As _any_ change to make to the component will make the tests fail. If the component doesn't render, you'll know in the first real test below.

hatched (changes requested):

src/components/PasswordToggle/PasswordToggle.test.tsx
bartaz: We add them, because we see them in other components so it seems it is a default part of the test suit.
Personally I also wonder about usefulness of them, because I've never seen them fail to reveal any issue. They fail when you change component and you expect it to change.

Seems like we should have a team-wide decision if and when to use snapshot tests. And if we don't want them, remove them. Because right now the seem like standard in react-component repo.

src/components/PasswordToggle/PasswordToggle.tsx
bartaz: I believe they based it on existing `Input` component to keep the props consistent: https://canonical-web-and-design.github.io/react-components/?path=/docs/input--text-input

If we want to change the approach to use of refs we should do it in both places I guess.

react-components repo is kind of museum of how our React experience evolved, so when adding new stuff it's always a compromise between being consistent with things that are already there and using newer/better approaches 😄

src/components/PasswordToggle/PasswordToggle.test.tsx
huwshimi: I suspect the mostly exist from when these components existed in MAAS. Since then we've largely moved away from snapshot testing in MAAS.

src/components/PasswordToggle/PasswordToggle.tsx
sowasred2012: I've made a change that I _think_ addresses this, take a look - it certainly seems a bit nicer to me anyway.

src/components/PasswordToggle/PasswordToggle.tsx
sowasred2012: @hatched We've just updated it, and it works, but we don't _really_ understand why - when you're around later we'd love a masterclass!

hatched (approved):
The updates look great, thx!

Issue #639, opened 2021-11-17
input error is not indicated for assistive technologies


petermakowski:
Add `aria-invalid=true` to Input when there's an error to indicate the error state to assistive technologies.

PR #640, opened 2021-11-17
add aria-invalid to Input when there's an error

petermakowski:
## Done

- add `aria-invalid="true"` to Input when there's an error
  - add tests

## Before
image

## After
image

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: #639 .

### DIFF
--- a/src/components/Input/Input.test.tsx
+++ b/src/components/Input/Input.test.tsx
@@ -32,4 +32,14 @@ describe("Input ", () => {
     const wrapper = mount(<Input takeFocus />, { attachTo: container });
     expect(wrapper.find("input").getDOMNode()).toBe(document.activeElement);
   });
+
+  it("sets aria-invalid to false when there is no error", () => {
+    const wrapper = mount(<Input type="text" />);
+    expect(wrapper.find("input").prop("aria-invalid")).toBe(false);
+  });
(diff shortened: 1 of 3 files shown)


bartaz:
Thanks, but I think we should do this on Vanilla examples as well and document it there in accessibility docs, otherwise we will lose track of accessibility features implemented in react, but not documented anywhere else in Vanilla.

petermakowski:
> Thanks, but I think we should do this on Vanilla examples as well and document it there in accessibility docs, otherwise we will lose track of accessibility features implemented in react, but not documented anywhere else in Vanilla.

Okay, I'll look into this. 👍

petermakowski:
> Thanks, but I think we should do this on Vanilla examples as well and document it there in accessibility docs, otherwise we will lose track of accessibility features implemented in react, but not documented anywhere else in Vanilla.

@bartaz  Added https://github.com/canonical-web-and-design/vanilla-framework/pull/4125

bartaz (approved):
LGTM, thanks!

Issue #663, opened 2021-11-30
PasswordToggle component submits form


steverydz:
If the `PasswordToggle` component is in a form, using the show/hide button submits the form. I think adding `type="button"` to it should fix the problem.

PR #671, opened 2021-12-06
Prevent password toggle component submitting form

bethcollins92:
## Done

- Add new `type` prop to password toggle component
- Set default `type=button` to prevent the button submitting form

### Storybook

```shell
yarn start
```

### QA steps

- Add the`<PasswordToggle>` component to the <Form> and don't include a type. Check the form doesn't submit when the password toggle button is used. 

## Fixes

Fixes: #663

### DIFF
--- a/src/components/PasswordToggle/PasswordToggle.tsx
+++ b/src/components/PasswordToggle/PasswordToggle.tsx
@@ -50,6 +50,10 @@ export type Props = PropsWithSpread<
      */
     success?: ReactNode;
     /**
+     * The content for success validation.
+     */
+    type?: "submit" | "reset" | "button";
+    /**
      * Optional class(es) to pass to the wrapping Field component
      */
     wrapperClassName?: string;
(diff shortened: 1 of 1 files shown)


steverydz (approved):
LGTM 👍

src/components/PasswordToggle/PasswordToggle.tsx
```diff
@@ -49,6 +49,10 @@ export type Props = PropsWithSpread<
      * The content for success validation.
      */
     success?: ReactNode;
+    /**
+     * The content for success validation.
+     */
+    type?: string;
```
bartaz, line 55: Would be nice to limit that to just the allowed types ("button", "submit", "reset"):
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

bartaz:
@bethcollins92 Could you rebase this branch, I think the name of required check has changed due to node version update.

src/components/PasswordToggle/PasswordToggle.tsx
bethcollins92: Good idea! done

Issue #638, opened 2021-11-17
`required` is not passed down to input 


petermakowski:
- Add `required` attribute to `input` element
  - it's destructured and only passed down to `Field` component - and ommited in `input` 
https://github.com/canonical-web-and-design/react-components/blob/547dc1cf78f32e3c6af28db9ae530b33df1bf550/src/components/Input/Input.tsx#L101-L110

PR #672, opened 2021-12-06
Pass required to input element

bethcollins92:
## Done

- Pass `required` to input element (currently it's only in the field component).

### Storybook

```shell
yarn start
```

### QA steps

- Check required exits on `input` element

## Fixes

Fixes: #638  .


petermakowski (approved):
LGTM, but would be nice to see a test for this (react testing library).

PR #676, opened 2021-12-08
Remove checkbox and radio input examples

bethcollins92:
## Done

- Remove `type="radio` and `type="checkbox"` from the Input examples
- Remove checkbox from default form
- Add `type?: string;` to type prop. Not sure how to make sure `checkbox` and `radio` inputs aren't used. Should this component support every other type of input apart from those two? 

### Storybook

```shell
yarn start
```

### QA steps

- Check docs and make sure radio and checkbox examples still work as expected

## Fixes

Fixes: [#1169](https://github.com/canonical-web-and-design/vanilla-squad/issues/1169) .


src/components/Input/Input.tsx
```diff
@@ -55,6 +55,10 @@ export type Props = PropsWithSpread<
      * Whether to focus on the input on initial render.
      */
     takeFocus?: boolean;
+    /**
+     * Whether to focus on the input on initial render.
+     */
+    type?: string;
```
hatched, line 61: > Not sure how to make sure checkbox and radio inputs aren't used. Should this component support every other type of input apart from those two?

If you'd like you can define the allowed values here like...
```suggestion
    type?: "text" | "password";
```

src/components/Input/Input.tsx
```diff
@@ -55,6 +55,10 @@ export type Props = PropsWithSpread<
      * Whether to focus on the input on initial render.
      */
     takeFocus?: boolean;
+    /**
+     * Whether to focus on the input on initial render.
```
hatched, line 59: copy/paste error :)

src/components/Input/Input.tsx
bethcollins92: Thanks, I saw that but wasn't sure which types of input this component should support? as if it's all of them apart from checkbox and input that would be quite a long list

src/components/Input/Input.tsx
bethcollins92: duh

src/components/Input/Input.tsx
bartaz: This made me think - maybe instead of removing those examples, we should make the `Input` component render checkboxes/radios correctly.

So it would accept all input types, and render a standard `<input>` for them, but if the type is set to checkbox or radio it would use correct CheckboxInput / RadioInput components internally?

bethcollins92: Yeah this sounds like a better idea. I imagine `<Input type="checkbox/radio">` is used in places and would break if we removed support for it. I'll take a look at this now

bethcollins92:
@bartaz I made a shot at including the radio and checkbox components in the input - any feedback welcome!

src/components/Input/Input.tsx
```diff
@@ -92,7 +114,23 @@ const Input = ({
     }
   }, [takeFocus]);
 
-  return (
+  return type === "checkbox" ? (
```
hatched, line 117: Try to avoid nesting ternaries as it makes it hard to follow.

I think you'll also want to spread the props on these new checkbox and radio inputs as well.

hatched (changes requested):

src/components/Input/Input.tsx
bethcollins92: Thanks @hatched - refactored. The tests are failing on this `labelFirst` variable, now we have included the checkbox and radio components I'm not sure we need this still - or am I missing something?

src/components/Input/Input.tsx
```diff
@@ -88,31 +114,56 @@ const Input = ({
...
+        forId={id}
+        help={help}
+        label={label}
+        labelClassName={labelClassName}
+        labelFirst={labelFirst}
+        required={required}
+        stacked={stacked}
+        success={success}
+      >
+        <input
```
bartaz, line 155: @bethcollins92 Initially I thought that we would put the `CheckboxInput` or `RadioInput` here inside the `Field` - now I'm not sure if that makes sense - does `Field` even support having checkbox/radio in it?

It seems that having special cases for checkbox/radio in this component only makes sense if it is inside the Field. If it's not I'm not sure if we want radio/checkbox in this component (if it will just be rendering CheckboxInput/RadioInput without the Field).

src/components/Form/Form.stories.mdx
```diff
@@ -49,11 +49,6 @@ Form controls have global styling defined at the HTML element level. Labels and
             label="Password"
           />
           <Input type="file" id="exampleInputFile2" label="File input" />
-          <Input
```
bartaz, line 52: We made the `Input` component work with checkboxes/radios, so we can keep this part of example, right?

bartaz (approved):
LGTM, just a suggestion to keep the (now updated) checkbox input in `Form` example

Issue #679, opened 2021-12-14
Navigation misaligned in 3.0


bethcollins92:
See [percy](https://percy.io/vanilla-framework/react-components/builds/14528329/changed/820903811?browser=firefox&browser_ids=19&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1280&widths=1280)

PR #680, opened 2021-12-14
Fix extra padding in search box nav example

bethcollins92:
## Done

- The logo was using `p-navigation__link` instead of `p-navigation__item`. This fixes the padding.

### Storybook

```shell
yarn start
```

### QA steps

- Check this [demo](https://react-components-656.demos.haus/?path=/story/searchbox--navigation) which is using vanilla 3.0, check that changing the classname fixes the padding 

## Fixes

Fixes: #679


bartaz (approved):
Thanks

Issue #682, opened 2021-12-14
Label broken - using old class name


bethcollins92:
see [here](https://percy.io/vanilla-framework/react-components/builds/14528329/changed/820903553?browser=firefox&browser_ids=19&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=1280)

PR #683, opened 2021-12-14
Replace p-label--new

bethcollins92:
## Done

- Replace `p-label--new` with `p-label--positive`

### Storybook

```shell
yarn start
```

### QA steps

- See [here](https://react-components-656.demos.haus/?path=/story/list--stretch), the demo of 3.0 shows the label is broken. Change the classname to `p-label--positive` and see the correct label

## Fixes

Fixes: #682  .

### DIFF
--- a/src/components/List/List.stories.mdx
+++ b/src/components/List/List.stories.mdx
@@ -115,7 +115,7 @@ If you want to display lists in a way that is more visually distinctive than the
       items={[
         { content: <h3>Documentation</h3> },
         {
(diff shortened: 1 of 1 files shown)


bartaz:
This will not work until we upgrade to 3.0, as that's when we introduced new class names (and removed the old ones).

bartaz:
Let's merge it to 3.0 branch, it makes sense there.

bartaz (approved):
Will work in 3.0 branch.

PR #690, opened 2022-01-14
Prepare-0.30.0

bartaz:
## Done

Prepare release 0.30.0 that updates to Vanilla 3.0.

We bump version to 0.30.0 instead of 0.25 to kind of signify the breaking change of upgrading to Vanilla 3.0.

## QA

Make sure demo works as expected.

Most of the files for code review are autogenerated by storybook.


bethcollins92 (approved):
version is updated and demo LGTM. So I'm guessing we still need to add expanding search/coloured chips examples to react components?

Issue #691, opened 2022-01-17
Add coloured Chips


bartaz:
Vanilla 3.0 added coloured variants of Chips component. This is not yet implemented in react-components.

React `Chip` component should have a `appearance` prop with values allowing to choose desired color (as in Vanilla, with default 'non-coloured' variant).

PR #692, opened 2022-01-19
Added colored chip

ilayda21:
## Done

- Added colored chips.

## QA

- Check rendered [Chips](https://react-components-692.demos.haus/?path=/docs/chip--default-story) component.
- Run Chip tests.

## Fixes

Fixes: #691

### DIFF
--- a/src/components/Chip/Chip.stories.mdx
+++ b/src/components/Chip/Chip.stories.mdx
@@ -39,25 +39,10 @@ It can be used to display a small value attached to a component.
 
 ### Appearance
 <Canvas>
-  <Story name="Positive-appearance">
+  <Story name="Appearance">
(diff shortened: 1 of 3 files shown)


src/components/Chip/Chip.stories.mdx
```diff
@@ -37,6 +37,31 @@ It can be used to display a small value attached to a component.
   </Story>
 </Canvas>
 
+### Appearance
+<Canvas>
+  <Story name="Positive-appearance">
```
bartaz, line 42: I think that one story with all 4 coloured chips in it is enough, we don't need separate example for each of them.

ilayda21: Done :+1:

bartaz (approved):
Great, thanks!

Issue #651, opened 2021-11-22
Spread <span> props on Spinner component


Caleb-Ellis:
[maas-ui PR context](https://github.com/canonical-web-and-design/maas-ui/pull/3311)

We should spread span props on the Spinner component in order for appropriate html attributes to be accessible e.g:

``` ts
type Props = {
  // spinner prop types
} & HTMLProps<HTMLSpanElement>;

const Spinner = ({
  // spinner props
  ...spanProps
}: Props): JSX.Element => (
// etc
```

PR #700, opened 2022-02-03
Spread <span> props on Spinner component

petermakowski:
## Done

- Spread <span> props on Spinner component
- This will allow adding an aria-label

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Add an aria-label to the spinner component and verify it's added to the outer span element correctly.

## Fixes

Fixes: #651 .


bethcollins92 (approved):
LGTM - thanks @petermakowski

Issue #590, opened 2021-09-08
Spinner is missing an aria role


hatched:
It would be nice to add an aria role to the Spinner component. It would make it more accessible and also make it easier to test with RTL.

huwshimi:
We should also add a spread to the wrapping `span` element so these kinds of attributes can be added as needed by component consumers.

bartaz:
Scope:
- investigate what aria attributes to add to Spinner
- add them in react component

PR #703, opened 2022-02-08
Add aria attributes to spinner

bethcollins92:
## Done

- Added `role`, `aria-live` and `label` text to the spinner component

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check the spinner with VoiceOver

## Fixes

Fixes: #590 .

### DIFF
--- a/src/components/Spinner/Spinner.stories.mdx
+++ b/src/components/Spinner/Spinner.stories.mdx
@@ -17,7 +17,7 @@ This is a [React](https://reactjs.org/) component for the Vanilla [Spin](https:/
 ### Default
 
 <Canvas>
-  <Story name="Default" args={{ role: "alert", ariaLive: "polite"}}>
+  <Story name="Default">
     {Template.bind({})}
   </Story>
 </Canvas>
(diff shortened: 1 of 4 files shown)


src/components/Spinner/Spinner.stories.mdx
```diff
@@ -17,15 +17,15 @@ This is a [React](https://reactjs.org/) component for the Vanilla [Spin](https:/
...
+  <Story name="Default" args={{ role: "alert", ariaLive: "polite"}}>
     {Template.bind({})}
   </Story>
 </Canvas>
 
 ### Spinner with text
 
 <Canvas>
-  <Story name="Text" args={{ text: "Loading..." }}>
+  <Story name="Text" args={{ text: "Loading...", role: "alert", ariaLive: "polite" }}>
```
bartaz, line 28: This also seems to repeat default values?

Instead of adding these props to existing examples maybe worth adding new example with some accessibility changes? For example how to change role and ariaLive to different values (not sure what different role would be though).

bartaz (changes requested):
Some comments on the examples

src/components/Spinner/Spinner.stories.mdx
bethcollins92: Ahh yes I forgot to remove these. I'll add an example with aria-live="assertive"

bethcollins92:
@bartaz thanks, made those changes

bartaz (commented):
LGTM, thanks

bartaz (approved):

Issue #707, opened 2022-02-11
Expose react-components hooks


petermakowski:
Should [canonical-web-and-design/react-components](https://github.com/canonical-web-and-design/react-components) document and expose* its hooks for other repositories to use? If not, should they live in a separate repository, like `canonical-web-and-design/react-hooks`?

I could find use for every single hook that we have in react-components in MAAS UI, for example. I'm pretty sure other canonical repositories could, too.

Can anyone see downsides of doing this?

I'd be happy to assist/lead this as it could help reduce duplication and bring more consistency across different canonical web React projects (and help stop solving the same problem multiple times). There are some hooks in MAAS UI that would be worth sharing with others, for example.

*Currently it's possible to import hooks via `"@canonical/react-components/dist/hooks";` but this is not officially supported AFAIK (there's no documentation and they're not added to the main [index.ts](https://github.com/canonical-web-and-design/react-components/blob/main/src/index.ts)).

PR #709, opened 2022-02-14
feat: expose react hooks

petermakowski:
## Done

- expose react hooks

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: #707  .

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -100,6 +100,14 @@ export type { TabsProps } from "./components/Tabs";
 export type { TextareaProps } from "./components/Textarea";
 export type { TooltipProps } from "./components/Tooltip";
 
+export {
+  useWindowFitment,
+  useListener,
+  usePrevious,
+  useThrottle,
+} from "./hooks";
+export type { WindowFitment } from "./hooks";
+
 export type {
   ClassName,
   Headings,
--- a/src/docs/getting-started.stories.mdx
+++ b/src/docs/getting-started.stories.mdx
@@ -1,6 +1,6 @@
 <Meta title="Getting started" />
 
-This package provides [React](https://reactjs.org/) components for [Vanilla Framework](http://vanillaframework.io).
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -100,6 +100,15 @@ export type { TabsProps } from "./components/Tabs";
 export type { TextareaProps } from "./components/Textarea";
 export type { TooltipProps } from "./components/Tooltip";
 
+export {
+  useWindowFitment,
+  useListener,
+  usePrevious,
+  useThrottle,
+  useId,
+} from "./hooks";
```
bartaz, line 109: Now that all hooks are exported (both from `hooks/index` and main index) should it affect how we import them internally?

Maybe doesn't have to be in this PR, but I see some inconsitencies, where sometimes we import via `from "hooks/useId"` and other times using `from "../../hooks".
I guess now we could just import all hooks `from "hooks"`?

src/index.ts
```diff
@@ -100,6 +100,15 @@ export type { TabsProps } from "./components/Tabs";
 export type { TextareaProps } from "./components/Textarea";
 export type { TooltipProps } from "./components/Tooltip";
 
+export {
+  useWindowFitment,
+  useListener,
+  usePrevious,
+  useThrottle,
+  useId,
```
bartaz, line 108: I wonder if there is a way to verify/test if we export everything?
I can see us forgetting to properly export things when we add new components or hooks. It happened couple times in the past…

bartaz (approved):
LGTM. Just some small comments/questions but looks good to go as it is.

src/index.ts
petermakowski: Absolutely, good point! Updated.

PR #711, opened 2022-02-16
test that all components & hooks are exported from the main index file

petermakowski:
## Done

- test that all components and hooks are exported from the main index file
  - thanks @bartaz for [suggesting this](https://github.com/canonical-web-and-design/react-components/pull/709#discussion_r807799705)
- add missing component exports

## QA

No need to QA manually, look at the test run from the first commit in this pull request - we had some missing exports and the test has failed:

image
https://github.com/canonical-web-and-design/react-components/runs/5219757859?check_suite_focus=true

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: # .


src/index.ts
```diff
@@ -28,6 +30,8 @@ export {
   NotificationSeverity,
 } from "./components/Notification";
 export { default as Pagination } from "./components/Pagination";
+export { default as PaginationButton } from "./components/PaginationButton";
+export { default as PaginationItem } from "./components/PaginationButton";
```
bartaz, line 34: I don't think we should be exporting these new components. They are supposed to be internally used by other components (`Accordion`, `Checkbox`, `Pagination`).

So I guess instead of adding them to `index` we should remove them from `components` list? And make sure when used internally they are always imported via relative path?

bartaz: Oh, sorry, I kindof thought that there is some `components/index`, but you just compare the contents of the `components` folder.

So in this case it would mean we have to move the components we don't want exported to subfolders or maintain some list of "internal" components that are not public API…

Any thoughts on that?

petermakowski: At the moment all these "internal" components seem to be used only in a single place each. It would make sense to move them inside of the component folder that they're being used in for now and revisit this when this becomes a problem?

Otherwise, like you said, using an "internal" folder (`components\internal`) that would be ignored by this test would work, too.

bartaz: @ilayda21 @ClementChaumel 

For a bit of context - we don't want to expose this additional internal components in the index file. So `AccordionSection` should be moved into `/components/Accordion`, pagination ones into `Pagination` folder, etc.
I guess the only one that is used in 2 places is `CheckableInput` (used in `RadioInput` and `CheckboxInput`. But I guess it's fine to move it into `Checkbox` and import it from there when used in `RadioInput`.

bartaz: Also, this was sitting in review for a while. So may need a rebase (some new components were likely added to the library since then).

petermakowski: Rebased, added missing `usePagination` export and updated test to skip hidden files and folders.

bartaz: @petermakowski Thanks, but I guess we still export the internal `AccordionSection`, `CheckableInput`, etc. And we wanted to hide them in the relevant component folders.

I started commenting on the issue as we wanted to work on this with guest devs, but thanks for the updates :)

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 
-import CheckableInput from "../CheckableInput";
-import type { Props as CheckableInputProps } from "../CheckableInput/CheckableInput";
+import CheckableInput from "./CheckableInput";
+import type { CheckableInputProps } from "./CheckableInput";
```
bartaz, line 4: It used to be `import type { Props as CheckableInputProps }`, as I think `CheckableInput` exports `Props`. Does it export `CheckableInputProps`?

bartaz: https://github.com/canonical/react-components/blob/86592fe06a99ea2a473aef3cdb67b7a72e611ca5/src/components/CheckboxInput/CheckableInput/CheckableInput.tsx#L8

bartaz (approved):
LGTM, thanks!

Issue #712, opened 2022-02-17
Switch component is missing 


minkyngkm:
React [Switch](https://vanillaframework.io/docs/patterns/switch) component is missing

PR #713, opened 2022-02-17
Add switch component

bethcollins92:
## Done

- Added new switch component with disabled example
- Wrote tests

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: #712 .

### DIFF
--- a/src/components/Switch/Switch.test.tsx
+++ b/src/components/Switch/Switch.test.tsx
@@ -1,0 +1,32 @@
+import { shallow } from "enzyme";
+import React from "react";
+
+import Switch from "./Switch";
+
+describe("<Switch />", () => {
+  // snapshot tests
+  it("renders and matches the snapshot", () => {
+    const component = shallow(<Switch label="Switch" />);
+
+    expect(component).toMatchSnapshot();
+  });
+
+  it("renders and matches the snapshot with disabled", () => {
+    const component = shallow(<Switch label="Disabled switch" disabled />);
+
+    expect(component).toMatchSnapshot();
+  });
+
+  //unit tests
+  it("renders a switch with a label", () => {
+    const component = shallow(<Switch label="Switch" />);
+
+    expect(component.find("label").first().hasClass("p-switch")).toBeTruthy();
+  });
+
+  it("renders a disabled switch", () => {
+    const component = shallow(<Switch label="Disabled switch" disabled />);
+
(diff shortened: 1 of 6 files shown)


src/components/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { shallow } from "enzyme";
```
bartaz, line 1: Now that we have `testing-library` it would be good to write new tests using it instead of enzyme.

There is an example in SummaryButton component:
https://github.com/canonical-web-and-design/react-components/blob/main/src/components/SummaryButton/SummaryButton.test.tsx

Could you update these tests so they don't use enzyme?

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * The label name for the switch
+     */
+    label: ReactNode;
+    /**
+     * Whether the switch is disabled or not
+     */
+    disabled?: boolean;
+  },
+  HTMLProps<HTMLInputElement>
```
bartaz, line 19: Looks like you allow passing additional props to the component (input element props), but you don't really read them or pass them to the input in the component?

bethcollins92:
Thanks for the review @bartaz - updated as suggested :)

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,44 @@
...
+>;
+
+export const Switch = ({
+  label,
+  disabled = false,
+  ...inputProps
+}: Props): JSX.Element => {
+  return (
+    <Field>
```
bartaz, line 28: I don't think we need to wrap it in Field component. It adds 2 additional wrapping divs for "form group" and "form validation".

This should just be a switch component without any additional form markup. So it's enough if `<label>` is root of this element.

bartaz (changes requested):
Sorry, just one more thing I noticed.

src/components/Switch/Switch.tsx
bethcollins92: Thanks, updated! @bartaz

bartaz:
@bethcollins92 You need to update snapshots in tests as well.

bethcollins92:
@bartaz Sorry I always forget that! Done

PR #723, opened 2022-03-09
feat: customisation of specific headers, rows and cells in ModularTable

petermakowski:
## Done

- Allow customisation of specific headers, rows and cells
  - made possible with: `getHeaderProps`, `getRowProps` and `getCellProps` which resolve additional custom props for specific elements
- **[breaking change]** improve type safety (accessors)
Previously TypeScript would not complain about an invalid accessor string value. Now it expects not just a string, but a correct value for column accessors.
  - before: `{ accessor: "disks", Header: "Disks",`
  - after `{ accessor: "status" as const, Header: "Status" }`

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: https://github.com/canonical-web-and-design/app-tribe/issues/761

### DIFF
--- a/src/components/ModularTable/ModularTable.test.tsx
+++ b/src/components/ModularTable/ModularTable.test.tsx
@@ -96,3 +96,21 @@ it("renders extra props", () => {
 
   expect(screen.getByTestId("testID")).toBeInTheDocument();
 });
+
+it("allows setting custom aria-label for specific cells", async () => {
+  render(
+    <ModularTable
+      columns={columns}
+      data={data}
+      footer="This is a footer"
+      getCellProps={({ value }) => ({
+        "aria-label":
+          value === "Waiting" ? "This is a cell with custom label" : undefined,
+      })}
+    />
+  );
+
+  expect(
+    screen.getByRole("cell", { name: "This is a cell with custom label" })
+  ).toBeInTheDocument();
+});
--- a/src/components/ModularTable/ModularTable.test.tsx
+++ b/src/components/ModularTable/ModularTable.test.tsx
@@ -4,10 +4,10 @@ import React from "react";
 import ModularTable from "./ModularTable";
 
 const columns = [
-  { accessor: "status", Header: "Status" },
-  { accessor: "cores", Header: "Cores", className: "u-align--right" },
-  { accessor: "ram", Header: "RAM", className: "u-align--right" },
-  { accessor: "disks", Header: "Disks", className: "u-align--right" },
+  { accessor: "status" as const, Header: "Status" },
+  { accessor: "cores" as const, Header: "Cores", className: "u-align--right" },
+  { accessor: "ram" as const, Header: "RAM", className: "u-align--right" },
+  { accessor: "disks" as const, Header: "Disks", className: "u-align--right" },
 ];
 const data = [
   {
@@ -97,12 +97,11 @@ it("renders extra props", () => {
   expect(screen.getByTestId("testID")).toBeInTheDocument();
 });
 
-it("allows setting custom aria-label for specific cells", async () => {
+it("allows setting custom aria-label for specific cells based on their value", async () => {
   render(
     <ModularTable
       columns={columns}
       data={data}
-      footer="This is a footer"
       getCellProps={({ value }) => ({
         "aria-label":
           value === "Waiting" ? "This is a cell with custom label" : undefined,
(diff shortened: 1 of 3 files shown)


src/components/ModularTable/ModularTable.tsx
```diff
@@ -25,21 +25,40 @@ export type Props<D extends Record<string, unknown>> = {
...
+
+function ModularTable<D extends Record<string, unknown>>({
   data,
   columns,
   emptyMsg,
   footer,
-}: Props<Record<string, unknown>>): JSX.Element {
-  const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } =
-    useTable({ columns, data });
+  getTableProps = defaultPropGetter,
```
huwshimi, line 43: I think setting default functions for optional function props can be a bit confusing, it makes it harder to understand when you're reading the code further down that the function call is optionally providing additional props.

src/components/ModularTable/ModularTable.tsx
```diff
@@ -25,21 +25,40 @@ export type Props<D extends Record<string, unknown>> = {
    * Optional extra row to display underneath the main table content.
    */
   footer?: ReactNode;
+  getTableProps?: TablePropGetter<D>;
+  getHeaderProps?: (column: Column<D>) => Record<string, unknown>;
```
huwshimi, line 29: It seems a shame that there isn't a more specific generic or return type for these that make them a bit more obvious how they're supported to be used (or at least at a quick look at react-table I couldn't see anything more specific).

On second thoughts, should/could we limit these to the element props they're used with e.g. `...=> HTMLProps<HTMLTableRowElement>` or do we need to allow for overriding specific react-table props as well?

huwshimi, line 29: It seems a shame that there isn't a more specific generic or return type for these that make them a bit more obvious how they're supported to be used (or at least at a quick look at react-table I couldn't see anything more specific).

On second thoughts, should/could we limit these to the element props they're used with e.g. `...=> HTMLProps<HTMLTableRowElement>` or do we need to allow for overriding specific react-table props as well?

huwshimi (commented):
A couple of suggestions/questions around the default getters and the prop return types.

This also needs needs some tests that these prop functions are doing what we expect.

src/components/ModularTable/ModularTable.tsx
```diff
@@ -27,21 +37,44 @@ export type Props<D extends Record<string, unknown>> = PropsWithSpread<
      * Optional extra row to display underneath the main table content.
      */
     footer?: ReactNode;
+    /**
+     * This function is used to resolve any props needed for a particular column's header cell.
+     */
+    getHeaderProps?: (
+      header: HeaderGroup<D>
+    ) => Partial<TableHeaderProps & HTMLProps<HTMLTableHeaderCellElement>>;
```
huwshimi, line 45: Great, thanks this much clearer.

src/components/ModularTable/ModularTable.test.tsx
```diff
@@ -4,10 +4,10 @@ import React from "react";
 import ModularTable from "./ModularTable";
 
 const columns = [
-  { accessor: "status", Header: "Status" },
-  { accessor: "cores", Header: "Cores", className: "u-align--right" },
-  { accessor: "ram", Header: "RAM", className: "u-align--right" },
-  { accessor: "disks", Header: "Disks", className: "u-align--right" },
+  { accessor: "status" as const, Header: "Status" },
```
huwshimi, line 7: I'm curious as to why this was necessary? Is this something we'll have to do when consuming the modular table in the apps (if so it'll be a breaking change)?

huwshimi (approved):
Looks good, thanks for fixing that up @petermakowski. Just one question around the `as const` usage.

src/components/ModularTable/ModularTable.tsx
petermakowski: Yeah, it was a good suggestion, thanks!

src/components/ModularTable/ModularTable.test.tsx
petermakowski: It is a breaking change I'm afraid. Previously TypeScript would not complain about an invalid accessor string value. Now - because I'm passing down the generic to react-table types they expect correct value for column accessors (not just a string, but e.g. "status").

petermakowski: Happy to open another PR and change this (or anything else) before release if needed.

petermakowski: Looping you in @bartaz

bartaz: Thanks. Saw this earlier but didn't have time to have a deeper look to understand the change, will try to have a look today.

I wonder how breaking of a change is it? The API or data structure doesn't seem to change. So is it just that TS compiler will complain?

Would it still work just complain about types? Would it matter at all if project itself uses JS and not TS?

petermakowski: I believe the TS build would fail (depending on how strict the config is in each project). JS projects would continue to work just fine.

petermakowski: The issue is detailed here with some examples: https://github.com/TanStack/react-table/discussions/2664#discussioncomment-335443

petermakowski: @bartaz I opened another PR that removes ModularTable strict type https://github.com/canonical-web-and-design/react-components/pull/733

This will give us a bit more time to think about this (stricter type is not required for the new ModularTable functionality to work).

PR #724, opened 2022-03-09
Add the Vanilla Label component.

huwshimi:
## Done

- Add the Vanilla Label component. There is already a form Label component so this had to be StatusLabel.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Visit the new StatusLabel storybook example (Edit: the demo service keeps timing out so you'll need to run this locally).
- Check that it appears like the Vanilla Label: https://vanillaframework.io/docs/patterns/labels

### DIFF
--- a/src/components/Button/Button.stories.mdx
+++ b/src/components/Button/Button.stories.mdx
@@ -1,6 +1,6 @@
 import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs/blocks";
 
-import Button from "./Button";
+import Button, { ButtonAppearance } from "./Button";
 
 <Meta
   title="Button"
@@ -11,6 +11,13 @@ import Button from "./Button";
         type: "text",
       },
       defaultValue: "Click me!",
+    },
+    appearance: {
+      control: {
+        options: ButtonAppearance,
+        type: "select",
+      },
+      defaultValue: ButtonAppearance.DEFAULT,
     },
   }}
 />
--- a/src/components/StatusLabel/StatusLabel.tsx
+++ b/src/components/StatusLabel/StatusLabel.tsx
@@ -5,11 +5,11 @@ import classNames from "classnames";
 import { ClassName, PropsWithSpread, ValueOf } from "types";
 
 export const StatusLabelAppearance = {
-  Caution: "caution",
-  Default: "",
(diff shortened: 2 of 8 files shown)


src/components/Button/Button.stories.mdx
```diff
@@ -12,6 +12,13 @@ import Button from "./Button";
       },
       defaultValue: "Click me!",
     },
+    appearance: {
```
huwshimi, line 15: Drive-by fix to show the button options in the canvas controls.

src/components/StatusLabel/StatusLabel.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import type { HTMLProps, ReactNode } from "react";
+
+import classNames from "classnames";
+import { ClassName, PropsWithSpread, ValueOf } from "types";
+
+export const StatusLabelAppearance = {
+  Caution: "caution",
```
bartaz, line 8: Why the camel case here? it seems in other enums we use `ALLCAPS` (like in buttons)?

huwshimi: Oh yeah sorry we've been moving to that syntax as it seems to be how it's written in the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/enums.html). I'll change this to allcaps for consistency.

huwshimi:
> The code itself looks fine, I'm just not sure about the name. `StatusLabel` is ok as a workaround, but I'm not sure we should settle on it. If we leave it like this it is going to be confusing, especially considering plans to possible merge vanilla and react-components into single project.
> 
> I guess the issue here shows that the label component in Vanilla is wrongly named, so probably we should try to rename it there and use the same name here. I'll try to bring this up on next Vanilla working group.
> 
> @huwshimi @petermakowski Is this change urgent for you, would it block you if we keep in on hold for another week or two?

This is not urgent at all, far from it, I built it in maas-ui but we didn't end up using it so I've just proposed it here so that the work doesn't get lost.

bartaz (approved):
LGTM

PR #727, opened 2022-03-14
Add missing props spreading for Chip, CodeSnippet, ModularTable and SearchBox

### LINKED ISSUE #722: Add event handlers for `CodeSnippet` interactions
## Context
On /advantage, we are trying to better understand how users interact with the page and get their token:
image

For this purpose, we would want to be able to log an event when the token is clicked or focused.

## Question
Could it be possible to pass down the `onClick` prop to the container?
Or if you think this problem is not generic enough, how would you suggest I go about making it work on our end?

## Side comment
While playing with it I saw that the inside of a `CodeSnippet` is not keyboard-focusable. I'm not sure how easy it would be for a keyboard only user to select what's inside it. Maybe something that could be looked at by the accessibility guild?

ClementChaumel:
## Done

- Update the Chip, CodeSnipper, ModularTable and Searchbox components to accept extra props

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Not exactly sure. I edited the stories locally to pass extra props to the components and checked they were indeed passed

## Fixes

Fixes #722

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -59,6 +59,7 @@ const Chip = ({
   selected,
   subString = "",
   value,
+  ...props
 }: Props): JSX.Element => {
   // When user tabs over chip and presses Enter or Space key, chip will trigger
   // click functionality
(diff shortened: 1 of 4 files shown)


src/components/Chip/Chip.tsx
```diff
@@ -92,9 +93,9 @@ const Chip = ({
 
   if (onDismiss) {
     return (
-      <span className={chipClassName} aria-pressed={selected}>
+      <span className={chipClassName} aria-pressed={selected} {...props}>
         {chipContent}
-        <button className="p-chip__dismiss" onClick={() => onDismiss()}>
+        <button className="p-chip__dismiss" onClick={onDismiss}>
```
ClementChaumel, line 98: This change prevent the component from recreating the function on each render

huwshimi (approved):
Looks good.

Issue #725, opened 2022-03-11
Add aria attributes to form help and errors


huwshimi:
When we generate the elements that provide help and error messages for form fields we should probably apply [aria-errormessage](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage) to the [error elements](https://github.com/canonical-web-and-design/react-components/blob/main/src/components/Field/Field.tsx#L71) and [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints#describing_with_aria) to the [help elements](https://github.com/canonical-web-and-design/react-components/blob/main/src/components/Field/Field.tsx#L68).

PR #728, opened 2022-03-14
Add error and help aria labels to form fields

huwshimi:
## Done

- Added aria labels to associate errors and help text to the field (some links to these roles in the issue: https://github.com/canonical-web-and-design/react-components/issues/725).
- Improved nanoid mock to avoid conflicts.

## QA
N/A no visual changes.

## Fixes

Fixes: #725.

### DIFF
--- a/src/__mocks__/nanoid.ts
+++ b/src/__mocks__/nanoid.ts
@@ -1,0 +1,10 @@
+let id = 0;
+
+beforeEach(() => {
+  id = 0;
+});
+
+export const nanoid = (): string => {
+  id++;
+  return `mock-nanoid-${id}`;
+};
--- a/src/components/AccordionSection/AccordionSection.test.tsx
+++ b/src/components/AccordionSection/AccordionSection.test.tsx
@@ -1,7 +1,6 @@
 import { shallow } from "enzyme";
 import React from "react";
 
-import { MOCK_UUID } from "../../setupTests";
 import AccordionSection from "./AccordionSection";
 
 describe("AccordionSection ", () => {
@@ -50,12 +49,12 @@ describe("AccordionSection ", () => {
     );
     wrapper.find(".p-accordion__tab").at(0).simulate("click");
 
-    expect(onTitleClick.mock.calls[0]).toEqual([true, MOCK_UUID]);
+    expect(onTitleClick.mock.calls[0]).toEqual([true, "mock-nanoid-1"]);
     wrapper.setProps({ expanded });
     // Clicking the title again should close the accordion section.
     wrapper.find(".p-accordion__tab").at(0).simulate("click");
 
-    expect(onTitleClick.mock.calls[1]).toEqual([false, MOCK_UUID]);
+    expect(onTitleClick.mock.calls[1]).toEqual([false, "mock-nanoid-1"]);
   });
 
   it("can use a key for expanded state", () => {
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -36,6 +36,10 @@ export type Props = {
    */
   help?: ReactNode;
   /**
+   * An id to give to the help element.
+   */
+  helpId?: string;
+  /**
    * Whether the component is wrapping a select element.
    */
   isSelect?: boolean;
@@ -63,15 +67,24 @@ export type Props = {
    * The content for success validation.
    */
   success?: ReactNode;
+  /**
+   * An id to give to the caution, error or success element.
+   */
+  validationId?: string;
 };
 
-const generateHelp = (help: Props["help"]) =>
-  help && <p className="p-form-help-text">{help}</p>;
+const generateHelp = (help: Props["help"], helpId: Props["helpId"]) =>
+  help && (
+    <p className="p-form-help-text" id={helpId}>
+      {help}
+    </p>
+  );
 
 const generateError = (
   error: Props["error"],
   caution: Props["caution"],
-  success: Props["success"]
+  success: Props["success"],
+  validationId: Props["validationId"]
 ) => {
   if (!error && !caution && !success) {
     return null;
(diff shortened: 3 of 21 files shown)


src/__mocks__/nanoid.ts
```diff
@@ -0,0 +1,10 @@
+let id = 0;
+
+beforeEach(() => {
+  id = 0;
+});
+
+export const nanoid = (): string => {
+  id++;
+  return `mock-nanoid-${id}`;
```
huwshimi, line 9: Added a unique id (in the same way as maas-ui) so that React Testing Library can correctly retrieve the right elements.

src/components/Field/Field.tsx
```diff
@@ -63,23 +67,32 @@ export type Props = {
    * The content for success validation.
    */
   success?: ReactNode;
+  /**
+   * An id to give to the caution, error or success element.
+   */
+  validationId?: string;
```
bartaz, line 73: Do we need separate props for different ids? Would it be enough to have one `id` prop for a Field, and have child element infer their ids from it like `help-${id}` `validation-${id}`? Or do we need to be able to pass individual ids separately?

src/components/AccordionSection/AccordionSection.test.tsx
```diff
@@ -50,12 +49,12 @@ describe("AccordionSection ", () => {
     );
     wrapper.find(".p-accordion__tab").at(0).simulate("click");
 
-    expect(onTitleClick.mock.calls[0]).toEqual([true, MOCK_UUID]);
+    expect(onTitleClick.mock.calls[0]).toEqual([true, "mock-nanoid-1"]);
```
bartaz, line 52: How do we know when writing tests what the id is gonna be? They reset to 0 every time? We need to know how many times nanoid is called in the components?

Also it seems that now when writing tests we need to know how this mocked version of nanoid works, that it returns ids starting with `mock-nanoid-`, etc. 

Not sure how problematic it may be in long term, but seems dev need to know much more about out specific setup of mocks.

src/components/Select/Select.test.tsx
```diff
@@ -43,10 +44,24 @@ describe("Select ", () => {
...
+    document.body.removeChild(container);
   });
 
   it("can take null options", () => {
     const wrapper = shallow(<Select options={null} />);
     expect(wrapper.exists()).toBe(true);
   });
 });
+
+describe("Select RTL", () => {
```
bartaz, line 56: Great that we write new tests in RTL already, but a bit of a pity to have 2 separate test suits in same file. But I get that it's not the scope of this to rewrite all the tests.

Maybe we need some RTL sprint to migrate all tests in react components.

src/components/Field/Field.tsx
huwshimi: We could pass a single id but then we'd have to know what the id syntax is in the parent component (Input, Select or our custom components in maas-etc.). I suppose if we document that well it wouldn't be so bad.

I can change this if you think that would be a better way to go?

src/components/AccordionSection/AccordionSection.test.tsx
huwshimi: For each test (each `it()`) the id starts at `...-1`. Without generating unique ids like this every element gets exactly the same id which causes clashes, which means that "<label for=.." aria labels etc. are associated with the wrong item (I guess the first item with an id).

If we don't mock the id at all then snapshots will receive new ids for every run, meaning that the snapshots won't pass.

This mock doesn't prevent a dev from using jest.mock in a single test to override the ids if they want to explicitly set an id (just like they could before), but also means that for the majority of cases it will just work.

Do you think that seems reasonable? Is there another approach you can think of?

src/components/Select/Select.test.tsx
huwshimi: Yeah it's not ideal. I can rewrite these with enzyme if you'd prefer.

bartaz: No, let's not write new tests in enzyme.

src/components/Field/Field.tsx
bartaz: I had a more detailed look how it's used in the `Input` component and it's fine I guess. You need to be able to pass it from parent.

Overall I find out react form components confusing a bit, the `Input` component that is not just an input but whole form field with label, validation, etc… But I guess the Vanilla messy form class names don't really help with that either :D

src/components/AccordionSection/AccordionSection.test.tsx
bartaz: @huwshimi I understand why we need it, it's just that to write a test with id now dev needs to know how the nanoid mock is setup and how the id name will be generated. And looking at the tests in 90% of cases we only need one id.

Maybe we can keep `MOCK_UUID` const with `mock-nanoid-1` value, so for all those basic cases when 1 id is needed it can still be used? But I guess this will make the cases when you need more ids more confusing…

I don't have any better idea at the moment.

bartaz (approved):
I left some comments for consideration, but looks good to go.

src/components/AccordionSection/AccordionSection.test.tsx
huwshimi: I've changed the Accordion tests to explicitly mock nanoid for the value we expect, do you think that makes it easier to understand?

bartaz: Yes, I think so. It's in the same test file, so it's easier to understand where it comes from. 

I know that the nanoid mock with incremental ids does kind of the same thing "automatically" so you don't have to copy it over and over between test files, but it also hides it from developer who writes or reads the tests. Having it mocked in the test file is more verbose (in a good way I think).

huwshimi: Great, thanks @bartaz let's try and use that pattern moving forward then.

Issue #739, opened 2022-03-22
Chips don't spread unless they're dismissible


huwshimi:
The non-dismissible version of the chip does not spread props.

https://github.com/canonical-web-and-design/react-components/blob/main/src/components/Chip/Chip.tsx#L108

Issue #738, opened 2022-03-22
Chips inside forms submit the form


huwshimi:
The chip button or chip dismiss button need `type="button"` to prevent submitting forms.

Issue #737, opened 2022-03-21
Chip spread overrides the chip classes


huwshimi:
If you provide additional classes to a chip they replace the `p-chip--...` classes here: https://github.com/canonical-web-and-design/react-components/blob/main/src/components/Chip/Chip.tsx#L92

PR #741, opened 2022-03-22
Improvements to the chip component.

huwshimi:
## Done

- Spread props on non-dismissible chips.
- Prevent form submissions when chips are inside forms.
- Prevent className spread wiping the "p-chip" classes.

## QA

N/A not really possible (I linked react-components to maas-ui and testing with my in-progress branch).

## Fixes

Fixes: #739.
Fixes: #738.
Fixes: #737.

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -26,10 +26,7 @@ export type Props = PropsWithSpread<
      * Function for handling chip div click event.
      */
     onClick?: (
-      event:
-        | MouseEvent<HTMLButtonElement>
-        | MouseEvent<HTMLSpanElement>
-        | { lead: string; value: string }
+      event: MouseEvent<HTMLButtonElement> | { lead: string; value: string }
     ) => void;
     /**
      * Function for handling dismissing a chip.
@@ -69,7 +66,7 @@ const Chip = ({
 }: Props): JSX.Element => {
   // When user tabs over chip and presses Enter or Space key, chip will trigger
   // click functionality
-  const onKeyDown = <E,>(e: KeyboardEvent<E>) => {
+  const onKeyDown = (e: KeyboardEvent<HTMLButtonElement>) => {
     // The " " value is what is returned for the spacebar
     if (e.key === " " || e.key === "Enter") {
       if (typeof onClick === "function") {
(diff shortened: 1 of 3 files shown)


huwshimi:
This should probably wait for https://github.com/canonical-web-and-design/vanilla-framework/issues/4366 before this can be reviewed.

src/components/Chip/Chip.tsx
```diff
@@ -89,27 +92,49 @@ const Chip = ({
...
+    onClick: onClick,
+  };
 
   if (onDismiss) {
     return (
-      <span className={chipClassName} aria-pressed={selected} {...props}>
+      <span
+        {...commonProps}
+        onKeyDown={(e) => onKeyDown<HTMLSpanElement>(e)}
+        role="button"
```
bartaz, line 115: Buttons can't be nested in each other. That's why chip with dismiss button is not interactive itself. It's bad from UX and accessibility point of view.

What kind of interactions do you need on dismissable chip other than dismissing it?

huwshimi: I've removed this part from the PR following the decision in https://github.com/canonical-web-and-design/vanilla-framework/issues/4366

huwshimi:
I've removed the click on span so the rest of this PR can be reviewed now (I'm not actually sure we'll need these changes for MAAS, but probably worth fixing these issues anyway).

src/components/Chip/Chip.tsx
```diff
@@ -89,27 +89,32 @@ const Chip = ({
...
+      [`p-chip--${appearance}`]: !!appearance,
+      "p-chip": !appearance,
+    },
+    props.className
+  );
 
   if (onDismiss) {
     return (
-      <span className={chipClassName} aria-pressed={selected} {...props}>
+      <span {...props} className={chipClassName}>
```
Caleb-Ellis, line 102: I'm not sure of the implications of being able to spread `HTMLProps<HTMLButtonElement>` onto a `<span>`... but if TS is fine with it I guess it's ok? The alternative would be to set it up so that if `onDismiss` exists then button props are allowed, otherwise span props are allowed etc but that might be more effort than its worth

Caleb-Ellis (approved):

Issue #732, opened 2022-03-16
Pagination has no semantics of currently active page


petermakowski:
Pagination has no semantics for indication of currently active page.
This makes for a bad accessibility and difficult to test using testing-library.

bartaz:
Vanilla pagination supports `aria-current="page"`, so it should be implemented into React component as well.

I guess that's where it can be fixed:

https://github.com/canonical-web-and-design/react-components/blob/01c63a24d877e699eb3a37ae18d00b02b4e884d7/src/components/PaginationItem/PaginationItem.tsx#L26-L33

PR #742, opened 2022-03-22
Add aria-current to Pagination Item

bethcollins92:
## Done

- Add `aria-current` to `Pagination Item` when it is active
- Add test for `aria-current`
- Update tests to use react testing library instead of jest 

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Click around and check nothing is broken
- Inspect the active button on the pagination docs, check `aria-current=page`

## Fixes

Fixes: #732

### DIFF
--- a/src/components/PaginationItem/PaginationItem.test.tsx
+++ b/src/components/PaginationItem/PaginationItem.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { render } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 
 import PaginationItem from "./PaginationItem";
 
@@ -15,29 +15,20 @@ describe("<PaginationItem />", () => {
 
   // unit tests
   it("displays the page number", () => {
-    const { container } = render(
-      <PaginationItem number={1} onClick={jest.fn()} isActive />
-    );
+    render(<PaginationItem number={1} onClick={jest.fn()} isActive />);
 
-    expect(container.querySelector("button")).toHaveTextContent("1");
+    expect(screen.getByRole("button", { name: "1" })).toBeInTheDocument();
   });
 
   it("sets isActive", () => {
-    const { container } = render(
-      <PaginationItem number={1} onClick={jest.fn()} isActive />
-    );
+    render(<PaginationItem number={1} onClick={jest.fn()} isActive />);
 
-    expect(container.querySelector("button")).toHaveClass("is-active");
+    expect(screen.getByRole("button")).toHaveClass("is-active");
   });
 
   it("sets aria-current when isActive is true", () => {
-    const { container } = render(
-      <PaginationItem number={1} onClick={jest.fn()} isActive />
-    );
+    render(<PaginationItem number={1} onClick={jest.fn()} isActive />);
 
-    expect(container.querySelector("button")).toHaveAttribute(
-      "aria-current",
-      "page"
-    );
+    expect(screen.getByRole("button", { current: "page" })).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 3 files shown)


src/components/PaginationItem/PaginationItem.test.tsx
```diff
@@ -1,32 +1,43 @@
...
   // unit tests
   it("displays the page number", () => {
-    const component = shallow(
-      <PaginationItem number={1} onClick={jest.fn()} />
+    const { container } = render(
+      <PaginationItem number={1} onClick={jest.fn()} isActive />
     );
 
-    expect(component.find("button").text()).toEqual("1");
+    expect(container.querySelector("button")).toHaveTextContent("1");
```
petermakowski, line 22: It's usually best to use testing library queries and screen, and not container methods directly.
```suggestion
    expect(screen.getByRole("button", { name: "1" })).toBeInTheDocument();
```

src/components/PaginationItem/PaginationItem.test.tsx
```diff
@@ -1,32 +1,43 @@
-import { shallow } from "enzyme";
 import React from "react";
+import { render } from "@testing-library/react";
```
petermakowski, line 2:
```suggestion
import { render, screen } from "@testing-library/react";
```

src/components/PaginationItem/PaginationItem.test.tsx
```diff
@@ -1,32 +1,43 @@
...
+  it("sets aria-current when isActive is true", () => {
+    const { container } = render(
       <PaginationItem number={1} onClick={jest.fn()} isActive />
     );
 
-    expect(component.find("button").hasClass("is-active"));
+    expect(container.querySelector("button")).toHaveAttribute(
+      "aria-current",
+      "page"
+    );
```
petermakowski, line 41:
```suggestion
    expect(getByRole("button", { current: "page" })).toBeInTheDocument();
```

src/components/PaginationItem/PaginationItem.tsx
```diff
@@ -28,6 +28,7 @@ const PaginationItem = ({
         "is-active": isActive,
       })}
       onClick={onClick}
+      aria-current={isActive ? "page" : "false"}
```
petermakowski, line 31: Have you double checked that `aria-current` is a valid attribute on buttons? I think I've only ever seen it being used on links.

petermakowski (commented):
Thanks for looking into this! A few minor comments below.

src/components/PaginationItem/PaginationItem.tsx
bethcollins92: I've had a look around, I can't anywhere which specifies it can/can't be on a button. But I did fin this [article](https://tink.uk/using-the-aria-current-attribute/) which uses it on a `<td>`, so figured a button is also valid?

petermakowski: OK awesome, thanks for checking this. Looks good 👍

petermakowski: (Nitpicking) What you did is perfectly valid (`aria-current="false"` is not exposed to the user), but what do you think about simplifying this and returning undefined?

This will skip the `aria-current` attribute for items other than for the current page, like in the [example on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#example).

```suggestion
      aria-current={isActive ? "page" : undefined}
```

petermakowski (approved):
LGTM. Thanks for refactoring to testing-library as well 👍

src/components/PaginationItem/PaginationItem.tsx
bethcollins92: Done! Thanks for the review @petermakowski :)

PR #761, opened 2022-04-14
feat: allow custom help text class

petermakowski:
## Done

- allow custom help text class

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/760

### DIFF
--- a/src/components/Field/Field.tsx
+++ b/src/components/Field/Field.tsx
@@ -36,6 +36,10 @@ export type Props = {
    */
   help?: ReactNode;
   /**
+   * Optional class(es) to pass to the help text element.
+   */
+  helpClassName?: ReactNode;
+  /**
    * An id to give to the help element.
    */
   helpId?: string;
@@ -80,11 +84,12 @@ export type Props = {
 const generateHelpText = ({
   help,
   helpId,
+  helpClassName,
   isTickElement,
-}: Pick<Props, "help" | "helpId" | "isTickElement">) =>
+}: Pick<Props, "help" | "helpId" | "helpClassName" | "isTickElement">) =>
   help ? (
     <p
-      className={classNames("p-form-help-text", {
+      className={classNames("p-form-help-text", helpClassName, {
         "is-tick-element": isTickElement,
       })}
       id={helpId}
@@ -138,6 +143,7 @@ const generateContent = ({
   labelFirst,
   labelNode,
   help,
+  helpClassName,
   error,
   caution,
   success,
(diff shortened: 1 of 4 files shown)


src/components/Field/Field.tsx
```diff
@@ -35,6 +35,10 @@ export type Props = {
    * Help text to show below the field.
    */
   help?: ReactNode;
+  /**
+   * Optional class(es) to pass to the help text element.
+   */
+  helpClassName?: ReactNode;
```
bartaz, line 41: I guess would be good to have some convention for adding those child class names, I think some other components do it as well (don't remember which one).

Not sure how detailed or strict we want to be on which elements have configurable class name and which don't. But at least some prop naming convention would be good.

`elementClassName` seems fine for start (and consistency with default `className`.

petermakowski: Agreed that this is needed (and I like your suggested naming convention).

I think we should have a common way of translating modifier classes that exist in vanilla to props in `react-components` as well, e.g. the `help` prop in `Input` should probably be named `helpText` instead to match `help-text` from Vanilla.

This would make working with react-components more intuitive for people who are already familiar with the vanilla framework (and vice-versa).

If we agree on a naming convention it might be worth updating all components props at some point even though there will be breaking changes.

petermakowski: I created an issue for this https://github.com/canonical-web-and-design/react-components/issues/762

petermakowski: Considering that the related element in vanilla has a base class `p-form-help-text`, would you consider a name of the element to be `formHelpText`? That would result in `formHelpTextClassName` instead of `helpClassName`. 

A bit lengthy, but I don't think that matters - having prop names that are long but consistent is much better than having props that are shorter but different in vanilla and react-components.

bartaz: Unfortunately form components in Vanilla have very inconsistent and lengthy names. There is no "p-form", but we do have "p-form-others"… and there are more issues like that, so forms are definitely not something we should base the naming convention on :D

petermakowski: Okay, should we go with my initial suggestion (`helpClassName`) then? It's consistent with the related `help` prop that already exists on this component.

Caleb-Ellis (approved):
LGTM! I think it's good that its consistent with the current `help` prop and we can define a better naming convention later on

PR #763, opened 2022-04-14
fix: tooltip on click should remain open until dismissed

petermakowski:
## Done

- fix: tooltip on click should remain open until dismissed
  - add cypress tests
  - add missing default port number to cypress commands

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check that the tooltip behaves as expected when activating on hover / keypress and can be dismissed
- You can use cypress test descriptions as a guidance

## Fixes

Fixes: https://github.com/canonical-web-and-design/react-components/issues/759


Caleb-Ellis (approved):
Nice!

Issue #134, opened 2020-06-15
Build Navigation component


Caleb-Ellis:
- Default
- Dark
- Full-width
- Sub navigation

Caleb-Ellis:
Putting this back into the backlog for now - got pretty tricky trying to align the Header component in maas-ui with a TS Navigation component in react-components. I have half-finished branches ready to pick up again when this comes back into an iteration.

cristinadresch:
@anthonydillon we were wondering why you marked this as high? do you know of any places that would use this?

anthonydillon:
Sorry, I must have missed this. Any web style project using react needs this, such as 360.

PR #767, opened 2022-04-22
Add Navigation component

huwshimi:
## Done

- Add a component to handle the Vanilla navigation pattern (originally built in MAAS here: https://github.com/canonical-web-and-design/maas-ui/pull/3833).

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the [navigation page](https://react-components-767.demos.haus/?path=/docs/navigation--default-story) in Storybook and check that the docs make sense and the examples work as expected.

## Fixes

Fixes: #134.


bartaz:
@huwshimi great, thanks.

Just wondering - we haven't seem to need this component in React in any project, MAAS is the only one I guess. And now it seems that there is work planned on moving to new layout for MAAS, so even MAAS won't need this navigation in React, right?

So, I'm just wondering if it's worth migrating it and maintaining here in React components?

huwshimi:
I think it'd be useful to have in react-components for any project adopting React. It's been quite painful to maintain the header in MAAS over the years because we were never quite aligned with Vanilla (or didn't keep up with the latest Vanilla markup).

I'd say any new project would find it quite a barrier to have to implement this themselves and sounds like there are other projects that could use this: https://github.com/canonical-web-and-design/react-components/issues/134#issuecomment-1106052069.

I do understand that this will be less useful to MAAS within the next year or so when we adopt the new layout, but that could be quite a while off still.

bartaz:
> I think it'd be useful to have in react-components for any project adopting React. It's been quite painful to maintain the header in MAAS over the years because we were never quite aligned with Vanilla (or didn't keep up with the latest Vanilla markup).

@huwshimi Sure, thanks for more context. I have it in my review queue.

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,262 @@
...
+                      <span className="p-navigation__search-label">Search</span>
+                    </button>
+                  </li>
+                ) : null
+              }
+            </ul>
+          ) : null}
+          {
+            // When the header has a search box and the user has opened the search
+            // form then this search box is displayed.
```
petermakowski, line 239: QA note on the UX of this.

Do you think you could set the search input field to receive focus when activated on search toggle button?

This would remove 1 extra click that's really unnecessary here. Ideally a user should be able to start typing the search query right after clicking the search button.

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Additional props to be applied to the nav element that wraps the
+     * left and right nav items.
+     */
+    navProps?: HTMLProps<HTMLElement> | null;
+    /**
+     * Additional props to be applied to the nav element that wraps the
+     * navigation items on the right hand side.
+     */
+    rightNavProps?: HTMLProps<HTMLUListElement> | null;
```
petermakowski, line 52: I think the API of this component can be a little confusing (in particular `leftNavProps`, `navProps`, `rightNavProps`). Ideally props should be self-explanatory.

It seems implicit in the name that `navProps` belongs to the same category as `leftNavProps` and `rightNavProps` (I'd intuitively think it's for the navigation group in the middle), but that's not the case. `navProps` are the props for the `nav` element whilst `leftNavProps` and `rightNavProps` are for the props of the inner `ul`. 

I wonder if we need a separate set of props at all. Although I'm not a big fan of this pattern (prefer compound components), It might make more sense to colocate item groups attributes with items themselves (and add an optional className/props attribute to each item in the array.

This is a pattern that `react-components` users are already familiar with (it's how we handle this problem in the MainTable https://github.com/canonical-web-and-design/react-components/blob/main/src/components/MainTable/MainTable.tsx#L48-L57)

huwshimi: Yeah it's not really brilliant is it. I guess I was hoping the comments in the props table would help to make it a bit more obvious. Could the names be clearer if they included "Element", like `navElementProps`?

I initially had a tuple for both the left and right navs, but realised that won't work for the case where you just need the right nav (like we need in MAAS in some cases).

Just to clarify, is your suggestion to have something like the following?

```jsx
<Navigation
    items={{
        items: [
            // The NavItem array 
        ],
        additionalLeftNavElementPropsHere: "",
    }}
    itemsRight={{
        items: [
            // The NavItem array 
        ],
        additionalRightNavElementPropsHere: "",
    }}
    navProps={{
        // We would still need this prop to get props to the nav element, the top level already spreads to <header>
        navElementPropsHere: "",
    }}
    additionalHeaderElementPropsHere=""
    ...
    />
```

src/components/Navigation/Navigation.tsx
huwshimi: Thanks! The docs also mention to close when esc is pressed so I've added that too.

src/components/Navigation/Navigation.tsx
petermakowski: > Just to clarify, is your suggestion to have something like the following?

Something along these lines. I thought we could even get rid of explicitly setting named groups, and simply pass them in an array.

With this structure we get rid of the props naming problem entirely.

```
<Navigation
    items={[{
        items: [
            // The NavItem array 
        ],
        additionalLeftNavElementPropsHere: "",
    },
    {
        items: [
            // The NavItem array 
        ],
        additionalRightNavElementPropsHere: "",
    }}
    ...
/>
```

huwshimi: > > Just to clarify, is your suggestion to have something like the following?
> 
> Something along these lines. I thought we could even get rid of explicitly setting named groups, and simply pass them in an array.
> 
> With this structure we get rid of the props naming problem entirely.
> 
> ```
> <Navigation
>     items={[{
>         items: [
>             // The NavItem array 
>         ],
>         additionalLeftNavElementPropsHere: "",
>     },
>     {
>         items: [
>             // The NavItem array 
>         ],
>         additionalRightNavElementPropsHere: "",
>     }}
>     ...
> />
> ```

I guess the awkward thing about using a tuple is that if you need to have items just on the right (like we sometimes do in MAAS) you have to include a null group `items={[null, {...}]}` which doesn't seem so obvious either...

petermakowski: True, but (correct me if I'm wrong) that's only because we relying on an empty ul element to align items to the right. I couldn't find an example for such use case on vanilla, but ideally we should be able to pass a className (to the nav group or container) to achieve the desired layout instead?

petermakowski: Even without having this use case covered by vanilla yet, simply adding `.u-flex--end` to `.p-navigation__nav` should do the trick? We would no longer need explicitly set item groups.

src/hooks/useOnEscape.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { renderHook } from "@testing-library/react-hooks";
+import { fireEvent } from "@testing-library/dom";
+
+import { useOnEscape } from "./useOnEscape";
+
+it("calls the callback when the escape key is pressed", () => {
+  const onEscape = jest.fn();
+  renderHook(() => useOnEscape(onEscape));
+  fireEvent.keyDown(document, { key: "Escape", code: "Escape" });
```
petermakowski, line 9: In a real world a whole sequence of keyboard events would be fired. It'd be better to not focus on implementation details though. A more user-centric way of testing this would be to use `userEvent.keyboard("{esc}")` instead.

src/hooks/useOnEscape.ts
```diff
@@ -0,0 +1,21 @@
+import { useCallback, useEffect } from "react";
+
+/**
+ * Handle the escape key pressed.
+ */
+export const useOnEscape = (onEscape: () => void) => {
```
petermakowski, line 6: Might be better to be more explicit here.
```suggestion
export const useOnEscapeKeyDown = (onEscape: () => void) => {
```
```suggestion
export const useOnEscapeKey = (onEscape: () => void) => {
```

src/components/Navigation/Navigation.tsx
huwshimi: Would that end up with a slightly awkward API in that there would be two mutually exclusive ways to align the groups? You'd have to either provide a tuple with both groups which would dictate which group was right or left (so you couldn't also have a class/prop to align them) or in the case that you only want to display one group you'd include one item in the array with a possible class/prop to align to the right.

In cases like MAAS where we display both states at different times we'd then have to do a bit of work to include the right things in the array with the right classes/props. That's certainly not hard to do, but I wonder if that becomes more messy to work with than isolating the props for left and right.

src/hooks/useOnEscape.test.tsx
huwshimi: For some reason when I tried that the events wouldn't fire. Do you happen to know if you have to get it to target the document somehow so the events fire? I wondered if it was an issue with the `react-hooks` test lib.

src/components/Navigation/Navigation.tsx
bartaz: @huwshimi @petermakowski it seems this PR leans into being a discussion about how we approach props API in our components. The more complex the component gets (like `MainTable`, this navigation, etc) the harder it is to pass all needed data as props and maintain flexibility that we are used to with Vanilla (being able to adjust HTML structure, class names, etc).

I'm starting to think that the way MUI approaches it (by using the composable child components) seems to be more flexible in this regard (for example: https://mui.com/material-ui/react-bottom-navigation/). 

I feel that we had a similar discussion in the past (when Modular Table was started maybe?), but I don't remember where it landed. 

I don't know what's the best approach, but the more complicated "items" props will be the harder it will be to use it or maintain in future. 
Or maybe some compromise would be to have a set of smaller composable components (`NavigationBar`, `NavigationItems`, `NavigationSearch`, etc) that you can use to custom build any "non standard" configuration for navigation, but we also provide `MainNavigation` component that you just pass props to that should cover most cases - kind of similar to what we have with `MainTable` being a controller component for  Table/TableRow/TableCell, etc…

petermakowski: Agreed @bartaz 

I'm in favour of breaking down into smaller composable components. There are a few different patterns that could work, I already mentioned [compound components](https://kentcdodds.com/blog/compound-components-with-react-hooks), simply grouping subcomponents by using a single prefix where sharing state is not required could work as well. 

The mui example you posted could work, too (although not everything I agree with, e.g. their approach of using boolean props to differentiate subcomponents probably isn't something we should follow - `<Grid container>
  <Grid item /> (...) </Grid>` ). https://mui.com/material-ui/react-grid/

huwshimi: We we initially talked about this we looked at the pattern of Material and other libraries and they often provided components without complex interactions, which leaves implementing the interactions/behaviour/functionality up to the consumers of the lib.

We however decided that react-components should provide both visual and functionality as, unlike a component library which is trying to be flexible, we want consistent behaviour across our sites/apps.

It's possible that we could revisit how we implement this, but we should make sure that we're able to implement behaviours in this lib.

src/hooks/useOnEscape.test.tsx
petermakowski: I tested this locally and had no issues: https://github.com/huwshimi/vanilla-react-components/compare/navigation-component...petermakowski:useOnEscape-userEvent

src/components/Navigation/Navigation.tsx
petermakowski: Okay, let's revisit this later. I created a separate issue to track this: https://github.com/canonical-web-and-design/react-components/issues/774

petermakowski:
I opened a PR with some changes to this component in maas-ui: https://github.com/canonical-web-and-design/maas-ui/pull/3908. If that looks ok to you, would be good to include them here before merge @huwshimi

petermakowski:
@huwshimi  I noticed that the mobile nav background doesn't match that of vanilla. 

image

petermakowski (approved):

src/hooks/useOnEscape.test.tsx
huwshimi: Oh great thanks. Not sure what I was doing before, but that seems to work now!

huwshimi:
> @huwshimi I noticed that the mobile nav background doesn't match that of vanilla.
> 
> image

It turns out this is an inconsistency in Vanilla. I've filed an issue: https://github.com/canonical-web-and-design/vanilla-framework/issues/4461.

huwshimi:
> I opened a PR with some changes to this component in maas-ui: [canonical-web-and-design/maas-ui#3908](https://github.com/canonical-web-and-design/maas-ui/pull/3908). If that looks ok to you, would be good to include them here before merge @huwshimi

Thanks, I've updated this PR with those changes.

src/components/Navigation/Navigation.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+});
+
+it("can open the search form", () => {
+  render(
+    <Navigation
+      logo={<img src="" alt="" />}
+      searchProps={{ onSearch: jest.fn() }}
+    />
+  );
+  userEvent.click(screen.getAllByRole("button", { name: "Search" })[0]);
```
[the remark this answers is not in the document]
huwshimi: I've added some integration tests for the search and mobile menu at different screen sizes.

cypress/integration/navigation.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  it("displays the mobile menu at small sizes", () => {
+    cy.viewport('iphone-6')
+    cy.findByRole("link", { name: "Products" }).should("not.exist");
+    cy.findByRole("button", { name: "Menu" }).should("be.visible");
+  });
+
+  it("can open the mobile menu", () => {
+    cy.viewport('iphone-6')
+    cy.findByRole("link", { name: "Products" }).should("not.exist");
+    cy.findByRole("button", { name: "Menu" }).click(0, 0);
```
petermakowski, line 20: I'm assuming this didn't work without passing exact pixels to click within the button and it's to workaround some visibility issue?

src/components/Navigation/Navigation.test.tsx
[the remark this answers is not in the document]
petermakowski: Awesome! Thanks.

petermakowski:
Not sure if this was like that last time I reviewed, but dropdown items on mobile look broken (do not extend to full width). 

#### This implementation
https://react-components-767.demos.haus/?path=/story/navigation--dropdown

image

#### Vanilla example 
https://vanillaframework.io/docs/patterns/navigation#dropdown
image

petermakowski:
> > @huwshimi I noticed that the mobile nav background doesn't match that of vanilla.
> > image
> 
> It turns out this is an inconsistency in Vanilla. I've filed an issue: [canonical-web-and-design/vanilla-framework#4461](https://github.com/canonical-web-and-design/vanilla-framework/issues/4461).

I couldn't find a single mention of the `.has-menu-open` class in vanilla documentation. 🤔  Might we be using something that's deprecated here?

huwshimi:
> > > @huwshimi I noticed that the mobile nav background doesn't match that of vanilla.
> > > image
> > 
> > 
> > It turns out this is an inconsistency in Vanilla. I've filed an issue: [canonical-web-and-design/vanilla-framework#4461](https://github.com/canonical-web-and-design/vanilla-framework/issues/4461).
> 
> I couldn't find a single mention of the `.has-menu-open` class in vanilla documentation. 🤔 Might we be using something that's deprecated here?

If you scroll down to the Expanding search box example you can see the grey background: https://vanillaframework.io/docs/patterns/navigation#expanding-search-box

cypress/integration/navigation.spec.js
huwshimi: Oh no, I just referenced how it was done in some of the other tests, but I'll remove that.

huwshimi:
> Not sure if this was like that last time I reviewed, but dropdown items on mobile look broken (do not extend to full width).
> 

That should be fixed with the next Vanilla release: https://github.com/canonical-web-and-design/vanilla-framework/issues/4444.

Not sure if that means we should hold off landing this until it's not broken.

bartaz:
> > Not sure if this was like that last time I reviewed, but dropdown items on mobile look broken (do not extend to full width).
> 
> That should be fixed with the next Vanilla release: [canonical-web-and-design/vanilla-framework#4444](https://github.com/canonical-web-and-design/vanilla-framework/issues/4444).
> 
> Not sure if that means we should hold off landing this until it's not broken.

It's practically ready to release, I'll try to publish it today.

PR #777, opened 2022-05-12
fix(List): uneven middot spacing

petermakowski:
## Done

- fix uneven middot spacing in a `List` component
- append a white space character at the end of each item if it's a `middot` list

### Note
CSS white space wrapping will prevent this space from being larger in case a component is already supplying items with a space appended to the end (2 spaces will be reduced to 1).

### Before
image

### After
image

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: https://github.com/canonical-web-and-design/app-tribe/issues/896
Fixes: https://github.com/canonical-web-and-design/react-components/issues/773

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -93,6 +93,7 @@ const generateItems = ({
         {...liProps}
       >
         {body}
+        {/* middot variant in Vanilla relies on whitespace between inline elements for correct spacing */}
         {middot ? " " : null}
       </li>
     );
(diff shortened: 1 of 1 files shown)


src/components/List/List.tsx
```diff
@@ -92,6 +93,7 @@ const generateItems = (
         {...liProps}
       >
         {body}
+        {middot ? " " : null}
```
bartaz, line 96: Would be good to have some comment around here to explain that the middot variant in Vanilla relies on whitespace between inline elements.

petermakowski: Sure, will add.

bartaz (approved):
LGTM

PR #796, opened 2022-06-29
Create badge component

ilayda21:
## Done

- A badge component is implemented.

## QA
- Open [the story book](https://react-components-796.demos.haus/?path=/story/badge--default-story).
- Verify the badge component is displayed and working properly. (for both badge types)
- Open "docs" page
- Verify the documentation done properly. 

## Fixes

Fixes: https://github.com/canonical-web-and-design/vanilla-squad/issues/1400

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@canonical/react-components",
-  "version": "0.37.2",
+  "version": "0.38.0",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "author": "Huw Wilkins <huw.wilkins@canonical.com>",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@canonical/react-components",
-  "version": "0.38.0",
+  "version": "0.37.2",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "author": "Huw Wilkins <huw.wilkins@canonical.com>",
--- a/src/components/Badge/Badge.stories.mdx
+++ b/src/components/Badge/Badge.stories.mdx
@@ -2,9 +2,9 @@ import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
 
 import Badge, { BadgeType } from "./Badge";
 
-<Meta 
-  title="Badge" 
-  component={Badge} 
+<Meta
+  title="Badge"
+  component={Badge}
   argTypes={{
     badgeType: {
       control: {
@@ -13,7 +13,7 @@ import Badge, { BadgeType } from "./Badge";
       },
       defaultValue: BadgeType.UNDEFINED_LARGE_NUMBER,
     },
-  }}  
+  }}
 />
 
 export const Template = (args) => <Badge {...args} />;
@@ -42,15 +42,25 @@ It can be used to display a numeric values.
 </Canvas>
 
 ### Colour Coding
+
 <Canvas>
   <Story name="Colour Coding">
-    <Badge value={9} isNegative/>
+    <Badge value={9} isNegative />
   </Story>
 </Canvas>
 
 ### Undefined Large Number
+
 <Canvas>
   <Story name="Undefined Large Number">
-    <Badge value={1000} badgeType={""}/>
+    <Badge value={1000} badgeType={""} />
+  </Story>
+</Canvas>
+
+### Rounded Large Number
+
+<Canvas>
+  <Story name="Rounded Large Number">
+    <Badge value={1234} badgeType={BadgeType.ROUNDED_LARGE_NUMBER} />
   </Story>
 </Canvas>
--- a/src/components/Badge/Badge.stories.mdx
+++ b/src/components/Badge/Badge.stories.mdx
@@ -1,0 +1,37 @@
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+
+import Badge from "./Badge";
+
+<Meta title="Badge" component={Badge} />
+
+export const Template = (args) => <Badge {...args} />;
+
+### Badge
+
+This is a [React](https://reactjs.org/) component for the Vanilla [Badge](https://vanillaframework.io/docs/patterns/badge).
+
+It can be used to display a numeric values.
+
+### Props
+
+<ArgsTable of={Badge} />
+
+### Default
+
+<Canvas>
+  <Story
+    name="Default"
+    args={{
+      value: 9,
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+### Colour Coding
+<Canvas>
+  <Story name="Colour Coding">
+    <Badge value={9} isNegative/>
+  </Story>
+</Canvas>
--- a/src/components/Badge/Badge.stories.mdx
+++ b/src/components/Badge/Badge.stories.mdx
@@ -54,7 +54,7 @@ It can be used to display a numeric values.
 When the amount of items is only relevant for small values, this variant should be used. When the value grows further than 999, the badge should show "999+".
 <Canvas>
   <Story name="Undefined Large Number">
-    <Badge value={1000} badgeType={""} />
+    <Badge value={1000} badgeType={BadgeType.UNDEFINED_LARGE_NUMBER} />
   </Story>
 </Canvas>
 
--- a/src/components/Badge/Badge.stories.mdx
+++ b/src/components/Badge/Badge.stories.mdx
@@ -51,6 +51,7 @@ It can be used to display a numeric values.
 
 ### Undefined Large Number
 
+When the amount of items is only relevant for small values, this variant should be used. When the value grows further than 999, the badge should show "999+".
 <Canvas>
   <Story name="Undefined Large Number">
     <Badge value={1000} badgeType={""} />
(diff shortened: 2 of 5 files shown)


src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const nextUnit = {
+    none: "k",
+    k: "M",
+    M: "B",
+    B: "T",
+  };
+
+  const round = (value, unit = "none") => {
+    console.log(value);
```
bartaz, line 58: Some `console.log` left in the code.

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const safeValue = value < 0 ? 0 : value;
+
+  const nextUnit = {
+    none: "k",
+    k: "M",
+    M: "B",
+    B: "T",
+  };
+
+  const round = (value, unit = "none") => {
```
bartaz, line 57: I guess the `round` and `clamp` functions could live outside of the component itself?
We probably don't need them as shared helpers yet, but they could be just static functions in the same file?

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+}: Props): JSX.Element => {
+  const badgeClassName = classNames(
+    {
+      [`p-badge--negative`]: !!isNegative,
+      "p-badge": !isNegative,
+    },
+    className
+  );
+
+  const safeValue = value < 0 ? 0 : value;
```
bartaz, line 48: I guess we could throw an error or show a warning in console when someone tries negative number?

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+   */
+  value: number;
+  /**
+   * The type of the badge component.
+   */
+  badgeType?: ValueOf<typeof BadgeType> | string;
+  /**
+   * The appearance of the badge.
+   */
+  isNegative?: boolean;
```
bartaz, line 25: This name gets really confusing if you think about it in the context of numbers (and negative numbers)…

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+    none: "k",
+    k: "M",
+    M: "B",
+    B: "T",
+  };
+
+  const round = (value, unit = "none") => {
+    console.log(value);
+    if (value < 1000) {
+      const truncatedValue = Number(value.toString().slice(0, 3));
```
bartaz, line 60: Why are we casting the value to string and back to a number again?
The value is `< 1000` in this conditional branch, so it should never have more than 3 characters?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@canonical/react-components",
-  "version": "0.37.2",
+  "version": "0.38.0",
```
bartaz, line 3: In React components we usually don't bump the version in each PR, because there always needs to be a "Prepare release" PR to update storyboook docs, so we do it at that point.

src/components/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+### Colour Coding
+
+<Canvas>
+  <Story name="Colour Coding">
+    <Badge value={9} isNegative />
+  </Story>
+</Canvas>
+
+### Undefined Large Number
+
```
bartaz, line 53: A sentence explaining what this option is would be helpful (that it clamps and adds +).

src/components/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+<Canvas>
+  <Story name="Undefined Large Number">
+    <Badge value={1000} badgeType={""} />
+  </Story>
+</Canvas>
+
+### Rounded Large Number
+
```
bartaz, line 61: A sentence explaining what this option is would be helpful (that it rounds and adds units).

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+ */
+export type Props = {
+  /**
+   * Numeric value to be displayed.
+   */
+  value: number;
+  /**
+   * The type of the badge component.
+   */
+  badgeType?: ValueOf<typeof BadgeType> | string;
```
bartaz, line 21: Do we need to allow any string here? I guess it should be limited to the allowed values only?

src/components/Badge/Badge.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+  <Story name="Colour Coding">
+    <Badge value={9} isNegative />
+  </Story>
+</Canvas>
+
+### Undefined Large Number
+
+<Canvas>
+  <Story name="Undefined Large Number">
+    <Badge value={1000} badgeType={""} />
```
bartaz, line 56: Does `badgeType` need to be empty here? Is this needed at all?

Or, for the sake of the example, show the whole value here `BadgeType.UNDEFINED_LARGE_NUMBER`

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+    k: "M",
+    M: "B",
+    B: "T",
+  };
+
+  const round = (value, unit = "none") => {
+    console.log(value);
+    if (value < 1000) {
+      const truncatedValue = Number(value.toString().slice(0, 3));
+      return `${truncatedValue}${unit === "none" ? "" : unit}`;
```
bartaz, line 61: It feels a bit strange to introduce an arbitrary "none" unit as a string, it seems it could be `undefined` or `null`? But it would affect the lookup for next unit.

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,85 @@
...
+    {
+      [`p-badge--negative`]: !!isNegative,
+      "p-badge": !isNegative,
+    },
+    className
+  );
+
+  const safeValue = value < 0 ? 0 : value;
+
+  const nextUnit = {
```
bartaz, line 50: Would it work as an array `["k", "M", "B", "T"]`? But that would affect the lookup for next unit, but a simple function should cover it?

bartaz (commented):
Looks great!

Thanks for detailed tests. Added couple comments inline for some small improvements.

package.json
ilayda21: Done :+1:

src/components/Badge/Badge.stories.mdx
ilayda21: :+1:

src/components/Badge/Badge.stories.mdx
ilayda21: :+1:

src/components/Badge/Badge.tsx
ilayda21: Right, we dont need it.

src/components/Badge/Badge.stories.mdx
ilayda21: It should be the undefined large number type. :+1:

src/components/Badge/Badge.tsx
ilayda21: Should we call it "isDark"?

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return value;
+  };
+
+  const formattedValue =
+    badgeType === BadgeType.ROUNDED_LARGE_NUMBER
+      ? round(safeValue)
+      : clamp(safeValue);
+
+  return <span className={badgeClassName}>{formattedValue}</span>;
```
bartaz, line 81: It should be also possible to pass any other valid HTML props to the react component, and they should be added to the Badge `<span>` element. Our other components should already do that, so you can see how it's done there.

ilayda21: Should we take value and html separately? Because we validate and change the value.

ilayda21: Got it know :+1:

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * The appearance of the badge.
+     */
+    isNegative?: boolean;
+    /**
+     * Optional class(es) to give to the badge.
+     */
+    className?: ClassName;
+  },
+  HTMLProps<HTMLElement>
```
bartaz, line 32: I'm not sure, but maybe there is a `HTMLProps<HTMLSpanElement>`?

ilayda21: Yes there is :+1:

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,92 @@
...
+const units = ["", "k", "M", "B", "T"];
+
+const round = (value: number, unit = 0) => {
+  if (value < 1000) {
+    const truncatedValue = Number(value.toString().slice(0, 4));
+    return `${truncatedValue}${units[unit]}`;
+  }
+
+  if (unit >= units.length - 1) {
+    return "+999T";
```
bartaz, line 45: This is more than 4 characters technically. And why + in the beginning?

I guess we are fine with leaving it as `999T` at that point?

PR #801, opened 2022-07-05
Allow both React v17 and v18 as peer dependency

bartaz:
## Done

- Allow using react-components with React v18 as peer dependency.

## QA

- Review if code change in package.json is correct


albertkol (approved):
What's the worst that could happen? :+1:

Issue #619, opened 2021-10-12
Broken homepage link in NPM


goulinkh:
The homepage (relative link) isn't working in the [package](https://www.npmjs.com/package/@canonical/react-components)'s page on NPM

PR #812, opened 2022-08-25
fix: Broken homepage link in NPM #619

petermakowski:
## Done

- fix Broken homepage link in NPM #619
- update all github urls to use canonical org

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: #619 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "bugs": {
     "url": "https://github.com/canonical/react-components/issues"
   },
-  "homepage": "https://github.com/canonical/react-components",
+  "homepage": "https://canonical.github.io/react-components",
   "devDependencies": {
     "@babel/cli": "7.18.9",
     "@babel/eslint-parser": "7.18.9",
--- a/package.json
+++ b/package.json
@@ -15,12 +15,12 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git@github.com:canonical-web-and-design/react-components.git"
+    "url": "git@github.com:canonical/react-components.git"
   },
   "bugs": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -15,12 +15,12 @@
...
     "type": "git",
-    "url": "git@github.com:canonical-web-and-design/react-components.git"
+    "url": "git@github.com:canonical/react-components.git"
   },
   "bugs": {
-    "url": "https://github.com/canonical-web-and-design/react-components/issues"
+    "url": "https://github.com/canonical/react-components/issues"
   },
-  "homepage": "/react-components/",
+  "homepage": "https://github.com/canonical/react-components",
```
bartaz, line 23: I think we had it as `/react-components` because Storybook uses it as a link under logo. So to make it work with GitHub-pages-hosted storybook we used to link to relative `/react-components`.

Maybe best to link it to `https://canonical.github.io/react-components` then? (one downside is that it will link to it in dev environment as well which may be confusing, but still better than broken link I guess).

bartaz (changes requested):

package.json
petermakowski: Good point, we're already linking to the repo from NPM, so `https://canonical.github.io/react-components/` would work better for the homepage. Will update.

BTW, it seems that the logo links to https://vanillaframework.io/ at the moment.

bartaz (approved):
LGTM

PR #819, opened 2022-09-08
feat: remove redundant id from useClickOutside

petermakowski:
## Done

- remove redundant id from useClickOutside
  - this is redundant as verifying if the clicked element is the current element can be achieved by using ref only

This is safe to remove as it's only being used in react-components itself and [TagSelector](https://github.com/canonical/maas-ui/blob/5647d8ed12c1f8a547b7633d035c96be05e3e908/src/app/base/components/TagSelector/TagSelector.tsx) in maas-ui

https://github.com/search?q=org%3Acanonical+useClickOutside&type=code

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the demo
- Verify that NavigationMenu opens and closes when clicking outside as before

## Fixes

Fixes: # .

### DIFF
--- a/src/hooks/useClickOutside.ts
+++ b/src/hooks/useClickOutside.ts
@@ -1,17 +1,13 @@
 import { MutableRefObject, useCallback, useEffect, useRef } from "react";
 
-import { useId } from "./useId";
-
 /**
  * Handle clicks outside an element.
- * @returns An id and ref to pass to the element to handle clicks
- * outside of.
+ * @returns A ref to pass to the element to handle clicks outside of.
  */
 export const useClickOutside = <E extends HTMLElement>(
   onClickOutside: () => void
-): [MutableRefObject<E>, string] => {
+): [MutableRefObject<E>] => {
   const wrapperRef = useRef<E | null>(null);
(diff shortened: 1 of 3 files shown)


src/hooks/useClickOutside.ts
```diff
@@ -24,18 +20,18 @@ export const useClickOutside = <E extends HTMLElement>(
...
+    [onClickOutside]
   );
 
   useEffect(() => {
     document.addEventListener("click", handleClickOutside, false);
     return () =>
       document.removeEventListener("click", handleClickOutside, false);
   }, [handleClickOutside]);
-  return [wrapperRef, id];
+  return [wrapperRef];
```
huwshimi, line 36: This might as well just return `wrapperRef` without the array.

huwshimi (approved):
Looks good!

src/hooks/useClickOutside.ts
petermakowski: True, I'll update.

PR #821, opened 2022-09-12
Add option for table captions

steverydz:
## Done

Added the option to add a hidden `caption` element to tables. This is to improve the experience for screen readers. See https://www.w3.org/WAI/tutorials/tables/caption-summary/

## QA

Check that the main table contains a `caption` element as the first child but that it is not visible on the screen

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: #854 .

### DIFF
--- a/src/components/MainTable/MainTable.test.tsx
+++ b/src/components/MainTable/MainTable.test.tsx
@@ -280,5 +280,20 @@ describe("MainTable", () => {
       expect(heads.first().prop("sort")).toEqual("none");
       expect(heads.at(1).prop("sort")).toEqual("ascending");
     });
+
+    it("shows a hidden caption if provided", () => {
+      const captionText = "This is a caption for screen readers";
+      const wrapper = mount(
+        <MainTable
+          defaultSort="status"
+          defaultSortDirection="descending"
+          headers={headers}
+          rows={rows}
+          sortable={true}
+          caption={captionText}
+        />
+      );
+      expect(wrapper.find("caption").text()).toEqual(captionText);
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


huwshimi:
Hey @steverydz, I know I'm butting in without any context, but I was curious as to what the benefit of using `caption` over `aria-label` is here? It seems to me that `<caption>` is designed for visible labels, and using `aria-label` seems to give the same result (from a quick test).

At the very least maybe I'd expect a `caption` prop to give a visible element so maybe this should have a different prop name?

As I said, I don't have the context so feel free to ignore this :).

steverydz:
Hey @huwshimi. Sorry, I should have added some context to the description of the PR.

This came from a discussion I had with James from HeX who has been working with us on accessibility. He said that all tables should have a caption because they are beneficial to users of screen readers, although they don't necessarily need to be visible on the screen.

You are right that `aria-label`, `aria-labelledby` and `aria-describedby` can all achieve this, and that `caption` is intended for visual purposes by default. The issue with using `aria-` attributes is that it could cause the caption/label be read out several times which would be frustrating for someone using a screen reader.

If used correctly it probably does make sense to use the `aria-` attributes most of the time, so long as they are used carefully and we are not *over-labelling* things :)

Perhaps changing the prop name to `hiddenCaption` would make more sense, to avoid any expection that the `caption` will be displayed. I've updated the PR.

@sowasred2012 Do you have any thoughts on this? I think I've got this right based on what James was saying.

huwshimi:
@steverydz thanks a lot, that makes sense. I think the prop name change helps.

anthonydillon:
Could you create a quick issue for this or add an estimate to track the effort

ClementChaumel (approved):
Outstanding 🤙

PR #822, opened 2022-09-13
Migrate Accordion to RTL

huwshimi:
## Done

- Migrate Accordion and AccordionSection to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/Accordion/Accordion.test.tsx
+++ b/src/components/Accordion/Accordion.test.tsx
@@ -1,12 +1,13 @@
-import { mount, shallow } from "enzyme";
 import React from "react";
 import * as nanoid from "nanoid";
 
 import Accordion from "./Accordion";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 describe("Accordion", () => {
   beforeEach(() => {
-    jest.spyOn(nanoid, "nanoid").mockReturnValue("mocked-nanoid");
+    jest.spyOn(nanoid, "nanoid").mockReturnValueOnce("mocked-nanoid");
   });
 
   afterEach(() => {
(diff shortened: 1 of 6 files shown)


src/components/Accordion/Accordion.test.tsx
```diff
@@ -47,14 +48,14 @@ describe("Accordion", () => {
         titleElement="h4"
       />
     );
-    expect(wrapper.find("AccordionSection").at(0).prop("titleElement")).toBe(
-      "h4"
-    );
+    // Query for the specific element as defined in the titleElement prop.
+    // eslint-disable-next-line testing-library/no-node-access
+    expect(document.querySelector("li h4")).toBeInTheDocument();
```
bartaz, line 53: "li h4" selector seems to be a bit too dependant on the HTML structure, and doesn't truly test what we want to test here (that the title element is h4).

Ideally we should get the title element using its role or text (we know the title) and check if it is an h4.

src/components/Accordion/AccordionSection/AccordionSection.test.tsx
```diff
@@ -27,16 +28,17 @@ describe("AccordionSection ", () => {
         titleElement="h4"
       />
     );
-
-    expect(wrapper.find("h4.p-accordion__heading")).toHaveLength(1);
-    expect(wrapper.find("h4").first().prop("aria-level")).toBeNull();
-    expect(wrapper).toMatchSnapshot();
+    // Query for the specific element as defined in the titleElement prop.
+    // eslint-disable-next-line testing-library/no-node-access
+    const title = document.querySelector("h4");
```
bartaz, line 33: Similar comment as above. We should query the title element (semantic way) and check if it is h4.

src/components/Accordion/Accordion.test.tsx
huwshimi: Done. Does that look better?

bartaz (approved):

PR #823, opened 2022-09-13
Migrate ActionButton to RTL.

huwshimi:
## Done

- Migrate ActionButton to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/ActionButton/ActionButton.tsx
+++ b/src/components/ActionButton/ActionButton.tsx
@@ -11,8 +11,8 @@ export const LOADER_MIN_DURATION = 400; // minimium duration (ms) loader display
 export const SUCCESS_DURATION = 2000; // duration (ms) success tick is displayed
 
 export enum Label {
-  Waiting = "Waiting for action to complete",
-  Success = "Action completed",
+  WAITING = "Waiting for action to complete",
+  SUCCESS = "Action completed",
 }
 
 export type Props = PropsWithSpread<
@@ -141,7 +141,7 @@ const ActionButton = ({
     >
       {showIcon ? (
         <Icon
-          aria-label={showLoader ? Label.Waiting : Label.Success}
+          aria-label={showLoader ? Label.WAITING : Label.SUCCESS}
           className={showLoader ? "u-animation--spin" : null}
           light={iconLight}
           name={icon}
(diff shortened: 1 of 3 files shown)


src/components/ActionButton/ActionButton.tsx
```diff
@@ -136,6 +141,7 @@ const ActionButton = ({
     >
       {showIcon ? (
         <Icon
+          aria-label={showLoader ? Label.Waiting : Label.Success}
```
bartaz, line 144: I guess it's good for the start, but ideally we will likely need to be able to pass a custom labels for loading and success here.

src/components/ActionButton/ActionButton.tsx
```diff
@@ -10,6 +10,11 @@ import type { ClassName, PropsWithSpread } from "types";
 export const LOADER_MIN_DURATION = 400; // minimium duration (ms) loader displays
 export const SUCCESS_DURATION = 2000; // duration (ms) success tick is displayed
 
+export enum Label {
+  Waiting = "Waiting for action to complete",
```
bartaz, line 14: Why TitleCase? Shouldn't it be UPPERCASE (like consts)?

src/components/ActionButton/ActionButton.tsx
bartaz: But I don't mean it for this PR, but something to consider in future.

src/components/ActionButton/ActionButton.tsx
huwshimi: I think we've been switching to that format because it's how a lot of the examples in TypeScript are written: https://www.typescriptlang.org/docs/handbook/enums.html, though the docs use both formats: https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time.

I've seen style guides recommend both formats. Maybe @petermakowski has more thoughts on this as I think he instigated the switch?

bartaz: Hmm… seems TypeScript community haven't decided yet: https://github.com/xojs/eslint-config-xo-typescript/issues/43

bartaz: Do we use enums elsewhere in our code? I guess we should just be consistent, the format doesn't matter too much.

petermakowski: We use enums in maas-ui for labels (although I'm personally not convinced they're a better choice than regular JS objects).

bartaz (approved):

PR #824, opened 2022-09-13
Migrate ArticlePagination to RTL.

huwshimi:
## Done

- Migrate ArticlePagination to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/ArticlePagination/ArticlePagination.test.tsx
+++ b/src/components/ArticlePagination/ArticlePagination.test.tsx
@@ -1,11 +1,11 @@
-import { shallow } from "enzyme";
+import { render, screen } from "@testing-library/react";
 import React from "react";
 
 import ArticlePagination from "./ArticlePagination";
 
 describe("ArticlePagination ", () => {
   it("renders", () => {
-    const wrapper = shallow(
+    render(
       <ArticlePagination
         nextLabel="Consectetur adipisicing elit"
         nextURL="#next"
@@ -13,13 +13,13 @@ describe("ArticlePagination ", () => {
         previousURL="#previous"
       />
     );
-    expect(wrapper).toMatchSnapshot();
+    expect(screen.getByRole("contentinfo")).toMatchSnapshot();
   });
 
   it("can add additional classes", () => {
-    const wrapper = shallow(<ArticlePagination className="extra-class" />);
-    const className = wrapper.prop("className");
-    expect(className.includes("p-article-pagination")).toBe(true);
-    expect(className.includes("extra-class")).toBe(true);
+    render(<ArticlePagination className="extra-class" />);
+    const pagination = screen.getByRole("contentinfo");
+    expect(pagination).toHaveClass("p-article-pagination");
+    expect(pagination).toHaveClass("extra-class");
   });
 });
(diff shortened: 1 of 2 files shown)


src/components/ArticlePagination/ArticlePagination.test.tsx
```diff
@@ -1,25 +1,25 @@
...
+    render(
       <ArticlePagination
         nextLabel="Consectetur adipisicing elit"
         nextURL="#next"
         previousLabel="Lorem ipsum dolor sit amet"
         previousURL="#previous"
       />
     );
-    expect(wrapper).toMatchSnapshot();
+    expect(screen.getByRole("contentinfo")).toMatchSnapshot();
```
bartaz, line 16: One thing that I don't like about RTL is selectors like this. 
How do I know that "ArticlePagination" is has `contentinfo` role? It's nowhere in the component code. My guess is it has this role because it's a "footer" element. But using `querySelector("footer")` would at least be explicit about it. In here we are pretending to be using semantic selector (because RTE forces us to it), but all we have is an indirect node query with confusing role that no one will know where it comes from.

Maybe for the cases like this (when we need root element of the component) we should have some testing ID convention in place? So all components will have some testing id on root element?

Thoughts @huwshimi @petermakowski ?

bartaz: Seems like there is a way to get to the container element at least: https://testing-library.com/docs/react-testing-library/api/#container-1

so when doing snapshots maybe we could use that instead of getting root element via implicit role?

```
const { container } = render(…);
expect(container).toMatchSnapshot();
```

Still, it won't solve the problem if we really want to access the root element of the component (and check class names on it or something).

petermakowski: > In here we are pretending to be using semantic selector (because RTE forces us to it), but all we have is an indirect node query with confusing role that no one will know where it comes from.

Testing library operates on the same roles that are exposed to the user. You can check the role of an element by using [logRoles](https://testing-library.com/docs/dom-testing-library/api-debugging/#logroles).

Alternatively you can navigate the UI using a screen reader or inspect using developer tools and you will get the same results.

image

> Maybe for the cases like this (when we need root element of the component) we should have some testing ID convention in place? So all components will have some testing id on root element?

No need to use a test id to get the root element of the component, you can use `container.firstChild`.
https://testing-library.com/docs/react-testing-library/api/#container-1

huwshimi: The downside of using the container rule is that at the moment we'd have to disable both the no container and no dom access linting rules for each test that we use it in.

I'll fix this PR up with whatever we decide in: https://github.com/canonical/react-components/issues/847.

bartaz: I'll approve and merge it as is for now, not to block RTL migration any longer.
But once we have all the tests in place would be good to get back, decides any rules/good practices and follow up according to them.

bartaz (approved):
I'll approve and merge it as is for now, not to block RTL migration any longer.
But once we have all the tests in place would be good to get back, decides any rules/good practices and follow up according to them.

PR #828, opened 2022-09-13
Migrate Chip to RTL.

huwshimi:
## Done

- Migrate Chip to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/Chip/Chip.test.tsx
+++ b/src/components/Chip/Chip.test.tsx
@@ -1,90 +1,100 @@
-import { shallow, mount } from "enzyme";
+import { render, screen, within } from "@testing-library/react";
 import React from "react";
 
-import Chip from "./Chip";
+import Chip, { Label } from "./Chip";
+import userEvent from "@testing-library/user-event";
 
 describe("Chip ", () => {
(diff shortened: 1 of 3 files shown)


src/components/Chip/Chip.test.tsx
```diff
@@ -1,140 +1,155 @@
...
 
-import Chip from "./Chip";
+import Chip, { Label } from "./Chip";
+import userEvent from "@testing-library/user-event";
 
 describe("Chip ", () => {
   it("renders default chip", () => {
-    const wrapper = shallow(<Chip value="positive" />);
-    expect(wrapper).toMatchSnapshot();
+    render(<Chip role="option" value="positive" />);
```
bartaz, line 9: Instead of passing arbitrary role just to use it in selector, maybe we can pass some test id? Or use `container` or `container.firstChild`.

huwshimi: I used a testid for this one so that I didn't have to disable the node access and container rules in all the tests.

ndv99 (approved):

PR #829, opened 2022-09-13
Migrate Code to RTL.

huwshimi:
## Done

- Migrate CodeSnippet to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/CodeSnippet/CodeSnippet.test.tsx
+++ b/src/components/CodeSnippet/CodeSnippet.test.tsx
@@ -1,24 +1,24 @@
-import { mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import React from "react";
 
 import CodeSnippet, { CodeSnippetBlockAppearance } from "./index";
 
 describe("CodeSnippet ", () => {
   it("renders a code block", () => {
-    const wrapper = mount(<CodeSnippet blocks={[{ code: "Test" }]} />);
-    expect(wrapper.find(".p-code-snippet__block").text()).toEqual("Test");
+    render(<CodeSnippet blocks={[{ code: "Test" }]} />);
+    expect(screen.getByText("Test")).toBeInTheDocument();
   });
 
   it("renders title for a code block", () => {
-    const wrapper = mount(
-      <CodeSnippet blocks={[{ title: "Title", code: "Test" }]} />
+    render(<CodeSnippet blocks={[{ title: "Title", code: "Test" }]} />);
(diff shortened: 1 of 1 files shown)


src/components/CodeSnippet/CodeSnippet.test.tsx
```diff
@@ -1,40 +1,41 @@
...
-    expect(wrapper.find(".p-code-snippet__block").text()).toEqual("Test");
+    render(<CodeSnippet blocks={[{ code: "Test" }]} />);
+    expect(screen.getByText("Test")).toBeInTheDocument();
   });
 
   it("renders title for a code block", () => {
-    const wrapper = mount(
-      <CodeSnippet blocks={[{ title: "Title", code: "Test" }]} />
+    render(<CodeSnippet blocks={[{ title: "Title", code: "Test" }]} />);
+    expect(screen.getByRole("heading", { name: "Title" })).toHaveClass(
```
bartaz, line 15: That's a good example of how I think we could write tests with RTL - use semantic RTL selectors, and check class names on them if needed.

src/components/CodeSnippet/CodeSnippet.test.tsx
```diff
@@ -44,26 +45,31 @@ describe("CodeSnippet ", () => {
...
-    const wrapper = mount(
+    render(
       <CodeSnippet
         blocks={[
           { appearance: CodeSnippetBlockAppearance.LINUX_PROMPT, code: "Test" },
         ]}
       />
     );
-    expect(wrapper.find(".p-code-snippet__block--icon").length).toBe(1);
+    expect(
```
bartaz, line 65: Could we test it other way around - get element by text (Test) and check if it has a classname with icon suffix?

bartaz: Same for other similar tests in same file.

huwshimi: The node that contains the text is in a child of the node with the class:
```html
<pre class="p-code-snippet__block--icon">
  <code>Test</code>
</pre>
```

I guess another way of doing this would be to use `.closest()` to find the parent element with the class (e.g. `screen.getByText("Test").closest(".p-code-snippet__block--icon")`) but the linter will also complain about that as `closest` is a standard DOM function, not a part of RTL.

bartaz: Oh, that's fair, let's keep it for now.

bartaz (approved):

PR #831, opened 2022-09-13
Migrate Col to RTL.

huwshimi:
## Done

- Migrate Col to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/Col/Col.test.tsx
+++ b/src/components/Col/Col.test.tsx
@@ -1,72 +1,76 @@
-import { shallow } from "enzyme";
+import { render, screen } from "@testing-library/react";
 import React from "react";
 
 import Col from "./Col";
 
 describe("Col ", () => {
   it("renders", () => {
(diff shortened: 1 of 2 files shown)


src/components/Col/Col.test.tsx
```diff
@@ -1,72 +1,76 @@
...
 describe("Col ", () => {
   it("renders", () => {
-    const wrapper = shallow(<Col size={3}>Test content</Col>);
-    expect(wrapper).toMatchSnapshot();
+    render(
+      <Col data-testid="col" size={3}>
+        Test content
+      </Col>
+    );
+    expect(screen.getByTestId("col")).toMatchSnapshot();
```
bartaz, line 13: This is a good way to do this I think. Set test id in the test itself, and use it to get an element.

bartaz (approved):
LGTM

PR #836, opened 2022-09-14
Migrate List to RTL.

huwshimi:
## Done

- Migrate List to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/List/List.test.tsx
+++ b/src/components/List/List.test.tsx
@@ -33,7 +33,7 @@ describe("List ", () => {
         ]}
       />
     );
-    const item = within(screen.getAllByRole("listitem")[0]).getByText("test");
+    const item = within(screen.getByRole("listitem")).getByText("test");
     expect(item).toBeInTheDocument();
     expect(item).toHaveClass("item-class");
   });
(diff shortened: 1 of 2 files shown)


src/components/List/List.test.tsx
```diff
@@ -12,72 +11,75 @@ describe("List ", () => {
...
+          {
+            content: "test",
+            className: "item-class",
+          },
+        ]}
+      />
     );
-    const item = wrapper.find(".p-list__item").first();
-    expect(item.prop("className").includes("item-class")).toBe(true);
+    const item = within(screen.getAllByRole("listitem")[0]).getByText("test");
```
bartaz, line 36: Why not directly `screen.getByText("test")`?

huwshimi: This test is to check that it constructed a `<li>` (or `listitem` role) from the object that was passed in. If we were to just check that the text was rendered this would pass if it didn't output a list.

Not sure why I used `getAllByRole` here, I'll update it to `screen.getByRole("listitem")`.

bartaz (approved):

PR #840, opened 2022-09-14
Migrate PasswordToggle to RTL.

huwshimi:
## Done

- Migrate PasswordToggle to React Testing Library to prepare for React 18.

## QA
n/a

### DIFF
--- a/src/components/PasswordToggle/PasswordToggle.tsx
+++ b/src/components/PasswordToggle/PasswordToggle.tsx
@@ -4,10 +4,15 @@ import type { InputHTMLAttributes, ReactNode } from "react";
 
 import Button from "../Button";
 import Field from "../Field";
-import Label from "../Label";
+import VanillaLabel from "../Label";
 
 import type { ClassName, PropsWithSpread } from "types";
 import { useId } from "hooks";
+
+export enum Label {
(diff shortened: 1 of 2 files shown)


src/components/PasswordToggle/PasswordToggle.tsx
```diff
@@ -4,11 +4,16 @@ import type { InputHTMLAttributes, ReactNode } from "react";
 
 import Button from "../Button";
 import Field from "../Field";
-import Label from "../Label";
+import VanillaLabel from "../Label";
 
 import type { ClassName, PropsWithSpread } from "types";
 import { useId } from "hooks";
 
+export enum Label {
```
bartaz, line 12: Oh, this makes it confusing. `Label` is a component and also an enum. And having it in title case `Label` for enum, makes it look like it's a component. 

But I don't have any better suggestion for now.

bartaz (approved):

src/components/PasswordToggle/PasswordToggle.tsx
huwshimi: Yeah good point, we could rename these to something like `LabelStrings` or `Text`.

PR #842, opened 2022-09-14
Migrate SearchAndFilter to RTL.

huwshimi:
## Done

- Migrate SearchAndFilter to React Testing Library to prepare for React 18.

## QA
n/a

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.test.tsx
+++ b/src/components/SearchAndFilter/SearchAndFilter.test.tsx
@@ -26,6 +26,12 @@ const sampleData = [
     chips: [{ value: "foo" }, { value: "bar" }, { value: "baz" }],
   },
 ];
+// eslint-disable-next-line testing-library/no-node-access
+const getPanel = () => document.querySelector(".p-search-and-filter__panel");
+
+const getSearchContainer = () =>
+  // eslint-disable-next-line testing-library/no-node-access
+  document.querySelector(".p-search-and-filter__search-container");
 
 describe("Search and filter", () => {
   it("renders", async () => {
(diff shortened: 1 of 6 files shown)


src/components/SearchAndFilter/SearchAndFilter.test.tsx
```diff
@@ -145,125 +167,201 @@ describe("Search and filter", () => {
...
     );
-    wrapper.find(".p-search-and-filter__input").simulate("click");
-    wrapper
-      .find(".p-search-and-filter__input")
-      .simulate("change", { target: { value: "Google" } });
-    const firstChip = wrapper.find(".p-chip").first();
-    expect(firstChip.find(".p-chip__value").html()).toEqual(
-      '<span class="p-chip__value"><strong>Google</strong></span>'
-    );
+    await waitFor(async () => {
```
bartaz, line 317: Many of these tests are getting quite verbose because of these events.

Is it neccessary to wrap all of them individually in `await waitFor`? Even if the content itself is `await userEvent.something…` ?

And also, we tent to query for the same element over and over again (like `screen.getByRole("searchbox", { name: Label.SearchAndFilter })`). Could we do it once per test and store in in a variable?

bartaz (approved):
I left a comment for consideration. But I'm approving it as is.

Let's have these tests in, get rid of Enzyme, update React and we can have a look into how to improve these tests later.

src/components/SearchAndFilter/SearchAndFilter.test.tsx
huwshimi: > Many of these tests are getting quite verbose because of these events.
> 
> Is it neccessary to wrap all of them individually in `await waitFor`? Even if the content itself is `await userEvent.something…` ?

As I understand it there are two different events these are waiting for. The `await userEvent.something…` is waiting for the click event but doesn't wait for other effects to complete. In this component once the panel is shown there is also a `useEffect` that responds to the state update. This state update then needs to be waited for hence the `await waitFor`.

I just had another try using things like `findBy` which is async, but I couldn't find another way to prevent the async errors.

> 
> And also, we tent to query for the same element over and over again (like `screen.getByRole("searchbox", { name: Label.SearchAndFilter })`). Could we do it once per test and store in in a variable?

I've moved all these instance to a function to get the element.

PR #843, opened 2022-09-14
Update SearchBox to RTL.

huwshimi:
## Done

- Migrate SearchBox to React Testing Library to prepare for React 18.

## QA
n/a

### DIFF
--- a/src/components/SearchBox/SearchBox.test.tsx
+++ b/src/components/SearchBox/SearchBox.test.tsx
@@ -1,76 +1,61 @@
-import { shallow, mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import React from "react";
 
-import SearchBox from "./SearchBox";
+import SearchBox, { Label } from "./SearchBox";
 
 describe("SearchBox ", () => {
   it("renders", () => {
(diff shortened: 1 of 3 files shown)


src/components/SearchBox/SearchBox.test.tsx
```diff
@@ -1,76 +1,61 @@
...
+import SearchBox, { Label } from "./SearchBox";
 
 describe("SearchBox ", () => {
   it("renders", () => {
-    const wrapper = shallow(<SearchBox />);
-    expect(wrapper).toMatchSnapshot();
+    render(<SearchBox />);
+    // Get the wrapping element for the snapshot.
+    // eslint-disable-next-line testing-library/no-node-access
+    expect(document.querySelector(".p-search-box")).toMatchSnapshot();
```
ndv99, line 12: Would getting by the "searchbox" role not work here?

ndv99 (approved):
Just one suggestion but it's optional, I'm guessing there's a reason it's been done like this

src/components/SearchBox/SearchBox.test.tsx
huwshimi: The searchbox role is only available on the text input but `.p-search-box` gets the wrapping element (which contains the label, input, buttons etc.)

PR #844, opened 2022-09-14
Migrate Select to RTL.

huwshimi:
## Done

- Migrate Select to React Testing Library to prepare for React 18.

## QA
n/a

### DIFF
--- a/src/components/Select/Select.test.tsx
+++ b/src/components/Select/Select.test.tsx
@@ -1,12 +1,11 @@
-import { render, screen } from "@testing-library/react";
-import { mount, shallow } from "enzyme";
+import { fireEvent, render, screen } from "@testing-library/react";
 import React from "react";
 
 import Select from "./Select";
 
 describe("Select", () => {
   it("renders", () => {
-    const wrapper = shallow(
+    render(
       <Select
         id="test-id"
         options={[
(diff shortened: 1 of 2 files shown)


src/components/Select/Select.test.tsx
```diff
@@ -15,45 +14,46 @@ describe("Select", () => {
           { value: "2", label: "Bionic Beaver" },
           { value: "3", label: "Xenial Xerus" },
         ]}
+        wrapperClassName="select"
       />
     );
-    expect(wrapper).toMatchSnapshot();
+    // eslint-disable-next-line testing-library/no-node-access
+    expect(document.querySelector("select")).toMatchSnapshot();
```
ndv99, line 21:
```suggestion
    expect(screen.getByRole("combobox")).toMatchSnapshot();
```

ndv99 (approved):
Looking good, just one suggested change but based on the rest of your code I'm guessing there's a reason you haven't done it like that, so I'm approving and leaving it to you

src/components/Select/Select.test.tsx
huwshimi: Hah, no idea why I did it like that! All the other tests in the file use `screen.getByRole("combobox")`. I've fixed it.

PR #849, opened 2022-09-14
Migrate Table to RTL.

huwshimi:
## Done

- Migrate Table to React Testing Library to prepare for React 18.

## QA

n/a

### DIFF
--- a/src/components/Table/Table.test.tsx
+++ b/src/components/Table/Table.test.tsx
@@ -3,30 +3,33 @@ import React from "react";
 
 import Table from "./Table";
 
+const TEST_ID = "table";
+
 describe("Table", () => {
   it("renders", () => {
-    const { container } = render(
-      <Table>
+    render(
+      <Table data-testid={TEST_ID}>
         <thead></thead>
       </Table>
     );
-    // eslint-disable-next-line testing-library/no-node-access, testing-library/no-container
-    expect(container.firstChild).toMatchSnapshot();
+    expect(screen.getByTestId(TEST_ID)).toMatchSnapshot();
   });
 
   it("can be expanding", () => {
-    render(<Table expanding={true} />);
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.test.tsx
```diff
@@ -1,36 +1,31 @@
...
 describe("Table", () => {
   it("renders", () => {
-    const wrapper = shallow(
+    render(
       <Table>
         <thead></thead>
       </Table>
     );
-    expect(wrapper).toMatchSnapshot();
+    expect(screen.getByRole("grid")).toMatchSnapshot();
```
bartaz, line 13: I guess we could use data id or container to get root element.

Looking at this, I'm not sure if `grid` is the right role for a table. It's meant for interactive grids, not static tables.

huwshimi: I've changed this to use the container.

src/components/Table/Table.test.tsx
```diff
@@ -1,36 +1,32 @@
...
   it("renders", () => {
-    const wrapper = shallow(
+    const { container } = render(
       <Table>
         <thead></thead>
       </Table>
     );
-    expect(wrapper).toMatchSnapshot();
+    // eslint-disable-next-line testing-library/no-node-access, testing-library/no-container
+    expect(container.firstChild).toMatchSnapshot();
```
ndv99, line 14: This doesn't feel like a good solution to me - based on the comments from Bartek, obviously getting it by role "grid" isn't ideal either so the best thing to do here might just be to use a data id.

ndv99: Either that or match the container directly to the snapshot I guess

huwshimi: I've updated these tests to use a test id.

ndv99 (approved):
Nice one mate!

PR #859, opened 2022-11-01
feat(ContextualMenu): allow node label

petermakowski:
## Done

- allow passing a node for ContextualMenu toggleLabel

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: # .

### DIFF
--- a/src/components/ContextualMenu/ContextualMenu.test.tsx
+++ b/src/components/ContextualMenu/ContextualMenu.test.tsx
@@ -1,4 +1,5 @@
 import { mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
 import React from "react";
 
 import ContextualMenu from "./ContextualMenu";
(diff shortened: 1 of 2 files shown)


ndv99 (approved):
Nice one, been needing this for a while. We should put this on notifs as well

petermakowski:
> Nice one, been needing this for a while. We should put this on notifs as well

@ndv99 Feel free to open a PR for it!

petermakowski:
> @petermakowski What's the context for this? It opens up adding **any** HTML elements into the button, and we don't want that. Label is supposed to be just a text.

It allows for more flexibility (e.g. adding an icon in front of text like in the example below).
image

bartaz:
> It allows for more flexibility (e.g. adding an icon in front of text like in the example below).
> image

OK. Not sure if contextual menu was designed and developed to cover such use case (from visual, functional, accessibility, etc. points of view). But I guess it's fine to keep it if that works for you.

I'm just a bit worried that this technically allows to put anything in there, regardless of being an icon/checkbox, regardless of contents of menu items.

Strictly speaking I would argue that this is a separate "ticked" variant of a contextual menu, that has these checkboxes built in, and doesn't allow users to put just anything in there.

But as I said, let's keep it for now, we can always restrict it more if it becomes misused.

PR #862, opened 2022-11-16
React 18

bartaz:
## Done

- Itemised list of what was changed by this PR.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: # .


bartaz:
@huwshimi @petermakowski 

Now that we got rid of enzyme I attempted upgrading to React 18. The components and storybook seems to run fine, but there are number of issues with build pipeline I don't know how to resolve (listed them in the description).

If you have any tips, or are able to jump on this PR and try to resolve it, that would be great.

Not sure if there is any rush to do this, while it seems some tools already moved to React 18, other still may not fully support it, so we need to make sure that everything we need works as expected.

huwshimi:
@bartaz the issue is how the eslint plugin resolution works when there are duplicates.

In our case `react` and `@typescript-eslint` are already included by the `react-app` extension ([here](https://github.com/facebook/create-react-app/blob/main/packages/eslint-config-react-app/base.js#L22) and [here](https://github.com/facebook/create-react-app/blob/main/packages/eslint-config-react-app/index.js#L45)). It looks like this is an issue that needs to be resolved upstream: https://github.com/facebook/create-react-app/issues/12400.

For now we can omit the duplicates from our config:

```diff
diff --git a/.eslintrc.js b/.eslintrc.js
index d801566..97bb378 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   parser: "@babel/eslint-parser",
-  plugins: ["react", "prettier", "cypress", "testing-library"],
+  plugins: ["prettier", "cypress", "testing-library"],
   extends: [
     "react-app", // Use the recommended rules from CRA.
     "plugin:cypress/recommended",
@@ -47,10 +47,8 @@ module.exports = {
     {
       files: ["**/*.ts?(x)"],
       parser: "@typescript-eslint/parser",
-      plugins: ["react", "prettier"],
       extends: [
         "react-app", // Uses the recommended rules from CRA.
-        "plugin:@typescript-eslint/recommended", // Uses the recommended rules from @typescript-eslint/eslint-plugin
         "plugin:prettier/recommended", // Ensure this is last in the list.
       ],
       parserOptions: {
```

bartaz:
Thanks @huwshimi, thanks. I thought that there may be something in config, but wasn't sure what is needed, and what not.
And the upstream issue is also a good find.

bartaz:
Thanks @huwshimi, with eslint out of the way I managed to untangle the rest of the issues and got it to a passing state.

Can you give this one a review?
Are our projects like MAAS JAAS, etc already on React 18? Will the new updated react-components still work with projects on React 17?

package.json
```diff
@@ -96,8 +95,6 @@
     "react-useportal": "1.0.17"
   },
   "resolutions": {
-    "@types/react": "17.0.48",
-    "@types/react-dom": "17.0.17",
     "postcss": "^8.3.11"
   },
   "peerDependencies": {
```
huwshimi, line 100: I think we probably should add `"@types/react"` here as well.

huwshimi (approved):
This looks good. It works when linked in the Juju dashboard. To get it working you have to link React versions as well so there's still a chance it might not work so maybe we should do a beta release once this lands.

package.json
bartaz: @huwshimi  As a peer dependency? Why would it be needed if we have it as dependency?

huwshimi: We have these peer dependencies:
```
"react": "^17.0.2 || ^18.0.0",
"react-dom": "^17.0.2 || ^18.0.0",
```
So adding `"@types/react": "^17.0.2 || ^18.0.0"` should mean that the correct types get resolved for the correct React version.

bartaz: Ah, yes, good call.

huwshimi:
I'd love to see this get landed as it's preventing those of us on React 18 from linking react-components to our projects to test react-components changes. Happy to do another review once the conflicts have been resolved.

huwshimi (approved):
@bartaz I hope you don't mind but I resolved the conflicts and added the types to the peer dependencies (as well as fixing the dep linting). 

I QAed this with the Juju Dashboard and it all seems to be working.

bartaz:
Thanks a lot @huwshimi, I don't mind at all. I know we just haven't been able to give react-components enough attention to unblock it ourselves so any help with that is more than welcome.

package.json
bartaz: @huwshimi If the react-components now use React 18, is it correct to have 17 in peer dependencies? Would it work with React 17, or should we have React 18 only in peer dependencies now?

huwshimi: I've just quickly tested it and react-components worked with a project using React 17. I think we want to have React 17 in peer deps for now otherwise those projects will get incorrect peer dependency warnings.

PR #868, opened 2022-12-06
Added optional takeFocusDelay to Input and make Modal focus on content rather than the close button

edlerd:


src/components/Modal/Modal.tsx
```diff
@@ -87,7 +87,9 @@ export const Modal = ({
     focusableModalElements.current = modalRef.current.querySelectorAll(
       focusableElementSelectors
     );
-    focusableModalElements.current[0]?.focus();
+    // when the close button is rendered, we want to focus on the 2nd element
+    const index = close && focusableModalElements.current.length > 1 ? 1 : 0;
```
bartaz, line 91: Could you rewrite this as more explicit if/else? Using `&&` together with `? :` makes this quite hard to parse.

src/components/Modal/Modal.test.tsx
```diff
@@ -83,13 +83,21 @@ describe("Modal ", () => {
...
+    expect(cancelButton).toHaveFocus();
 
     await user.tab();
 
-    expect(cancelButton).toHaveFocus();
+    expect(closeButton).toHaveFocus();
+  });
 
-    await user.tab();
+  it("traps focus within the modal close button", async () => {
```
bartaz, line 93: The name of the test seems a bit misleading, maybe "focuses on close button if there are no other focusable elements"?

bartaz:
@edlerd Hey David, thanks for the updates. What is the context of these changes? Are they related (Modal and Input) being in the same PR? In what cases do you need to delay the focus on input?

edlerd:
> @edlerd Hey David, thanks for the updates. What is the context of these changes? Are they related (Modal and Input) being in the same PR? In what cases do you need to delay the focus on input?

Context is usage in LXD-UI. We have some forms in Modals and want to autofocus on the 1st input in the form instead of the close button. It seems reasonable to focus on the content first, no?

Actually, not related, but they seemed both so small that I avoided the overhead of two PRs. Though, might have been the cleaner solution...

The delay is for the case the Input is in an Element that is flying in with animation (i.e. a Panel coming in from the left). I'll attach a quick video shortly.

src/components/Modal/Modal.test.tsx
edlerd: changed as well.

bartaz:
> The delay is for the case the Input is in a Wrapper that is flying in with animation (i.e. a Panel coming in from the right). I'll attach a quick video shortly. Chrome is moving the view so that the focused element is 100% visible, so instead of sliding in panel, we get a sliding in rest of the content. By delaying the focus until after the animation is finished (i.e. 300ms) this can be avoided.
> 

Thanks. Makes sense.

Normally I'd prefer the two (Modal / Input) to be separate PRs. The overhead is minimal, but later it's easier to track what change happened when and why.

But indeed they are quite small so isn't worth splitting it now.

bartaz (approved):
LGTM, thanks!

PR #869, opened 2022-12-07
Stop event propagation when Esc key is pressed on a Modal

lorumic:
When we have modals mounted above other layouts that handle the Esc key (e.g. the side panel), closing the modal by hitting the Esc key causes the underlying side panel to close too. This PR addresses that, by stopping the propagation of the event, and then it executes the `close` callback as before.

Before:

[before.webm](https://user-images.githubusercontent.com/56583786/206166193-a4d2b387-0bc4-473e-b5ac-e8b4b357b9b4.webm)

After:

[after.webm](https://user-images.githubusercontent.com/56583786/206166250-c1a1ff0e-2077-4a54-8850-65f4899d084a.webm)


src/components/Modal/Modal.tsx
```diff
@@ -74,8 +74,13 @@ export const Modal = ({
     }
   };
 
+  const handleEscKey = (e: KeyboardEvent<HTMLDivElement>) => {
+    e.stopPropagation();
```
bartaz, line 78: I think using `stopImmediatePropagation()` may be safer here.

Other key events may be handled on the same element (document), so I think using `stopImmediatePropagation()` would prevent any other events be triggered.

lorumic: Good point! I have changed that @ [94c5a54](https://github.com/canonical/react-components/commit/94c5a54914d1d1a429b04240dd03aefaba6a5eb2)

bartaz (approved):

PR #878, opened 2023-01-19
Feat: ModularTable now has the ability to use the useSortBy plugin.

moisesbenzan:
## Done

- Added `useSortBy` plugin capabilities to the `ModularTable` component.
- Added `sortable` prop to the `ModularTable` component.

### DIFF
--- a/src/components/ModularTable/ModularTable.stories.mdx
+++ b/src/components/ModularTable/ModularTable.stories.mdx
@@ -191,6 +191,98 @@ columns = {
   </Story>
 </Preview>
 
+### Sortable columns
+
+<Preview>
+  <Story name="Sortable">
+    <ModularTable
+      sortable
+      getCellProps={({ value }) => ({
+        className: value === "1 minute" ? "p-heading--5" : "",
+      })}
+      columns={React.useMemo(
+        () => [
+          {
+            Header: "ID",
+            accessor: "buildId",
+            sortType: 'basic',
+            Cell: ({ value }) => <a href="#">#{value}</a>,
+          },
+          {
+            Header: "Architecture",
+            accessor: "arch",
+            sortType: 'basic',
+          },
+          {
+            Header: "Build Duration",
+            accessor: "duration",
+            className: "u-hide--small",
+            sortType: 'basic',
+          },
+          {
+            Header: "Result",
+            accessor: "result",
+            sortType: 'basic',
+            Cell: ({ value }) => {
+              switch (value) {
+                case "released":
+                  return "Released";
+                case "failed":
+                  return "Failed";
+                default:
+                  return "Unknown";
+              }
+            },
+            getCellIcon: ({ value }) => {
+              switch (value) {
+                case "released":
+                  return ICONS.success;
+                case "failed":
+                  return ICONS.error;
+                default:
+                  return false;
+              }
+            },
+          },
+          {
+            Header: "Build Finished",
+            accessor: "finished",
+            className: "u-align-text--right",
+            sortType: 'basic',
+          },
+        ],
+        []
+      )}
+      data={React.useMemo(
+        () => [
+          {
+            buildId: "5432",
+            arch: "arm64",
+            duration: "5 minutes",
+            result: "released",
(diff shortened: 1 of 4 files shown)


src/components/ModularTable/ModularTable.tsx
```diff
@@ -97,9 +105,14 @@ function ModularTable({
                       : "",
                   },
                   { ...getHeaderProps?.(column) },
+                  // Only call this if we want it to be sortable too.
+                  sortable ? column.getSortByToggleProps() : undefined,
                 ])}
               >
                 {column.render("Header")}
+                <span>
+                  {column.isSorted ? (column.isSortedDesc ? " 🔽" : " 🔼") : ""}
```
bartaz, line 114: We shouldn't introduce new elements and emoji in the table UI.

Vanilla has it's own built-in styles for sortable tables: https://vanillaframework.io/docs/base/tables#sortable
Technically all it needs it adding `aria-sort` attribute on the `th` element (with one of the values of "none" "ascending" "descending"), so ideally `column.isSorted` should affect `aria-sort` on header, not add content to it.

src/components/ModularTable/ModularTable.stories.mdx
```diff
@@ -191,6 +191,98 @@ columns = {
   </Story>
 </Preview>
 
+### Sortable columns
+
+<Preview>
+  <Story name="Default">
```
bartaz, line 197: There can't be two "Default" stories, each needs unique name.

```suggestion
  <Story name="Sortable">
```

src/types/react-table-config.d.ts
```diff
@@ -1,12 +1,53 @@
...
 // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-table/Readme.md
 
-import "react-table";
+import {
+  UseSortByColumnOptions,
+  UseSortByColumnProps,
+  UseSortByHooks,
+  UseSortByInstanceProps,
+  UseSortByOptions,
+  UseSortByState,
```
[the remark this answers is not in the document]
moisesbenzan: Yea, no, I pushed a new commit with the only required one. This was more like leaving everything set up for a future implementation but we won't really do much with them now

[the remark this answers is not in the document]
bartaz: Oh cool, that's fine. 
Being future proof is OK as well, I was just a bit overwhelmed by all these new additions there, that didn't seem necessary.

But I also know that this library kind of makes you explicit about this stuff.

Thanks for the update though!

bartaz:
@moisesbenzan Nice work! Would it be possible to add at least some basic test for it? To make sure that the new "sortable" prop works as expected, not only by interactive demo?

moisesbenzan:
> @moisesbenzan Nice work! Would it be possible to add at least some basic test for it? To make sure that the new "sortable" prop works as expected, not only by interactive demo?

I'm not familiar with the testing framework used here :/ If you point me to the right direction I might be able to

moisesbenzan:
Added a basic test for the sortable option

src/components/ModularTable/ModularTable.test.tsx
```diff
@@ -42,6 +42,18 @@ it("renders all rows and columns", async () => {
   });
 });
 
+it("renders all rows and columns when the sortable option is set", async () => {
```
bartaz, line 45: Thanks, sorry, I haven't noticed your previous comment about giving you some tips.

Would it be possible to replicate a simple test like this one from `MainTable`?
https://github.com/canonical/react-components/blob/main/src/components/MainTable/MainTable.test.tsx#L177-L187

At least to verify "aria-sortable" is added to column headers. We can treat that as a minimal confirmation of the `sortable` feature doing something.

If you don't have time, I can do it myself sometime later this week to finalize and merge your PR.

moisesbenzan: Made a test following the basic scheme of that linked test case. Thanks!

bartaz (approved):
Thanks a lot @moisesbenzan, this looks great!

PR #879, opened 2023-01-24
fix: Prevent Slider from breaking if default value is 0

yurii-vasyliev:
## Done

Fixed Slider layout when the initial value is `0`

### DIFF
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -85,7 +85,7 @@ export const Slider = ({
   if (navigator?.userAgent?.includes("AppleWebKit")) {
     // Range inputs on Webkit browsers don't have a built-in "filled" portion,
     // so instead it is handled here as a background.
-    const val = inputProps.value || inputProps.defaultValue;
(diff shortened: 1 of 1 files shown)


src/components/Slider/Slider.tsx
```diff
@@ -85,7 +85,7 @@ export const Slider = ({
   if (navigator?.userAgent?.includes("AppleWebKit")) {
     // Range inputs on Webkit browsers don't have a built-in "filled" portion,
     // so instead it is handled here as a background.
-    const val = inputProps.value || inputProps.defaultValue;
+    const val = inputProps.value ?? inputProps.defaultValue;
```
bartaz, line 88: I forgot that this exists in modern JS 😁

yurii-vasyliev: Yeah, so that way we consider only `null` or `undefined` as "falsy", but not an empty string for example. Not sure, that would satisfy all conditions.

bartaz: I think it should be fine, the most common case is when the prop is not specified. If we need more strict checking we could also do that on TypeScript type level.

PR #880, opened 2023-01-24
Add centered prop to pagination component

steverydz:
## Done
- Added a `centered` prop to the `Pagination` component
- Update the markup to the [current Vanilla pattern](https://vanillaframework.io/docs/patterns/pagination)

## QA
- Go to https://react-components-880.demos.haus/?path=/story/pagination--default-story
- Toggle the `centered` boolean and check that the pagination is centered when set to `true`
- Go to https://react-components-880.demos.haus/?path=/docs/pagination--default-story#centered
- Check that there is an example of centered pagination

### DIFF
--- a/src/components/Pagination/Pagination.stories.mdx
+++ b/src/components/Pagination/Pagination.stories.mdx
@@ -57,3 +57,17 @@ The pagination component should be used to navigate between pages of content. De
     />
   </Story>
 </Canvas>
+
+### Centered
+
+<Canvas>
+  <Story name="Centered">
+    <Pagination
+      itemsPerPage={10}
+      totalItems={50}
+      paginate={() => {}}
+      currentPage={1}
+      centered
+    />
+  </Story>
+</Canvas>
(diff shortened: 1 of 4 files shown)


bartaz:
@steverydz Thanks for the PR, but I'm not sure if we need such feature that is basically adding a util class name to a component. We would need to add similar ones to all components practically for consistency.

All components support passing custom class names to allow adding any custom utils to them. So in this case

`<Pagination className="u-align--center" …>` should hopefully do the trick?

Or is there a particular reason why this doesn't work, or is making it specific prop for this component is more justified?

steverydz:
@bartaz The class gets applied to the `nav` element which wraps the pagination list. For the centering to work it needs to be applied to `ul.p-pagination` directly.

bartaz:
@steverydz Ah, yes, you are right, didn't notice that.

BTW, I just noticed that this react component is still using deprecated HTML structure. Compared to [examples in Vanilla](https://vanillaframework.io/docs/patterns/pagination), the `p-pagination` should be on the `nav` element, with `p-pagination__items` on `ol`.

This doesn't change the fact that the centered util needs to be on the list element, but would be good to fix the class names while we are already touching the component, if you don't mind :)

steverydz:
@bartaz I've updated the markup

Issue #875, opened 2023-01-12
Stepped List component p tag warning error  


albertkol:
The stepped List component surrounds the List item in a p tag which has the validator complain.

Basically almost every HTML element will have the validator complain about. Could we perhaps surround everything in a div instead?

![image](https://user-images.githubusercontent.com/6387619/212037915-df8b5072-7c0a-4e93-910e-5433f5bcfea7.png)

shdq:
Hello @albertkol. I hope you're doing well!

I just applied to Canonical and decided to look at open-source repos of the company. It seems you need a hand with this, so I looked at the issue and I think you're talking about nested lists.

This code would also produce the same semantic error. So it is not only about the stepped list.

```typescript
const innerItems = [
  {
    title: "My title",
    content:
      "My not very long sentence with small details.",
  },
];

const outerItems = [
  {
    content: <List items={innerItems} />,
  },
];

<List items={outerItems} />;

```

I suggest a quick fix for this changing `p` to `div`. I think `span` would fit better, but it won't work because it is `inline` element and there is no `css` rule in the corresponding class of vanilla framework (which is in another repo) that makes the element display as `block`. 

I also want to ask for @bartaz thoughts about this issue since he is the main contributor to this repo. And If it is ok, I can help with this issue.

Small feedback:

I reproduced this issue in the sandbox and there are no instructions for how to add styles along with the `react-components` package. I guess I need also to install `vanilla-framework`, and `sass`, build and import styles, but it wasn't needed to reproduce this issue. So it would be great to have clear instructions in this repo about how to have styles out of the box.

I also noticed that [vanilla framework API on lists](https://vanillaframework.io/docs/patterns/lists) is not fully supported. For example `Nested Count`, but it is out of the scope of this issue.

bartaz:
@shdq Hi Sergei, thanks for reaching out and looking into that. I guess Albert went a bit over the top with the labels, but it's definitely bug that would be worth fixing.

I believe the issue mentioned is about the stepped list variant (https://canonical.github.io/react-components/?path=/docs/list--default-story#vertical-stepped). And the issue with wrongly nested tags can happen if you pass multiple `p` inside the `content`.

We are likely to look into this ourselves when working on bugfixes in next couple weeks, but if you want to look into that and provide a PR @shdq feel free to do so, and we can assist in getting it reviewed and merged.



As for your other questions:

[...]

Issue #789, opened 2022-06-15
The `content` wrapper is a p and it throws `validateDOMNesting` errors


ClementChaumel:
Some elements should not be childs of `<p>` tags and it throws some errors
![image](https://user-images.githubusercontent.com/11927929/173817451-a5de835e-5bd4-40a2-b08f-e55f2427d2a3.png)

This is because the content container is a `<p>`
![image](https://user-images.githubusercontent.com/11927929/173817246-ee5135d0-3348-4b8b-afbe-df71e6d20d15.png)

We should update it to a `<div>` or something that would prevent such errors.

PR #886, opened 2023-02-14
fix(List): fix stepped list dom validation error for nested paragraphs

shdq:
## Description of the changes

Hello there! I hope you're doing great!

This PR is intended to fix DOM validation error for the stepped `List` with nested paragraphs.

**Related issues**: #875 #789 

Wrapper HTML element for the content of an item from `<p>` to `<div>`.

**Before changes**

```html
<ol class="p-stepped-list">
    <li class="p-stepped-list__item">
        <h3 class="p-stepped-list__title">Title and content as string</h3>
        <p class="p-stepped-list__content">Content as string</p>
    </li>
</ol>
```

**After changes**

```html
<ol class="p-stepped-list">
    <li class="p-stepped-list__item">
        <h3 class="p-stepped-list__title">Title and content as string</h3>
        <div class="p-stepped-list__content">Content as string</div>
    </li>
</ol>
```

After the change, when you provide one or several paragraphs as `content` it won't be an issue.

```typescript

const items = [
    {
        title: "With two paragraphs of text",
        content: (
            <>
              <p>Paragraph 1</p>
              <p>Paragraph 2</p>
            </>
        ),
    }
];

```

## QA

Unit tests were added for this case, and they pass with no regression on other tests. Linted the code and tests (I would suggest adding a pre-push hook for linting, because there are plenty of warnings in other components).

I checked the storybook: stepped `List` rendered correctly. Packed as a local package, installed in an empty Vite project: it also worked well.

Should I also add a story with this use case to the storybook?

If you have any questions, I would be happy to answer!

**Outside of the scope of this PR**

- I [issued](https://github.com/canonical/vanilla-framework/issues/4656) a bug in `vanilla framework`, that is connected to the stepped `List`.
- There is a problem with the types for `List` items, for the stepped list. Items prop always has the type `ListItem[]`, but supposed to be `SteppedListItem[]`. If you want I can create an issue and we can continue the conversation there.

### DIFF
--- a/src/components/List/List.test.tsx
+++ b/src/components/List/List.test.tsx
@@ -2,9 +2,10 @@ import { render, screen, within } from "@testing-library/react";
 import React from "react";
 
 import List from "./List";
+import type { ListItem, SteppedListItem } from "./List";
 
 describe("List ", () => {
-  let items: string[];
+  let items: ListItem[];
 
   beforeEach(() => {
     items = ["test", "items", "here"];
@@ -135,7 +136,7 @@ describe("List ", () => {
 
     it("should render item's content inside <div>", () => {
       // Prepare
-      const items = [{ title: "Title", content: "Content" }];
+      const items: SteppedListItem[] = [{ title: "Title", content: "Content" }];
 
       // Act
       render(<List stepped items={items} />);
@@ -153,7 +154,7 @@ describe("List ", () => {
 
     it("can have several paragraphs in item's content", () => {
       // Prepare
-      const items = [
+      const items: SteppedListItem[] = [
         {
           title: "With two paragraphs of text",
           content: (
--- a/src/components/List/List.test.tsx
+++ b/src/components/List/List.test.tsx
@@ -78,61 +78,103 @@ describe("List ", () => {
     expect(screen.getAllByRole("listitem")[0]).toHaveClass("is-ticked");
   });
 
-  it("can be stepped", () => {
-    render(
-      <List
-        stepped={true}
-        items={[
-          {
-            title: "Test title",
-            content: "test",
-            className: "item-class",
-            titleElement: "h1",
-          },
-        ]}
-      />
-    );
-    expect(screen.getByRole("list")).toHaveClass("p-stepped-list");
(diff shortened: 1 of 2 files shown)


shdq:
Hey there! 

The workflow failed because of a bug in the types for the list component.

Basically, if you have `stepped={true}` property it should be `SteppedListItem[]` type for items, and `ListItem[]` overwise.

Another problem with types is that `title` prop should be omitted from HTML attributes because there is a global `title` attribute and they are conflicting.

I created all the possible variants for items for the stepped list:

```typescript
const items: SteppedListItem[] = [
    { title: "Title and content as string", content: "Content as string" },
    { title: "With content as paragraph", content: <p>Paragraph</p> },
    {
      title: (
        <>
          With&nbsp;<strong>bold text</strong>&nbsp;and text
        </>
      ),
      content: (
        <>
          <p>Paragraph 1</p>
          <p>Paragraph 2</p>
        </>
      ),
    },
    {
      // title and content as numbers
      title: 1234567890,
      content: 1234567890,
    },
    {
      title: "With custom class name",
      content: '"my-class" added to <li>',
      className: "my-class",
    },
    {
      title: "With title as <h2>",
      content: "Item with changes title element",
      titleElement: "h2",
    },
  ];
```

But also there is an option "Stepped without headings" in [docs](https://vanillaframework.io/docs/patterns/lists#vertical-stepped) that is not supported by types: `title` is a required property in `SteppedListItem`.

I mentioned this problem above, and I wrote a fix for this.

@ClementChaumel  I want to hear your thoughts on this.

ClementChaumel:
Hey @shdq !

What you've done so far looks good! If you can fix the type error and maybe rename the item title to something less ambiguous I think we're looking at a closed issue.

Please do push your changes and when the tests pass I'll give it another review.

Thank you very much for the time and thoughts you put into this!

ClementChaumel (approved):
Great stuff!

ClementChaumel:
Thank you for your work on this issue and the detail you went into explaining your reasoning! That's a resounding :+1: !! :)

PR #887, opened 2023-02-16
Add onClear prop to SearchBox

goulinkh:
## Done

- Add an optional onClear prop and call it when clear button is clicked

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Type in the SearchBox component
- Provide a function for the `onClear` prop
- Make sure that the function is called when clicking on the reset button


huwshimi (approved):
Works great! If you felt like it you could also update line 105 to `onChange={(evt) => onChange?.(evt.target.value)}` as that is giving errors when the onChange prop is not defined.

bartaz (approved):
LGTM

PR #889, opened 2023-02-17
Fix storybook docs add-on

bartaz:
## Done

Seems like I found this fix [long time ago](https://github.com/storybookjs/storybook/issues/18876#issuecomment-1211644672) but somehow didn't push it back then?
Time to fix it now, that we updated storybook dependencies and it got broken again…

## QA

Go to storybook (https://react-components-889.demos.haus/?path=/story/accordion--default-story) and make sure Docs tab is available:

image


ClementChaumel (approved):
Great 👏️

PR #892, opened 2023-02-20
Add script to automate the preparation of a new release

ClementChaumel:
Added a script to automate prep of new releases because I'm lazy and forgetful

### DIFF
--- a/prepare
+++ b/prepare
@@ -40,6 +40,9 @@ git checkout -b prepare-$new_version-release origin/main
 # Update version in package.json
 yarn version --new-version $new_version
 
+# Update dependencies
+yarn
+
 # Build docs
 yarn build-docs
 
--- a/prepare
+++ b/prepare
@@ -32,7 +32,7 @@ new_version="$major.$minor.$patch"
 echo "Preparing $new_version"
 
 # Switch to new branch
-git checkout -b prepare-$new_version-release main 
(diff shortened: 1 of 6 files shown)


prepare
```diff
@@ -0,0 +1,50 @@
...
+  minor=0
+  patch=0
+fi
+
+# create new version
+new_version="$major.$minor.$patch"
+echo "Preparing $new_version"
+
+# Switch to new branch
+git checkout -b prepare-$new_version-release main
```
bartaz, line 35: Just to be sure a GH main branch is used, not local one (in case it has any local changes).
Also it would be good to do a `git fetch` first?

```suggestion
git checkout -b prepare-$new_version-release origin/main 
```

prepare
```diff
@@ -0,0 +1,53 @@
...
+git fetch
+
+# Switch to new branch
+git checkout -b prepare-$new_version-release origin/main 
+
+# Update version in package.json
+yarn version --new-version $new_version
+
+# Build docs
+yarn build-docs
```
bartaz, line 44: should we do `yarn install` first as well? just to be sure everything is installed as in main branch

bartaz:
Not sure why the test started failing? It doesn't seem related…

bartaz (approved):
LGTM

PR #893, opened 2023-03-01
Add zIndex prop for Tooltip component [WD-2282]

lorumic:
## Done

Added `zIndex` prop for the `Tooltip` component. Setting this property can be useful, for example, to allow a `Tooltip` to appear over an `l-aside` panel.

## Fixes

Fixes https://github.com/canonical/vanilla-framework/issues/4607.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -70,6 +70,10 @@ export type Props = {
    * An optional class to apply to the tooltip message element.
    */
   tooltipClassName?: string;
+  /**
+   * The z-index value of the tooltip message element.
+   */
+  zIndex?: number;
 };
 
 const getPositionStyle = (
@@ -175,6 +179,7 @@ export const adjustForWindow = (
  * @param [position="top-left"] Position of the tooltip relative to the element.
  * @param positionElementClassName An optional class to apply to the element that wraps the children.
  * @param tooltipClassName An optional class to apply to the tooltip message element.
+ * @param zIndex The z-index value of the tooltip message element.
  */
 const Tooltip = ({
   autoAdjust = true,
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -328,6 +334,7 @@ const Tooltip = ({
                 className="p-tooltip__message"
                 ref={messageRef}
                 id={tooltipId}
+                style={{ zIndex: zIndex }}
```
bartaz, line 337: What would happen if someone passes invalid value here?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -70,6 +70,10 @@ export type Props = {
    * An optional class to apply to the tooltip message element.
    */
   tooltipClassName?: string;
+  /**
+   * The z-index value of the tooltip message element.
+   */
+  zIndex?: string;
```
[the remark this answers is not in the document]
lorumic: Not a particularly important reason, it's just that these properties are more often strings and I thought it would be easier to pass it as `zIndex="999"` rather than `zIndex={999}`. I can convert it to a number if it can be helpful to avoid passing invalid values - as per your other comment.

src/components/Tooltip/Tooltip.tsx
lorumic: I just tested, the invalid value would be ignored. But maybe, as you mentioned in the other comment, this risk can be avoided by setting the `zIndex` prop's type to `number`. I will change that.

bartaz:
@lorumic Could you add a simple test to https://github.com/canonical/react-components/blob/main/src/components/Tooltip/Tooltip.test.tsx to make sure z-index is passed correctly to correct element?

lorumic:
> @lorumic Could you add a simple test to https://github.com/canonical/react-components/blob/main/src/components/Tooltip/Tooltip.test.tsx to make sure z-index is passed correctly to correct element?

Done

ClementChaumel (approved):
LGTM

Issue #897, opened 2023-03-10
Tooltip adding hidden elements when closed


petermakowski:
Currently the Tooltip component is adding a hidden element to the DOM that’s displayed off screen even when a tooltip is closed. 

This hinders performance especially when a large number of tooltips is rendered at a time (e.g. MAAS machine list).

As an example, on machine listing page with 200 machines that have elements with tooltips once the tooltip is removed from the DOM when closed the render time goes down to ~330ms down from ~450ms.

PR #898, opened 2023-03-10
fix: render tooltip only when open

petermakowski:
## Done

- fix: render tooltip only when open
- this change removes the tooltip elements from the DOM when not open

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open the demo
- Make sure the tooltip works exactly the same way as before

## Fixes

Fixes: https://github.com/canonical/react-components/issues/897

### DIFF
--- a/src/components/Tooltip/Tooltip.test.tsx
+++ b/src/components/Tooltip/Tooltip.test.tsx
@@ -36,12 +36,13 @@ describe("Tooltip", () => {
     expect(screen.getByRole("link", { name: "Canonical" })).toHaveFocus();
   });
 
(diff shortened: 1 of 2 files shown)


bartaz:
@petermakowski Was this causing some issues that tooltip was in DOM, but hidden?

petermakowski:
> @petermakowski Was this causing some issues that tooltip was in DOM, but hidden?

Yes, described in the linked issue. https://github.com/canonical/react-components/issues/897

PR #902, opened 2023-03-15
Notifications: allow override of title element, as the h5 might disturb the heading hierarchy

edlerd:
## Done

- Make the wrapper tag for Notification title configurable. The default `h5` might disturb the heading hierarchy of the page.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- check that notifications still render the same as before

### DIFF
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import React, { useEffect, useRef } from "react";
+import React, { ElementType, useEffect, useRef } from "react";
 import type { HTMLProps, ReactNode } from "react";
 
 import Button, { ButtonAppearance } from "../Button";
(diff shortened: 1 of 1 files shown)


src/components/Notification/Notification.tsx
```diff
@@ -97,6 +101,7 @@ const Notification = ({
   timeout,
   timestamp,
   title,
+  titleElement: Component = "h5",
```
ClementChaumel, line 104: Here I feel like renaming the prop makes the render bit slightly more confusing

ClementChaumel: Ah, right. Got it

edlerd: Renamed this to `TitleComponent` just now.

edlerd:
Answered all the open questions above.
I rebased on main and the cypress tests started to fail. Seems like Tooltip tests in main are broken atm.

ClementChaumel:
> Answered all the open questions above. I rebased on main and the cypress tests started to fail. Seems like Tooltip tests in main are broken atm.

Thanks! I'll look at fixing the issue with tooltips then we can merge this!

ClementChaumel (approved):
Nice work 👌️

PR #907, opened 2023-03-27
Add sub-rows to ModularTable

huwshimi:
## Done

- Added support for the React Table [subRows prop](https://react-table-v7.tanstack.com/docs/api/useTable#row-properties). It was previously possible to pass in this prop, but it didn't do anything.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the new [subrows example in Storyboard](/?path=/docs/modulartable--subrows).
- Check that you can sort the rows and that the subrows sort within the groups.

## Fixes

This will be used to support sorting the grouped action log rows in Juju Dashboard: https://github.com/canonical/jaas-dashboard/issues/1033.

### DIFF
--- a/src/components/ModularTable/ModularTable.tsx
+++ b/src/components/ModularTable/ModularTable.tsx
@@ -65,7 +65,58 @@ export type Props<D extends Record<string, unknown>> = PropsWithSpread<
   HTMLProps<HTMLTableElement>
 >;
 
-function ModularTable({
+const generateCell = <D extends Record<string, unknown>>(
+  cell: Cell<D>,
+  getCellProps: Props<D>["getCellProps"]
+) => {
+  const hasColumnIcon = cell.column.getCellIcon;
+  const iconName = hasColumnIcon && cell.column.getCellIcon?.(cell);
+
+  return (
+    <TableCell
+      {...cell.getCellProps([
+        {
+          className: cell.column.className,
+        },
+        {
+          className: hasColumnIcon ? "p-table__cell--icon-placeholder" : "",
+        },
+        { ...getCellProps?.(cell) },
+      ])}
+    >
+      {iconName && <Icon name={iconName} />}
+      {cell.render("Cell")}
+    </TableCell>
+  );
+};
+
+const generateRows = <D extends Record<string, unknown>>(
+  rows: Row<D>[],
+  prepareRow: (row: Row<D>) => void,
+  getRowProps: Props<D>["getRowProps"],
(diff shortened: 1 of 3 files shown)


src/components/ModularTable/ModularTable.tsx
```diff
@@ -65,7 +65,58 @@ export type Props<D extends Record<string, unknown>> = PropsWithSpread<
...
+    tableRows.push(
+      <TableRow {...row.getRowProps(getRowProps?.(row))}>
+        {row.cells.map((cell) => generateCell<D>(cell, getCellProps))}
+      </TableRow>
+    );
+    if (row.subRows?.length) {
+      tableRows = tableRows.concat(
+        generateRows<D>(row.subRows, prepareRow, getRowProps, getCellProps)
+      );
+    }
```
huwshimi, line 114: This is the only change to support subRows, the rest of the changes were moving the row generation to a function so that the main rows and sub-rows can both be generated by the same func.

ClementChaumel:
Looks good, I just noticed during QA that when sorting, the rows get swapped depending on which sub row should be first.
Here is what I mean:
Peek 2023-04-17 16-28

Is that the expected behaviour?

huwshimi:
> Looks good, I just noticed during QA that when sorting, the rows get swapped depending on which sub row should be first. Here is what I mean: Peek 2023-04-17 16-28 [ Peek 2023-04-17 16-28 ](https://user-images.githubusercontent.com/11927929/232517246-9307ad64-e0da-420e-a70e-f0f62565097d.gif) [ ](https://user-images.githubusercontent.com/11927929/232517246-9307ad64-e0da-420e-a70e-f0f62565097d.gif)
> 
> Is that the expected behaviour?

What's happening there is that the normal rows are being sorted first and then the subrows are getting sorted after that. So in that example sorbet and ice-cream are being sorted by their scoops value which moves the groups around and then the flavours (sub-rows) get sorted within the group.

ClementChaumel:
> 
> What's happening there is that the normal rows are being sorted first and then the subrows are getting sorted after that. So in that example sorbet and ice-cream are being sorted by their scoops value which moves the groups around and then the flavours (sub-rows) get sorted within the group.

Cool yeah I figured! I just wanted to check that you knew about it and that was the expected behaviour!

ClementChaumel (approved):
LGTM!

Issue #911, opened 2023-04-03
Clicking on tooltip fires parent onClick


huwshimi:
When a tooltip message is clicked, such as selecting the text or clicking a link, the click event is propagated to any parent click handlers.

PR #912, opened 2023-04-03
Fix tooltip click propagation

huwshimi:
## Done

- Prevent clicks on the tooltip message from propagating to parent click handlers.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the tooltip stories and add `onClick={() => console.log("parent clicked")}` to the wrapping template div: https://github.com/canonical/react-components/blob/main/src/components/Tooltip/Tooltip.stories.mdx?plain=1#L20.
- Open storybook and go to the tooltip story.
- Hover the button and then click the tooltip or the link inside the tooltip.
- The "parent clicked" message should not get logged.

## Fixes

Fixes: #911.

### DIFF
--- a/src/components/Tooltip/Tooltip.test.tsx
+++ b/src/components/Tooltip/Tooltip.test.tsx
@@ -78,6 +78,48 @@ describe("Tooltip", () => {
     expect(clickHandler).toHaveBeenCalled();
   });
 
+  it("does not propogate clicks on the tooltip message to the parent", async () => {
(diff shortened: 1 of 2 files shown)


bartaz:
Interesting, I didn't realise events propagate through portals…

goulinkh (approved):
LGTM

PR #913, opened 2023-04-17
Fix table sorting

edlerd:
## Done

- When clicking a `MainTable` heading that is not sortable, we used to reset to no sorting. Fixed that.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Try sorting a [main table](https://react-components-913.demos.haus/?path=/story/maintable--sortable) and click on headers that are sortable and not-sortable by. The network columns is not sortable and clicking it should not change the sort direction. This used to be broken before.


ClementChaumel:
Looks good! Can you add a test to cover the change?

edlerd:
> Looks good! Can you add a test to cover the change?

Sure, added a test case.

ClementChaumel (approved):
LGTM

ClementChaumel:
> > Looks good! Can you add a test to cover the change?
> 
> Sure, added a test case.

Thanks!

PR #919, opened 2023-05-09
[SearchAndFilter] add onPanelToggle and onExpandChange event handlers

lorumic:
## Done

- Added the following 2 optional callback props to the `SearchAndFilter` component:
1. `onPanelToggle`: triggered when the filter panel is shown/hidden.
2. `onExpandChange`: triggered when the search box has overflowing chips (= chips going new-line) and is expanded/collapsed.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Add a `SearchAndFilter` component with the `onPanelToggle` prop set.
- Check that when clicking on the search box to show the filter panel and when clicking outside to hide it, the `onPanelToggle` function is called.
- Add a `SearchAndFilter` component with the `onExpandChange` prop set.
- Add many chips so that they overflow in the search box.
- Check that when expanding/collapsing the search box, the `onExpandChange` function is called.

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.tsx
+++ b/src/components/SearchAndFilter/SearchAndFilter.tsx
@@ -26,6 +26,10 @@ export type Props = {
    */
   returnSearchData: (searchData: SearchAndFilterChip[]) => void;
   /**
+   * A function that is called when the filter panel is shown/hidden.
+   */
+  onPanelToggle?: () => void;
+  /**
    * A function that is called when the height of the search container changes.
    */
   onHeightChange?: () => void;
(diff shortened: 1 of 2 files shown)


src/components/SearchAndFilter/SearchAndFilter.tsx
```diff
@@ -25,12 +25,17 @@ export type Props = {
    * A function that is called when the search data changes.
    */
   returnSearchData: (searchData: SearchAndFilterChip[]) => void;
+  /**
+   * A function that is called when the height of the search container changes.
+   */
+  onHeightChange?: () => void;
```
edlerd, line 31: Maybe better `onSearchBoxExpand`?

lorumic: I don't think that name makes it clear that the function is going to be called also when the search box is collapsed, not only when it's expanded.

edlerd (approved):
LGTM

PR #925, opened 2023-06-02
fix(card) avoid aria-labelledby reference when no title is passed

edlerd:
## Done

- avoid broken reference to label on a card with no title


petermakowski (approved):
LGTM

PR #926, opened 2023-06-02
fix(tooltip) avoid broken reference to tooltip when not mounted

edlerd:
## Done

- avoid setting aria ref from tooltip anchor when the tooltip is not rendered


petermakowski (approved):
LGTM

PR #928, opened 2023-06-09
Forward close callback to children of ContextualMenu WD-4203

edlerd:
## Done

- Forward close callback to children of ContextualMenu

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check new contextualmenu--toggle-no-links entry in stroybook

## Fixes

Fixes: WD-4203


src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -224,7 +224,7 @@ const ContextualMenuDropdown = <L,>({
         }
       >
         {dropdownContent
-          ? dropdownContent
+          ? React.cloneElement(dropdownContent as any, { close: closePortal })
```
huwshimi, line 227: This will cause issues for any component that does not handle the `close` prop as it will try and pass it to the DOM (see the test error here: https://github.com/canonical/react-components/actions/runs/5221836743/jobs/9426601147?pr=928#step:7:3735).

Maybe this prop should accept either a `ReactNode` or a function that accepts a close prop and returns a `ReactNode`. That way it would be backwards compatible and also we'd get type checking of the render function.

src/components/ContextualMenu/ContextualMenu.test.tsx
```diff
@@ -266,4 +267,24 @@ describe("ContextualMenu ", () => {
     );
     expect(screen.getByText("toggleLabel component text")).toBeInTheDocument();
   });
+
+  it("forwards close callback to children that are not links", async () => {
+    const ContextMenuContent: FC<{ close?: () => void }> = ({ close }) => (
```
huwshimi, line 272: Only a very minor thing, but we don't typically use `FC` in this project (though I think the `FC` type is better in React 18). Not a big deal being as this is in a test, but this could just be `const ContextMenuContent = ({ close }: { close?: () => void }) => (`.

huwshimi (commented):
This is looking good, just needs a fix to handle dropdowns that don't support the `close` prop.

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -224,7 +230,9 @@ const ContextualMenuDropdown = <L,>({
         }
       >
         {dropdownContent
-          ? dropdownContent
+          ? typeof dropdownContent === "function"
+            ? React.cloneElement(dropdownContent(closePortal))
```
huwshimi, line 234: Does this still need `cloneElement` on the result of the function?

src/components/ContextualMenu/ContextualMenu.test.tsx
```diff
@@ -266,4 +267,20 @@ describe("ContextualMenu ", () => {
     );
     expect(screen.getByText("toggleLabel component text")).toBeInTheDocument();
   });
+
+  it("forwards close callback to child when it is a function", async () => {
```
huwshimi, line 271: Maybe we should also have a test for the ReactNode case as well.

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -29,7 +35,7 @@ export type Props<L> = PropsWithSpread<
     /**
      * The menu content (if the links prop is not supplied).
      */
-    children?: ReactNode;
+    children?: ReactNode | ((close: () => void) => ReactElement);
```
huwshimi, line 38: This could be `ContextualMenuDropdownProps["children"]` to save some duplication.

huwshimi (approved):
Thanks for those fixes @edlerd. Just a couple more small comments.

edlerd:
> Thanks for those fixes @edlerd. Just a couple more small comments.

Thanks for the review. I like how simple and clear the change has become with the adjustments.

PR #931, opened 2023-06-19
Add notification provider WD-4256

edlerd:
## Done

- Added NotificationProvider.
- Allows to trigger messages with the `const notify = useNotify();` hook easily, like  `notify.success("The thing finished` or `notify.failure("the thing failed", exception)`.
- Added a convenient `<NotifyConsumer>`, that can be placed anywhere and will render any triggered notification. This can be placed flexibly on top of a form, in a modal or used in an overlay.

Fixes: [WD-4256](https://warthogs.atlassian.net/browse/WD-4256)

[WD-4256]: https://warthogs.atlassian.net/browse/WD-4256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/NotificationProvider/NotificationProvider.test.tsx
+++ b/src/components/NotificationProvider/NotificationProvider.test.tsx
@@ -1,0 +1,79 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import {
+  NotificationConsumer,
+  NotificationProvider,
+  useNotify,
+} from "./NotificationProvider";
+import { BrowserRouter } from "react-router-dom";
+import Button from "../Button";
+import { act } from "react-dom/test-utils";
+
+describe("NotificationProvider", () => {
+  it("stores and renders notifications", async () => {
+    const handleRetry = jest.fn();
+
+    const TriggerComponent = () => {
+      const notify = useNotify();
+      const handleSuccess = () => notify.success("My success!");
+      const handleClear = () => notify.clear();
+      const handleInfo = () => notify.info("Some more details", "Test info");
+      const handleError = () =>
+        notify.failure(
+          "Fail title",
+          new Error("error message"),
+          "Button caused a failure",
+          [
+            {
+              label: "Retry",
+              onClick: handleRetry,
+            },
+          ]
+        );
+
+      return (
+        <div>
+          <Button onClick={handleSuccess} data-testid="success-btn" />
+          <Button onClick={handleClear} data-testid="clear-btn" />
+          <Button onClick={handleError} data-testid="error-btn" />
+          <Button onClick={handleInfo} data-testid="info-btn" />
+        </div>
+      );
+    };
+
+    render(
+      <div>
+        <NotificationProvider>
+          <TriggerComponent />
+          <div data-testid="notification-consumer">
+            <NotificationConsumer />
+          </div>
+        </NotificationProvider>
+      </div>,
+      { wrapper: BrowserRouter }
+    );
+
+    const clickBtn = async (testId: string) =>
+      await act(async () => await userEvent.click(screen.getByTestId(testId)));
+
+    expect(screen.getByTestId("notification-consumer")).toBeEmptyDOMElement();
+
+    await clickBtn("success-btn");
+    expect(screen.getByTestId("notification-consumer")).toMatchSnapshot();
+
+    await clickBtn("clear-btn");
+    expect(screen.getByTestId("notification-consumer")).toBeEmptyDOMElement();
+
+    await clickBtn("error-btn");
+    expect(screen.getByTestId("notification-consumer")).toMatchSnapshot();
+
+    expect(handleRetry).not.toHaveBeenCalled();
+    await clickBtn("notification-action");
+    expect(handleRetry).toHaveBeenCalled();
+
+    await clickBtn("info-btn");
+    expect(screen.getByTestId("notification-consumer")).toMatchSnapshot();
+  });
+});
--- a/src/components/NotificationProvider/NotificationProvider.tsx
+++ b/src/components/NotificationProvider/NotificationProvider.tsx
@@ -1,0 +1,103 @@
+import React, {
+  createContext,
+  FC,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
+import {
+  NotificationType,
+  NotificationHelper,
+  QueuedNotification,
+  NotifyProviderProps,
+} from "./types";
+import { useLocation } from "react-router-dom";
+import isEqual from "lodash/isEqual";
+import { info, failure, success, queue } from "./messageBuilder";
+import Notification, { DefaultTitles } from "../Notification/Notification";
+
+const NotifyContext = createContext<NotificationHelper>({
+  notification: null,
+  clear: () => undefined,
+  failure: () => undefined,
+  success: () => undefined,
+  info: () => undefined,
+  queue: () => undefined,
+});
+
+export const NotificationProvider: FC<NotifyProviderProps> = ({ children }) => {
+  const [notification, setNotification] = useState<NotificationType | null>(
+    null
+  );
+  const { state, pathname } = useLocation() as QueuedNotification;
+
+  const clear = () => notification !== null && setNotification(null);
+
+  const setDeduplicated = (value: NotificationType) => {
+    if (!isEqual(value, notification)) {
+      setNotification(value);
+    }
+    return value;
+  };
+
+  useEffect(() => {
+    if (state?.queuedNotification) {
+      setDeduplicated(state.queuedNotification);
+      window.history.replaceState({}, "");
+    } else {
+      clear();
+    }
+  }, [state, pathname]);
+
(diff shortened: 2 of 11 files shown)


src/components/NotificationProvider/messageBuilder.tsx
```diff
@@ -0,0 +1,49 @@
...
+  QueuedNotification,
+} from "./types";
+import React, { ReactNode } from "react";
+
+export const queue = (notification: NotificationType): QueuedNotification => {
+  return { state: { queuedNotification: notification } };
+};
+
+export const info = (
+  message: string | ReactNode,
```
lorumic, line 13: I think we can do without the `string | ReactNode`, here and in the other places, since a `ReactNode` can implicitly already be a string.
```suggestion
  message: ReactNode,
```

src/components/NotificationProvider/messageBuilder.tsx
```diff
@@ -0,0 +1,49 @@
...
+  title?: string
+): NotificationType => {
+  return {
+    message,
+    title,
+    type: "information",
```
lorumic, line 19: I think we can avoid hard-coded values here, and reference the value from the `NotificationSeverity` enum (to be imported):
```suggestion
    type: NotificationSeverity.INFORMATION,
```
And similarly for the other methods below.

src/components/NotificationProvider/messageBuilder.tsx
```diff
@@ -0,0 +1,49 @@
...
+  message?: string | ReactNode,
+  actions?: NotificationAction[]
+): NotificationType => {
+  return {
+    actions,
+    message:
+      error && typeof error === "object" && "message" in error ? (
+        <>
+          {message} {error.message}
+        </>
```
lorumic, line 42: We're forcing the `error` object to have some very specific "shape" here in order to display a detailed error message. Maybe we should define a type for it at this point, and remove the `unknown`?

src/components/NotificationProvider/NotificationProvider.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+      const handleInfo = () => notify.info("Some more details", "Test info");
+      const handleError = () =>
+        notify.failure(
+          "Fail title",
+          new Error("error message"),
+          "Button caused a failure",
+          [
+            {
+              label: "Retry",
+              onClick: () => {},
```
lorumic, line 29: Would it make sense to also test this function? With `toHaveBeenCalled`?

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+    if (title) {
+      return title;
+    }
+    switch (type) {
+      case "negative":
+        return "Error";
+      case "positive":
+        return "Success";
+      case "information":
+        return "Info";
```
lorumic, line 101: Same comment here, about the hard-coded values of the `NotificationSeverity` enum. Moreover, maybe also these "default" titles can have their dedicated enum and then just be referenced here?

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+      default:
+        return "";
```
lorumic, line 103: Let's maybe return `undefined` here and avoid the `figureTitle() || undefined` below?

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+      });
+    }
+    window.dispatchEvent(new Event("resize"));
+  }, [notify.notification]);
+
+  if (!notify.notification) {
+    return null;
+  }
+  const { actions, title, type, message } = notify.notification;
+  const figureTitle = () => {
```
lorumic, line 91: Personally I don't really like functions named `figureSomething`. I'd rather name this something like `getTitleOrDefault`.

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  useEffect(() => {
+    if (ref.current?.hasAttribute("scrollIntoView")) {
+      ref.current.scrollIntoView({
+        behavior: "smooth",
+        block: "center",
+        inline: "start",
+      });
+    }
+    window.dispatchEvent(new Event("resize"));
```
lorumic, line 84: I wonder if this should dispatch a custom event instead? So that it can be intercepted and handled separately than the standard `resize` event. Maybe that way it would be less "hacky"? What do you think?

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <NotifyContext.Provider value={helper}>{children}</NotifyContext.Provider>
+  );
+};
+
+export function useNotify() {
+  return useContext(NotifyContext);
+}
+
+export const NotificationConsumer: FC = () => {
```
lorumic, line 72: I'm not convinced by the way this was integrated in the react-components code base. I think the hook should be separated, and placed in a dedicated file inside the "hooks" folder. The FC only should be nested under components instead.

lorumic:
This is a nice wrapper around the `Notification` component, to make it easier and quicker to use and hopefully many projects can benefit of it. I suggest the other reviewers to also have a look at the code, and maybe come up with some ideas and suggestions? Can't this be nicely used in your projects as well?

src/components/NotificationProvider/messageBuilder.tsx
edlerd: When using `.catch((error) => ...` there is often no type information on the error available. We want to allow forwarding the error directly in that cases, hence we type it as unkown here, to allow `.catch((error) => notify.failure("some message", error))`

lorumic: My point was that currently this only works when the error object that we pass has this form:
```
{
  "message": "Something"
}
```
Which is unlikely to always be the case. I think this implementation is not robust and should be improved. We should not rely on the "message" key to be there to display a detailed error message. The key might be called "msg" or whatnot and then this would not work.

edlerd: The JS `Error` type has the message field, so when fetching a `throw new Error("foo")`, the code will pick up the right message. JS does not have proper type checking, and you can throw any object. But I wouldn't want to support (or know how to render) custom objects beeing thrown. I think expecting Errors (or the message field to be present) is a useful behavior.

src/components/NotificationProvider/NotificationProvider.tsx
edlerd: I think we shouldn't throw this event at all - we can bind the `notify.message` as a dependency to a `useEffect` if we want to run something in a caller on changing notifications.

src/components/NotificationProvider/messageBuilder.tsx
lorumic: Ok, I get your motivations. So maybe this can be simplified a bit like this?
```suggestion
      error && error instanceof Error ? (
        <>
          {message} {error.message}
        </>
```
Since the motivation behind is "we assume the `error` to be an `Error` type, with a `message` field in it", right?

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  if (!notify.notification) {
+    return null;
+  }
+  const { actions, title, type, message } = notify.notification;
+
+  return (
+    <div ref={ref}>
+      <Notification
+        title={title ? title : DefaultTitles[type]}
```
lorumic, line 94: What about:
```suggestion
        title={title ?? DefaultTitles[type]}
```
?

src/components/NotificationProvider/NotificationProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  useEffect(() => {
+    if (state?.queuedNotification) {
+      setDeduplicated(state.queuedNotification);
+      window.history.replaceState({}, "");
+    } else {
+      clear();
+    }
+  }, [state, pathname]);
```
lorumic, line 51: Apparently this makes the linter a bit unhappy:
```
  51:6  warning  React Hook useEffect has missing dependencies: 'clear' and 'setDeduplicated'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```
It's just a warning, but maybe we can fix it?

src/components/NotificationProvider/NotificationProvider.tsx
lorumic: Sure, even better.

src/components/NotificationProvider/NotificationProvider.tsx
edlerd: Including them would run the hook too often so we'll have to accept the warning.

lorumic (approved):
Thanks for applying all the changes

PR #933, opened 2023-06-28
fix(ActionButton) reset height and width after loading finished

edlerd:
## Done

- reset height and width after loading finished on the ActionButton
- it used to leak setting height and width on load and keeping it set in the dom


lorumic (approved):
LGTM

PR #934, opened 2023-07-04
fix(theme) use default colors for the custom theme WD-4880

edlerd:
## Done

- use default colors for the custom theme
- fixes the contrast for the navigation font

### Storybook

```shell
yarn start
```

### QA steps

- check storybook navigation colours

## Fixes

Fixes: WD-4880

### DIFF
--- a/.storybook/vanillaish.js
+++ b/.storybook/vanillaish.js
@@ -3,13 +3,7 @@ import { create } from "@storybook/theming/create";
 export default create({
   base: "light",
 
-  colorPrimary: "#fff",
-  colorSecondary: "rgba(0,0,0,0.05)",
-
   // UI
-  appBg: "#fff",
-  appContentBg: "#fff",
-  appBorderColor: "#cdcdcd",
   appBorderRadius: ".125rem",
 
   // Typography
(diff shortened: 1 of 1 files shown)


edlerd:
Demo starting at https://react-components-934.demos.haus/

.storybook/vanillaish.js
```diff
@@ -3,33 +3,15 @@ import { create } from "@storybook/theming/create";
...
   base: "light",
 
-  colorPrimary: "#fff",
-  colorSecondary: "rgba(0,0,0,0.05)",
-
   // UI
-  appBg: "#fff",
-  appContentBg: "#fff",
-  appBorderColor: "#cdcdcd",
   appBorderRadius: ".125rem",
```
bartaz, line 7: I'm not sure what it does, but we probably don't need that either. We are removing all rounded corners from Vanillla in 4.0.

.storybook/vanillaish.js
```diff
@@ -3,33 +3,15 @@ import { create } from "@storybook/theming/create";
...
-  // Toolbar default and active colors
-  barTextColor: "#111",
-  barSelectedColor: "#666",
-  barBg: "#fff",
-
   // Form colors
-  inputBg: "#fff",
-  inputBorder: "#cdcdcd",
-  inputTextColor: "#111",
   inputBorderRadius: ".125rem",
```
bartaz, line 15: Same here, we probably don't need that rounded corner.

bartaz (approved):
LGTM, with little comment

.storybook/vanillaish.js
edlerd: This is for the controls - without, it is even rounder. I think it is better to keep it.

.storybook/vanillaish.js
edlerd: The default is much more round, I think better to keep it.

Issue #945, opened 2023-07-17
Focus on confirm in ConfirmationModal


huwshimi:
It would be nice if there was a way to set the focus on the confirm button when opening the modal so that users can press enter to confirm.

This should probably only happen if the confirm is positive (maybe it should focus on cancel if the action is negative). Here's how we do something similar in Juju Dashboard (though it only happens for "neutral" actions): https://github.com/canonical/juju-dashboard/blob/main/src/components/ConfirmationModal/ConfirmationModal.tsx#L23.

bartaz:
As discussed in the PR from accessibility point of view if the action is destructible (or at least hard to revert) modal should not focus on the action button, but rather cancel/close (which happens right now I believe).

Confirmation modal by definition is meant to be used to confirm actions that may have irreversible consequences. 

If action is neutral, or easy to undo, why have confirmation modal at all? Can't it be triggered immediately (instead of opening the modal)?

bartaz:
Triage: closing this for now, as it doesn't seem to be needed as requested.

If we want to consider it as a feature, this likely needs to be raised with UX.

PR #949, opened 2023-07-23
Fix ConfirmationModal to focus on positive confirm button#945

britneywwc:
## Done

* Allow ConfirmationModal to focus on the positive confirm button when opening the modal, letting users press enter to confirm
* If confirm button is not positive, modal focus will be on the cancel button

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

* Use ConfirmationModal component and change`confirmButtonAppearance`  to `positive/negative/neutral`. Click enter to verify which buttons are clicked.

## Fixes

Fixes: #945

### DIFF
--- a/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -1,5 +1,5 @@
-import React, { MouseEvent, ReactElement } from "react";
-import type { ReactNode } from "react";
+import React, { MouseEvent, ReactElement, useRef, useEffect } from "react";
+import type { ReactNode, MutableRefObject, RefObject } from "react";
 import { PropsWithSpread, ValueOf } from "types";
 import Button, { ButtonAppearance } from "components/Button";
 import Modal, { ModalProps } from "components/Modal";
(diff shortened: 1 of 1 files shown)


src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -43,27 +43,50 @@ export const ConfirmationModal = ({
...
+      const positiveButton: HTMLButtonElement | null =
+        modalRef?.current?.querySelector("button.p-button--positive");
+      // If the modal has a positive button then focus on that after opening
+      // to make a better keyboard navigation experience.
+      if (positiveButton) {
+        console.log("positive button focus");
+        positiveButton.focus();
+      } else {
+        // If there is no button then focus on the modal wrapper.
+        modalRef.current.focus();
```
lorumic, line 59: This is already the default behaviour - no need to add an else block for it.

src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -43,27 +43,50 @@ export const ConfirmationModal = ({
...
+  }, []);
+
+  return (
+    <div
+      className="p-confirmation-modal"
+      ref={modalRef as RefObject<HTMLDivElement>}
     >
-      {children}
-    </Modal>
+      <Modal
```
lorumic, line 69: The additional `div` wrapper with the unused `p-confirmation-modal` class is unnecessary: we can set the ref to the `Modal` directly - which is a `div` itself.

src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -43,27 +43,50 @@ export const ConfirmationModal = ({
...
+  const modalRef: MutableRefObject<HTMLElement> = useRef(null);
+
+  useEffect(() => {
+    if (modalRef.current) {
+      const positiveButton: HTMLButtonElement | null =
+        modalRef?.current?.querySelector("button.p-button--positive");
+      // If the modal has a positive button then focus on that after opening
+      // to make a better keyboard navigation experience.
+      if (positiveButton) {
+        console.log("positive button focus");
```
lorumic, line 55: I think it's better to avoid `console.log` calls unless strictly necessary, and this doesn't seem to be the case.
```suggestion
```

bartaz:
@britneywwc Thanks for the suggestion. Can you explain a bit why you think such change is an improvement?

Our current focus rules with Modal component we are following the accessibility recommendations from https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/:

image

When opening modal the first focusable element is focused by default, and it's recommended for it to be a button that closes the modal (so it's easy to go back if it was opened by mistake).

It's even more important if the action on the modal is destructive. And with ConfirmationModal it seems by definition we ask user for confirmation because the action they are about to perform is hard or impossible to revert. In such case it's recommended for default focus NOT to be the main action, but cancel action, so it's easier to cancel than perform the action by mistake.

So from accessibility point of view, I don't think we want to focus on primary button by default.

Of course, these are not strict rules and there may be situations when focusing on something else will not have bad consequences. So it may be worth considering updating the Modal component, so that the default focused can be changed, but it should be used with caution and I don't know if we have currently a use case for such feature.

britneywwc:
@bartaz Thanks for your input. Initially I thought this change would be necessary due to users' expectations when they are prompted with a "positive" Confirm button. Although after reading the information you have shared, I agree that it is a much safer option to stay focused on a non-destructive option. If there is currently no use case for such feature then I will close this PR for now.

bartaz:
Thanks @britneywwc, it's always good to have such discussions, so if you have any other ideas for improvements feel free to propose them as issues or PRs!

PR #952, opened 2023-08-02
feat: Add autoResetSortBy prop to ModularTable

vladimir-cucu:
BREAKING CHANGE: The default value of `autoResetSortBy` in `ModularTable` was changed from `true` to `false`. The `autoResetSortBy` optional prop specifies whether the sort by needs to be reset after each data change.

## Done

- Added `autoResetSortBy` optional prop to `ModularTable` component.

### DIFF
--- a/src/components/ModularTable/ModularTable.tsx
+++ b/src/components/ModularTable/ModularTable.tsx
@@ -69,6 +69,10 @@ export type Props<D extends Record<string, unknown>> = PropsWithSpread<
      * The direction of the initial sort.
      */
     initialSortDirection?: "ascending" | "descending";
+    /**
+     * Whether the sort by needs to be reset after each data change.
+     */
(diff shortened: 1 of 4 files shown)


src/components/ModularTable/ModularTable.tsx
```diff
@@ -136,6 +140,7 @@ function ModularTable<D extends Record<string, unknown>>({
   getRowId,
   initialSortColumn,
   initialSortDirection,
+  autoResetSortBy = true,
```
huwshimi, line 143: I wonder if this should be a breaking change and set it to `false`. I'm not sure anyone expects the whole table to be reset when the data changes (this happens even if it's a small nested update to the data) and I can't really think of a situation where you would WANT it to reset. What do you think?

huwshimi (approved):
Just one comment about maybe making this a breaking change.

Issue #946, opened 2023-07-18
ModularTable blank headers show sort link


huwshimi:
When a ModularTable header is blank it should not be sortable, or if it has `disableSortBy: true` on a column the header should not show a hover state.

I think this is happening because the header still gets an `aria-sort` property which Vanilla uses to style the header.

This is the hover state when the header is empty:
<img width="380" alt="Screenshot 2023-07-18 at 10 15 34 am" src="https://github.com/canonical/react-components/assets/361637/181400b4-e97c-4c9f-a90b-adf988761389">

PR #955, opened 2023-08-03
fix: Remove conditionally the aria-sort attribute and ability to sort from ModularTable

vladimir-cucu:
## Done

- When sorting is disabled for a specific column and/or there is no text being rendered within the column header, the column shouldn't have an aria-sort attribute and clicking on the respective column header won't trigger any sorting action.

## Details

- Fixes: #946

### DIFF
--- a/src/components/ModularTable/ModularTable.tsx
+++ b/src/components/ModularTable/ModularTable.tsx
@@ -171,25 +171,34 @@ function ModularTable<D extends Record<string, unknown>>({
 
   const showEmpty: boolean = !!emptyMsg && (!rows || rows.length === 0);
 
-  // Fuction returns whether table can be sorted by a specific column.
+  // Function returns whether table can be sorted by a specific column.
+  // Returns true if sorting is enabled for the column and there is text
+  // being rendered within the column header, otherwise returns false.
   const isColumnSortable = (column: HeaderGroup<D>) => {
-    // Function returns whether an instance of ReactNode has text within it.
-    // Recursively goes over all children from the node and searches for text.
-    const hasText = (node: ReactNode): boolean => {
-      if (!isValidElement(node)) {
-        return (
-          (typeof node === "string" || typeof node === "number") &&
-          !!String(node).trim()
-        );
-      }
-      const children = node.props.children;
-      if (Array.isArray(children)) {
-        return children.some((child: ReactNode) => hasText(child));
-      }
-      return hasText(children);
-    };
+    if (!column.canSort) {
+      return false;
+    }
 
-    return column.canSort && hasText(column.render("Header"));
+    const headerNodesQueue: ReactNode[] = [column.render("Header")];
+    // Go through all header child nodes and search for text.
+    while (headerNodesQueue.length) {
+      const frontNode = headerNodesQueue.shift();
+      if (isValidElement(frontNode)) {
+        headerNodesQueue.push(frontNode.props.children);
+      }
+      if (
+        !isValidElement(frontNode) &&
+        (typeof frontNode === "string" || typeof frontNode === "number") &&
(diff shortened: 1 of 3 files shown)


huwshimi (commented):
It looks like there are a few cases that this doesn't quite handle correctly, not sure how to fix them, but this is what I saw:
1. if I set `Header: ""` or `Header: " "` then the headers are still sortable and show the arrows when I click on the header.
2. If I remove `Header` then it doesn't show the arrows but clicking on the header sorts the table.
3. If I use JSX in the header e.g. `Header: <i>Build Duration</i>` then it sorts when I click it but doesn't show the arrows.

src/components/ModularTable/ModularTable.tsx
```diff
@@ -171,20 +171,45 @@ function ModularTable<D extends Record<string, unknown>>({
 
   const showEmpty: boolean = !!emptyMsg && (!rows || rows.length === 0);
 
+  // Fuction returns whether table can be sorted by a specific column.
+  const isColumnSortable = (column: HeaderGroup<D>) => {
+    // Function returns whether an instance of ReactNode has text within it.
+    // Recursively goes over all children from the node and searches for text.
+    const hasText = (node: ReactNode): boolean => {
```
vladimir-cucu, line 178: I think we could potentially modify this part not to use recursion and instead simulate the call stack with a local stack/queue. However, I'm not really sure if this would be a better approach, as the code of the new approach might look less intuitive and we will call this function on the header, which we don't expect to be a deeply nested node and thus we won't really pollute the call stack with recursion anyway.

vladimir-cucu: The following implementation of `isColumnSortable` using a local queue instead of recursion might be a better approach:

```
const isColumnSortable = (column: HeaderGroup<D>) => {
    if (!column.canSort) {
      return false;
    }

    const headerNodeQueue: ReactNode[] = [column.render("Header")];
    while (headerNodeQueue.length) {
      const frontNode = headerNodeQueue.shift();
      if (isValidElement(frontNode)) {
        headerNodeQueue.push(frontNode.props.children);
      }
      if (!isValidElement(frontNode) && Array.isArray(frontNode)) {
        frontNode.forEach((node) => headerNodeQueue.push(node));
      }
      if (
        !isValidElement(frontNode) &&
        (typeof frontNode === "string" || typeof frontNode === "number") &&
        !!String(frontNode).trim()
      ) {
        return true;
      }
    }

    return false;
  };
```

src/components/ModularTable/ModularTable.tsx
```diff
@@ -171,20 +171,54 @@ function ModularTable<D extends Record<string, unknown>>({
...
+        !!String(frontNode).trim()
+      ) {
+        return true;
+      }
+      if (!isValidElement(frontNode) && Array.isArray(frontNode)) {
+        frontNode.forEach((childNode) => headerNodesQueue.push(childNode));
+      }
+    }
+
+    return false;
```
huwshimi, line 201: This works really nicely, but now that I see the complexity of this I wonder if we should only disable it if the user doesn't provide `Header` or provides a string of whitespace, but not disable it if they provide JSX that doesn't have any content (and if they want it disabled when they've provided empty JSX they need to pass `disableSortBy: true`)

If we did that then I think this could all be replaced with something like: `return typeof column.Header === "string" ? !!column.Header.trim() : isValidElement(column.Header)`.

This would also allow headers that have content provided by CSS e.g. icons such as `<i className="p-icon--smiley" />` (not sure if we actually do that anywhere but I can't see a reason to not allow it).

huwshimi (approved):
Just one suggestion about simplifying it a little.

PR #959, opened 2023-08-22
Support ReactNode in the CodeSnippet component.

huwshimi:
## Done

- Support a `ReactNode` as the value of the code to display in CodeSnippet.

## Fixes

https://warthogs.atlassian.net/browse/WD-5870

### DIFF
--- a/src/components/CodeSnippet/CodeSnippetBlock.tsx
+++ b/src/components/CodeSnippet/CodeSnippetBlock.tsx
@@ -21,7 +21,7 @@ export type Props = {
   /**
    * The code snippet to display.
    */
-  code: string;
+  code: ReactNode;
   /**
    * Content to show below the code snippet.
    */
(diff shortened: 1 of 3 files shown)


src/components/CodeSnippet/CodeSnippetBlock.tsx
```diff
@@ -59,13 +59,22 @@ export default function CodeSnippetBlock({
...
-  let numberedCode: JSX.Element[];
+  let numberedCode: ReactNode;
 
   if (isNumbered) {
     className += "--numbered";
 
     // wrap code lines in spans (and preserve the whitespace)
-    const lines = code.split(/\r?\n/);
+    let lines: ReactNode[];
+    if (Array.isArray(code)) {
```
vladimir-cucu, line 69: Not sure how necessary this is, but maybe we can add an example in docs for this?

vladimir-cucu (approved):
Looks good to me! 🙂

src/components/CodeSnippet/CodeSnippetBlock.tsx
huwshimi: Good idea. I don't think it's totally obvious how to use it, especially with numbered lines, so I've added an example.

PR #906, opened 2023-03-20
Scroll the search and filter dropdown

huwshimi:
## Done

- Constrain the search and filter dropdown to stay within the viewport.

### QA steps

- View the search and filter dropdown [default story in isolation](https://react-components-906.demos.haus/iframe.html?args=&id=search-and-filter--default-story&viewMode=story).
- Click on the input to open the dropdown.
- Resize your window so that the entire dropdown doesn't fit.
- The dropdown should resize so that it doesn't get cut off and the contents should scroll.

## Fixes

Fixes: https://github.com/canonical/jaas-dashboard/issues/707.

## Screenshots

Screen Recording 2023-03-20 at 4 27 18 pm

PR #961, opened 2023-08-29
Contextual menu scrolling

huwshimi:
## Done

- Add a prop to handle scrolling the contextual menu if it doesn't fit on the screen (handled in the same way as the search and filter dropdown: https://github.com/canonical/react-components/pull/906).

## QA

- Open `src/components/ContextualMenu/ContextualMenu.stories.mdx`.
- Scroll down to the "Toggle" story and duplicate the links objects so that there are quite a few links.
- add `scrollOverflow: true,` to the args.
- Run Storybook (`yarn start`).
- Open `<storybook-url>/iframe.html?id=contextualmenu--toggle&viewMode=story`
- Click on the button to open the menu.
- Resize your window so that the menu doesn't fit on the screen.
- The menu should resize so that it doesn't get cut off, and the content should scroll.

## Screenshots

Screenshot 2023-08-29 at 12 31 09 pm

### DIFF
--- a/src/components/ContextualMenu/ContextualMenu.tsx
+++ b/src/components/ContextualMenu/ContextualMenu.tsx
@@ -23,7 +23,7 @@ export enum Label {
 export type Props<L> = PropsWithSpread<
   {
     /**
-     * Whether the menu should adjust to fit in the screen.
+     * Whether the menu should adjust its horizontal position to fit on the screen.
      */
     autoAdjust?: boolean;
     /**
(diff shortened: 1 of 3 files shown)


huwshimi:
> I don't mind having it directly implemented in React components, but seems like it would be good to ensure it's well supported from Vanilla perspective.

If we do decide to support this in Vanilla, we should also address the other dropdowns, like the one for the search and filter pattern, as we also had to fix that one recently as well: https://github.com/canonical/react-components/pull/906.

huwshimi:
> I don't mind having it directly implemented in React components, but seems like it would be good to ensure it's well supported from Vanilla perspective.
> 
> Currently Vanilla doesn't directly support this, and while adding `max-height` and `overflow-x` from React works now, without official Vanilla support we can't ensure it will not break at some point if changes are made to contextual menu.
> 
> I guess we could at least add an example in Vanilla with such use case (and max-height/overflow added) to make sure we don't break it by regression. Or maybe `overflow-x` could be set to auto by default, in CSS (I'm not sure if it would break something when it's not needed).

I've filed in issue so that we can follow up on this: https://github.com/canonical/vanilla-framework/issues/4864.

Issue #944, opened 2023-07-13
Move docs out of main


huwshimi:
To simplify releasing the package we should move docs to a different branch and remove them from main.

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

We'll need to investigate only updating the docs when a new release is created.

bartaz:
Had a quick look and yes, configuring different branch is trivial:

<img width="887" alt="image" src="https://github.com/canonical/react-components/assets/83575/2acb3053-c274-43d1-aae9-930b44f4e7c1">

But we would need to have a workflow that pushes to that branch on release, but…

It seems that you don't even need to host your static code anymore, as you can use github actions to deploy:

<img width="968" alt="image" src="https://github.com/canonical/react-components/assets/83575/4ee816aa-644b-4525-b0be-3488044846ca">


Quite possibly all we need is to use the https://github.com/actions/deploy-pages action in our release workflow, but I didn't fully investigate how it works.

PR #963, opened 2023-08-30
Build docs using GitHub Actions

huwshimi:
## Done

- Remove docs build files from the repository.
- Set up GitHub Actions to publish the docs on release.

## Fixes

Fixes: #944.


huwshimi:
Once this lands we'll need to change the setting to use the action to build the docs instead of using the current build files.

.github/workflows/publish-on-release.yml
```diff
@@ -23,3 +23,28 @@ jobs:
       - run: npm publish --access public
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+  build-docs:
```
bartaz, line 26: I would add `needs: publish-npm` to make sure we don't update the docs if npm publish fails for any reason.

bartaz:
As this affects the release process there are some docs that may need updating. Like: https://github.com/canonical/react-components/blob/main/PUBLISH-NPM-PACKAGE.md

Also the "prepare-release" script mentioned there is built to automate the `build-docs` step, which is not needed. So the script may need updating (or removing completely).

I guess once we land this, the whole "Prepare release" PR would be reduced to changing version number, which in practice we could move into individual PRs that do the change (as in Vanilla), to remove the need of "Prepare release" PR completely.

Anyway, I'd suggest at least slightly updating the https://github.com/canonical/react-components/blob/main/PUBLISH-NPM-PACKAGE.md docs, to mention the changes, and maybe mention somewhere that docs are built automatically by GH action.

I also wonder if there is still a way to manually build the docs in case we need it outside of the release process.

bartaz (approved):
Looks good. I'd just add the dependency between docs job and publish to npm job as mentioned in inline comment, and possibly adjust the docs.

.github/workflows/publish-on-release.yml
huwshimi: Good thinking, I've updated that.

huwshimi:
> As this affects the release process there are some docs that may need updating. Like: https://github.com/canonical/react-components/blob/main/PUBLISH-NPM-PACKAGE.md

These docs didn't actually mention building the docs or automatically publishing to NPM, so I've added details about both of them.

> 
> Also the "prepare-release" script mentioned there is built to automate the `build-docs` step, which is not needed. So the script may need updating (or removing completely).

In this PR I removed the docs building from `prepare-release` but left the script to update the version, but I think just updating the version might be less confusing than running the script. I've updated the docs to reflect this and removed the `prepare-release` script.

> 
> I guess once we land this, the whole "Prepare release" PR would be reduced to changing version number, which in practice we could move into individual PRs that do the change (as in Vanilla), to remove the need of "Prepare release" PR completely.

I think that's a good idea.

> 
> Anyway, I'd suggest at least slightly updating the https://github.com/canonical/react-components/blob/main/PUBLISH-NPM-PACKAGE.md docs, to mention the changes, and maybe mention somewhere that docs are built automatically by GH action.

Done and done.

> 
> I also wonder if there is still a way to manually build the docs in case we need it outside of the release process.

Yes! You can configure an action to allow it to be manually run. I've done this and added a section to the docs.

PR #965, opened 2023-08-31
Separate docs workflow

huwshimi:
## Done

- Separate the workflow for publishing the docs on release, so that when it is manually run it doesn't also publish the package to NPM.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,36 @@
+name: Publish docs
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  build-docs:
+    name: Build Storybook docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Pages
+        id: pages
+        uses: actions/configure-pages@v3
+      - name: Install dependencies
+        run: yarn install
+      - name: Build Storybook
+        run: yarn build-docs
+      - name: Upload artifact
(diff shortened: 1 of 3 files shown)


bartaz:
If we make it a separate workflow, is there a way to trigger it automatically only when publishing to npm succeeded? 
As far as I understand they will simply run in parallel, right?

huwshimi:
> If we make it a separate workflow, is there a way to trigger it automatically only when publishing to npm succeeded? As far as I understand they will simply run in parallel, right?

I started to investigate that, but realised that for us to also do manual docs builds we wouldn't want it to need a NPM publish as well.

As I type this I've just remembered that you can share jobs between workflows. I'll investigate adding it back to the main publish workflow to run if the NPM publish success, and also have a second workflow that we can trigger manually.

bartaz:
Overall I don't remember last time when we did need to publish docs manually, so it's not a high priority requirement at the moment I think.
On the other hand, there have been situations when npm publish failed (due to version error, networking, etc), so I'd rather have docs deploy job dependent on the npm publish. So I'm fine with it all being in the single release workflow for now.

But thanks for investigating the options.

huwshimi:
> Overall I don't remember last time when we did need to publish docs manually, so it's not a high priority requirement at the moment I think. On the other hand, there have been situations when npm publish failed (due to version error, networking, etc), so I'd rather have docs deploy job dependent on the npm publish. So I'm fine with it all being in the single release workflow for now.
> 
> But thanks for investigating the options.

I've just updated it! It wasn't so hard (as long as I got it all right).

bartaz (approved):
LGTM, thanks. Let's try it out!

PR #971, opened 2023-09-01
Avoid duplicate first or last page button, when total pages is one above truncateThreshold

edlerd:
## Done

- fixed pagination edge case for when the `totalPages +1 = truncateThreshold` to avoid duplicating the button for the first or last page

### QA steps

- See tests are valid and passing

### DIFF
--- a/src/components/Pagination/Pagination.test.tsx
+++ b/src/components/Pagination/Pagination.test.tsx
@@ -407,4 +407,72 @@ describe("<Pagination />", () => {
     );
     expect(mockHandleBack).toHaveBeenCalledTimes(1);
   });
+
+  it("renders all pages on page 1 without duplicating first or last page", async () => {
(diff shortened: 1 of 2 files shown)


huwshimi (approved):
This looks great. It could be nice to also remove the ellipsis when on the first or last pages when all the items are displayed.

It seems to work fine if you're on the middle items:

Screenshot 2023-09-05 at 9 38 23 am

But at the start and end it displays ellipsis:

Screenshot 2023-09-05 at 9 38 13 am

edlerd:
> This looks great. It could be nice to also remove the ellipsis when on the first or last pages when all the items are displayed.

Good catch, I fixed this and added a test for this case as well.

huwshimi (approved):
Looks great! Thanks @edlerd.

Issue #973, opened 2023-09-07
Pressing "Esc" key doesn't close Modal


vladimir-cucu:
When pressing the "Esc" key on an opened `Modal`, it doesn't close the `Modal`. Instead, a console error is thrown (as can be seen in the attached screenshot). This issue also affects `ConfirmationModal`.

Current issue was observed on Version 116.0.5845.140 (Official Build) (64-bit) of Google Chrome on Ubuntu 22.04.3 LTS.

![Screenshot from 2023-09-07 23-23-24](https://github.com/canonical/react-components/assets/108150922/2d8fc271-1e31-4886-aca3-560a5b05e998)

PR #974, opened 2023-09-07
fix: Pressing Esc key closes Modal

vladimir-cucu:
## Done

- Pressing Esc key closes the `Modal` instead of throwing a console error.
- This PR fixes the same issue in `ConfirmationModal`.

## QA

- Open any `Modal` in Storybook and then press Esc key.
- The `Modal` should close with no error shown in developer console.

## Fixes

Fixes: #973

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -1,12 +1,6 @@
 import classNames from "classnames";
 import React, { ReactElement, useRef, useEffect } from "react";
-import type {
-  HTMLProps,
-  ReactNode,
-  MutableRefObject,
-  KeyboardEvent,
-  RefObject,
-} from "react";
+import type { HTMLProps, ReactNode, MutableRefObject, RefObject } from "react";
 import { ClassName, PropsWithSpread } from "types";
 import { useId } from "hooks/useId";
 
@@ -54,7 +48,7 @@ export const Modal = ({
 
   const modalRef: MutableRefObject<HTMLElement> = useRef(null);
   const focusableModalElements = useRef(null);
-  const handleTabKey = (e: KeyboardEvent<HTMLDivElement>) => {
+  const handleTabKey = (e: React.KeyboardEvent<HTMLDivElement>) => {
     if (focusableModalElements.current.length > 0) {
       const firstElement = focusableModalElements.current[0];
       const lastElement =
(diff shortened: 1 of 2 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -75,7 +75,15 @@ export const Modal = ({
   };
 
   const handleEscKey = (e: KeyboardEvent<HTMLDivElement>) => {
```
huwshimi, line 77: If you remove the `KeyboardEvent` from the import you could replace this with:

```suggestion
  const handleEscKey = (e: KeyboardEvent | React.KeyboardEvent<HTMLDivElement>) => {
```

This will make it easier to work with the event below.

src/components/Modal/Modal.tsx
```diff
@@ -75,7 +75,15 @@ export const Modal = ({
...
-    e.nativeEvent.stopImmediatePropagation();
+    if (e.nativeEvent?.stopImmediatePropagation) {
+      e.nativeEvent.stopImmediatePropagation();
+    } else if ((e as unknown as Event).stopImmediatePropagation) {
+      (e as unknown as Event).stopImmediatePropagation();
+    } else if (e.stopPropagation) {
+      e.stopPropagation();
+    } else {
+      console.log("Couldn't stop Esc key press propagation.");
+    }
```
huwshimi, line 86: With the type updated above then you could check for the method like this:

```suggestion
    if ("nativeEvent" in e && e.nativeEvent?.stopImmediatePropagation) {
      e.nativeEvent.stopImmediatePropagation();
    } else if ("stopImmediatePropagation" in e) {
      e.stopImmediatePropagation();
    } else if (e.stopPropagation) {
      e.stopPropagation();
    }
```

This means you don't have to cast to `unknown`. I also think we probably don't need the console.log.

huwshimi (commented):
This is looking good. I've made some suggestions to improve the types. I also think it would be good to add a test that the propagation doesn't occur (you'll need to render another element in the test that also listens for escape).

src/components/Modal/Modal.tsx
vladimir-cucu: I also think this approach is a lot better and removes the weird type conversion to `unknown`. Made this change and added a test for stopImmediatePropagation as suggested in 28d767dbf25ecbf215350ac3f504b018da9ec2ba.

huwshimi (approved):
Thanks, this looks good now.

PR #989, opened 2023-10-23
fix(ConfirmationModal): Use base appearance for cancel btn in ConfirmationModal

edlerd:
## Done

- use base appearance for cancel button in `ConfirmationModal`

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open the [confirmation modal](https://react-components-989.demos.haus/?path=/story/confirmationmodal--default-story) and ensure the cancel button has a base appearance without border.

### DIFF
--- a/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -48,7 +48,11 @@ export const ConfirmationModal = ({
       buttonRow={
         <>
           {confirmExtra}
-          <Button className="u-no-margin--bottom" onClick={props.close}>
+          <Button
+            appearance="base"
+            className="u-no-margin--bottom"
+            onClick={props.close}
+          >
             {cancelButtonLabel}
           </Button>
           <Button


lorumic (commented):
Code-wise this is good, but should probably pass from some UX and/or visual design first?

lorumic (approved):

acarege:
Hello dear team,

I think that the cancel button should be a default button for two reasons:

1. As this is a destructive modal the main button is negative. Which means that the opposite of that action is positive. According to Vanilla we use _"a default button to indicate a positive action that isn't necessarily the main call-to-action"_ which I think fits this case whereas _"a base button... is often used alongside a default button"_. So according to Vanilla guidelines a default button seems more appropriate.

2. In my personal opinion, when the "Cancel" button doesn't have borders it looks very light compared to the destructive one that is already dominant due to colours. That is something we don't want unless we're guiding user towards the destructive action. 

Cheers!

bartaz:
Rejected by design.

PR #994, opened 2023-11-09
ci: Add PR linting

petermakowski:
## Done
This PR adds a PR Lint action that enforces conventional commits convention for pull request titles. 

- ci: add pr lint

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: # .

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -1,0 +1,17 @@
+name: "Lint PR"
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


bartaz:
@petermakowski Would it be worth documenting something what is a valid PR name? or is that clear enough when it fails?

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,17 @@
...
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-22.04
```
bartaz, line 13: any reason not to use latest?

bartaz (approved):
Couple small questions in comments to discuss if needed, but it's fine as is as well.

petermakowski:
> Couple small questions in comments to discuss if needed, but it's fine as is as well.

Thanks, will think about updating the docs with more info next.

bartaz:
> > Couple small questions in comments to discuss if needed, but it's fine as is as well.
> 
> Thanks, will think about updating the docs with more info next.

Thanks. I tested how it fails, and error message is quite informative, so it's not a high priority, but once we get to the point of having automatic releases set up it would be good to have the complete process described somewhere.

.github/workflows/pr-lint.yml
petermakowski: We generally try to use exact dependencies for predictable runs. Happy to change if you feel strongly about this.

Issue #975, opened 2023-09-12
Show tooltips after slight delay


huwshimi:
Currently tooltips appear instantly, however this causes them to appear while you're moving your mouse past an element, sometimes causing the tooltip to block what you want to click on, and requiring the user to do a little dance to get to the right element.

![Screen Recording 2023-09-12 at 2 40 11 pm](https://github.com/canonical/react-components/assets/361637/bf0f6217-804a-4bdf-8ebd-a5f499b0cc19)

We could introduce a very small delay that requires the user to pause on an element before the tooltip appears.

Prateek32177:
Hi @huwshimi Can i work on this?

Do you need a hardcoded delay or a delay prop is required which gives additonal customisation for user to utilize the prop to set custom delay for opening the portal? addition to delay do we need to give immediate portal opening option also or only with delay is fine?

huwshimi:
> Hi @huwshimi Can i work on this?

Sure!


> Do you need a hardcoded delay or a delay prop is required which gives additonal customisation for user to utilize the prop to set custom delay for opening the portal?

By default I think the delay should be set in the few hundred milliseconds range, but I think it would be good to have a prop to allow changing the delay length.

> addition to delay do we need to give immediate portal opening option also or only with delay is fine?

I think the delay should be able to be disabled. I think this could be done using the delay length prop (above) e.g. `delay={0}`.

akmittal:
Hi @huwshimi 

I have created a PR to fix this https://github.com/canonical/react-components/pull/998. I am using default delay of 500ms.

PR #998, opened 2023-11-15
feat: Add tooltip delay(500ms by default) after mouseover

akmittal:
## Done

- Add delay of 500ms before showing tooltip after mouseover
- Delay is configurable from Props

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: #975  .

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -74,6 +74,10 @@ export type Props = {
    * The z-index value of the tooltip message element.
    */
   zIndex?: number;
+  /**
+   * Deplay in ms after which Tooltip will appear
+   */
+  delay?: number;
 };
 
 const getPositionStyle = (
@@ -191,6 +195,7 @@ const Tooltip = ({
   positionElementClassName,
   tooltipClassName,
   zIndex,
+  delay = 500,
 }: Props): JSX.Element => {
   const wrapperRef = useRef<HTMLElement>(null);
   const messageRef = useRef<HTMLElement>(null);
@@ -204,6 +209,13 @@ const Tooltip = ({
   });
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
   const tooltipId = useId();
+  const [timer, setTimer] = useState<ReturnType<typeof setTimeout> | null>(
+    null
+  );
+  const cancelableClosePortal = useCallback(() => {
+    clearTimeout(timer);
+    closePortal();
+  }, [timer, closePortal]);
 
   useEffect(() => {
     if (isOpen && !followMouse && wrapperRef.current) {
@@ -252,10 +264,10 @@ const Tooltip = ({
   const handleKeyPress = useCallback(
     (event: KeyboardEvent) => {
       if (event.key === "Escape") {
-        closePortal();
+        cancelableClosePortal();
       }
     },
-    [closePortal]
+    [cancelableClosePortal]
   );
 
   useEffect(() => {
@@ -276,7 +288,7 @@ const Tooltip = ({
         ? !messageRef.current?.contains(e.relatedTarget)
         : e.target !== messageRef.current
     ) {
-      closePortal();
+      cancelableClosePortal();
     }
   };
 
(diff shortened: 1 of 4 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -191,6 +195,7 @@ const Tooltip = ({
   positionElementClassName,
   tooltipClassName,
   zIndex,
+  delay = 500,
```
huwshimi, line 198: I wonder if 500 is a little too long? Playing around with the number it seemed like 350 was a good amount of time to me.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -289,6 +301,11 @@ const Tooltip = ({
     openPortal();
   };
 
+  const deplayedOpenPortal = useCallback(() => {
+    const timeout = setTimeout(openPortal, delay);
```
huwshimi, line 305: This is currently causing a console error when the tooltip is opened. This is because the mouse event is not passed to `openPortal`.

There are two possible fixes. One is to pass the event through to the openPortal call. The second would be to upgrade to the latest version of `react-useportal` and then change the `usePortal` call above to `usePortal({ programmaticallyOpen: true})`.

huwshimi (commented):
Thanks a lot for this PR @akmittal!

I've left a few comments and it would also be good to add a few tests:
- Check that the portal is not opened until the delay time.
- Check that blur before the delay time cancels the timeout.
- Check that setting the delay to `0` causes the tooltip to appear instantly.

To write those tests you will need to use Jest's fake timers.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -74,6 +74,10 @@ export type Props = {
    * The z-index value of the tooltip message element.
    */
   zIndex?: number;
+  /**
+   * Deplay in ms after which Tooltip will appear
```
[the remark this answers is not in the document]
akmittal: Made the change to add default value(350ms)

src/components/Tooltip/Tooltip.tsx
akmittal: Changed default to 350ms

src/components/Tooltip/Tooltip.tsx
akmittal: Went with the first option and passed `MouseEvent` to `openPortal `

akmittal:
> Thanks a lot for this PR @akmittal!
> 
> I've left a few comments and it would also be good to add a few tests:
> 
> * Check that the portal is not opened until the delay time.
> * Check that blur before the delay time cancels the timeout.
> * Check that setting the delay to `0` causes the tooltip to appear instantly.
> 
> To write those tests you will need to use Jest's fake timers.

Thanks for reviewing.
I have resolved review comments, still working on test cases.

huwshimi (commented):
This is looking great. One last issue: when the tests run they have an error `Warning: An update to Tooltip inside a test was not wrapped in act(...)` on most of the tests. To fix this you'll need to wrap the events that open the tooltip in `await act(async ...)` eg.

```javascript
    await act(async () => {
      await user.hover(screen.getByRole("button", { name: "Child" }));
    });
```

akmittal:
> This is looking great. One last issue: when the tests run they have an error `Warning: An update to Tooltip inside a test was not wrapped in act(...)` on most of the tests. To fix this you'll need to wrap the events that open the tooltip in `await act(async ...)` eg.
> 
> ```js
>     await act(async () => {
>       await user.hover(screen.getByRole("button", { name: "Child" }));
>     });
> ```

Completed the changes as suggested

huwshimi:
@akmittal thanks a lot for those fixes. The test run is showing some more act() warnings: https://github.com/canonical/react-components/actions/runs/6961479419/job/18948453606?pr=998#step:7:9.

I really appreciate you're work on this PR, it's very close to being complete, but I'm happy to fix up those act calls if you'd like me to.

akmittal:
> @akmittal thanks a lot for those fixes. The test run is showing some more act() warnings: https://github.com/canonical/react-components/actions/runs/6961479419/job/18948453606?pr=998#step:7:9.
> 
> I really appreciate you're work on this PR, it's very close to being complete, but I'm happy to fix up those act calls if you'd like me to.

Yes please go ahead

huwshimi (approved):
Thanks again for your contribution @akmittal! I've fixed the act warnings and slightly improved a couple of the tests. I'm going to merge this now.

Issue #958, opened 2023-08-21
Modal opens with the content already scrolled


andesol:
## Current behaviour

When you open the modal, text is already scrolled. You have to scroll up to read from the beginning.
![image](https://github.com/canonical/react-components/assets/20623084/a071d9c3-a70e-4708-b1f6-c7d0dee6685f)

## Expected behaviour
The modal's content starts off with the beginning of the text being visible. This seems to be the behaviour if you implement the modal using the css library.
![image](https://github.com/canonical/react-components/assets/20623084/44b5c256-1016-4542-a8af-f27459a694c0)

PR #999, opened 2023-11-20
fix(modal): do not scroll content when opened

andesol:
### Problem
The modal opens with the content scrolled (when there's enough content to be scrollable), so the user can't read from the beginning. It seems like this behaviour was introduced in [this commit](https://github.com/canonical/react-components/commit/6633f75a53f04114ec64435259c2017f2c435358), where the focus is moved to one of the buttons on the footer.

### Solution
Focus without scrolling into view. 

### QA steps

- `dotrun`
- Go to the `Modal` component in storybook
- Resize window till the modal becomes scrollable (or add some text to make it more obvious)
- Check that when the modal is opened the scrollbar is on top 

## Fixes

Fixes: #958

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -99,7 +99,7 @@ export const Modal = ({
     if (close && focusableModalElements.current.length > 1) {
       focusIndex = 1;
     }
-    focusableModalElements.current[focusIndex]?.focus();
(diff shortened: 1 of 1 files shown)


andesol:
Is the support for this option good enough? https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#browser_compatibility

edlerd (approved):
LGTM, thanks for the fix!

PR #1000, opened 2023-11-22
fix: contextual menu container overflow

petermakowski:
## Done

This pull request fixes the menu position when scrolling in overflowing containers.

It also simplifies the implementation by moving the dropdown element to the container.
This way, we're able to rely on CSS positioning for the menu and do not have to calculate it manually using JavaScript.

It also adds isEnabled options to existing hooks to be able to use them as a replacement for the logic that was provided by use-portal.

- fix contextual menu dropdown not being scrolled correctly in overflowing containers
- replace deprecated ClientRect with DOMRect
- fix dropdown props being passed incorrectly to the wrapper
- simplify implementation by removing the Portal and replacing it with useState
- add options to useOnEscapePressed and useClickOutside
- rename useClickOutside to useOnClickOutside for consistency
- remove the portal ref hack

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the overflow example
- Ensure that once you open the dropdown and scroll, menu position is in sync with the toggle
- Ensure you can switch autoAdjust on and off
- When enabled, Dropdown menu position should be adjusted in case of a long menu when position is set to the left.

## Fixes

Fixes: https://warthogs.atlassian.net/browse/MAASENG-2407


ndv99 (approved):
This looks good to me :+1:

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -221,48 +200,42 @@ const ContextualMenuDropdown = <L,>({
...
     <span
-      className={wrapperClass}
-      style={positionStyle as React.CSSProperties}
+      className={classNames("p-contextual-menu__dropdown", dropdownClassName)}
+      id={id}
+      aria-hidden={isOpen ? "false" : "true"}
+      aria-label={Label.Dropdown}
+      ref={dropdown}
+      style={{
+        ...(positionStyle as React.CSSProperties),
```
edlerd, line 209: I think there is no need to cast here.
```suggestion
        ...positionStyle,
```

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -246,45 +249,36 @@ const ContextualMenu = <L,>({
...
-        // Update the coordinates so that the menu stays relative to the
-        // toggle button.
-        updatePositionCoords();
-      }
-    },
-    [closePortal, positionNode, updatePositionCoords]
-  );
+  useOnClickOutside(wrapper, handleClose, {
+    isEnabled: closeOnOutsideClick,
+  });
```
edlerd, line 261: The `isEnabled` flag seems a bit unnatural to have in the hook.
We can let the client do that: `useOnClickOutside(wrapper, closeOnOutsideClick ? handleClose : () => {});`.

Same for the 'useOnEscapePressed` hook, that seemed cleaner without the flag.

edlerd (approved):
Looks good functionality wise. Two nitpicks about code style / simplifications. Feel free to address or keep as is.

huwshimi:
Hi @petermakowski, this was using a portal so that it could escape its container to prevent issues such as `overflow: hidden` on a parent (same as we do for tooltips, portals etc.) With this change you can see that the menus get cut off in the storybook examples:

Screenshot 2023-11-23 at 9 45 58 am

Screenshot 2023-11-23 at 9 51 48 am


Any ideas for mitigating that issue?

src/components/ContextualMenu/ContextualMenu.tsx
petermakowski: This is a pattern borrowed from react-query and I believe it's quite common: https://tanstack.com/query/v4/docs/react/guides/disabling-queries

huwshimi:
> > Hi @petermakowski, this was using a portal so that it could escape its container to prevent issues such as `overflow: hidden` on a parent (same as we do for tooltips, portals etc.)
> > Any ideas for mitigating that issue?
> 
> @huwshimi Thanks for chipping in. Very good point, clearly I was missing some context here. I intended for this to be fixed by removing `overflow: hidden` from a container, but this might not work in all cases.
> 
> In the examples that you pasted specifically this can be worked around by using automatic positioning and displaying the menu at the top of a trigger when not enough space is available at the bottom (the approach that we're exploring in MAAS currently). Having said that, it's a big change in approach and something that not all teams may want to adopt. I'll give this some more thought. Marking this as do not merge for now.

Thanks Peter, we've definitely had cases in the past where we couldn't control the overflow which is why we ended up using portals, but maybe there's another approach we haven't considered.

petermakowski:
@huwshimi This is ready for another review. I reverted changes related to relative positioning, keeping node position in sync with the container is now handled similarly to window resize.

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -255,6 +260,7 @@ const ContextualMenu = <L,>({
 
   const onResize = useCallback(
     (evt) => {
+      console.log("on resize!");
```
huwshimi, line 263: Oops, got a console.log here.

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -270,19 +276,27 @@ const ContextualMenu = <L,>({
...
   );
 
+  const onScroll = useCallback(
+    (e) => {
+      const parent = getPositionNode(wrapper.current, positionNode);
+      // update position if the scroll event is triggered by the parent of the menu
+      if (parent && e.target.contains(parent)) {
+        // Update the coordinates so that the menu stays relative to the
+        // toggle button.
+        console.log("update position!");
```
huwshimi, line 286: Oh and another console.log here too.

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -220,11 +221,9 @@ const ContextualMenuDropdown = <L,>({
   }, [adjustedPosition, updatePositionStyle]);
 
   return (
-    <span
-      className={wrapperClass}
-      style={positionStyle as React.CSSProperties}
-      {...props}
-    >
+    // Vanilla Framework uses .p-contextual-menu parent modifier classnames to determine the correct position of the .p-contextual-menu__dropdown dropdown (left, center, right).
+    // Extra span wrapper is required as the dropdown is rendered in a portal.
```
huwshimi, line 225: Thanks for this comment, it's really helpful.

huwshimi (approved):
Works great now, thanks for the improvement @petermakowski!

PR #1009, opened 2024-01-05
feat: adjust contextual dropdown vertical position

petermakowski:
## Done
This adjusts contextual dropdown vertical position based on the available space in the parent container. It's assuming that the container that should be used for the measurement is the first first scrollable parent of the dropdown. 

This is fairly computationally expensive, but because it happens on open it shouldn't cause issues in real-world scenarios.

- feat: adjust contextual dropdown vertical position
- add missing Button import to ContextualMenu story 

## QA
- Go to http://localhost:9009/?path=/story/contextualmenu--overflowing-container
- Verify that the first dropdown displays its content below the button and second one displays the content above.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

## Fixes

Fixes: https://warthogs.atlassian.net/browse/MAASENG-2428


src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -55,14 +56,18 @@ export type Props<L = null> = {
  */
 const getPositionStyle = (
   position: Position,
+  verticalPosition: VerticalPosition,
```
edlerd, line 59: In Tooltip we have combined positions of `btm-left`, `top-left`, `left`, etc. Maybe we should stick to the combined way of positioning? To stay backward compatible `left` can be the same as `btm-left`.

petermakowski: I did think about it, but decided against it to keep things simple. I'll try and give that a go.

petermakowski: I did think about it, but decided against it to keep things simple. I'll try and give that a go.

petermakowski: Having looked at it I'd rather treat this as a future improvement, I don't want to break the existing logic and it's already quite complex. To do this right it would need major refactoring.

As is, vertical position will be automatically adjusted and displayed at the bottom (which is as it works currently).

ndv99 (approved):
I'm happy to merge this, QA is solid, no noticeable issues in code either

Issue #1011, opened 2024-01-09
Input does not receive focus on label click without specified ID


petermakowski:
## Description:
When using the Input component, clicking on the associated label does not trigger focus on the input element if no id attribute is specified for the input. The current implementation requires an id to associate the label with the input correctly.

## Expected Behavior:
The Input component should automatically assign a unique id if one is not provided by the user. This ensures that the label and input elements are correctly associated, allowing for focus to be moved to the input when the label is clicked.

## Actual Behavior:
Clicking on the label does not move focus to the associated input when an id is not explicitly provided. This leads to accessibility violation.

## Steps to Reproduce:

1. Import the Input component from @canonical/react-components.
2. Render the Input component without specifying an id.
3. Try clicking on the label.
4. Observe that focus does not move to the input field.

## Suggested Solution:
Modify the Input component to automatically generate and assign a unique id when one is not provided. This will ensure proper label-input association as per accessibility standards.

bartaz:
@petermakowski It seems it was fixed for `Input` - would that be needed also for other input types? like select, textarea, etc?

petermakowski:
Quite possibly, I haven't checked. Depends how label is implemented there (from what I've seen it's usually not a wrapper around input which means it needs an id to be associated with this input).

PR #1012, opened 2024-01-09
fix: input label without specifying id

petermakowski:
## Done
This PR modifies the Input component to automatically generate and assign a unique id when one is not provided. This ensures correct label-input association as per accessibility standards.

- fix: input label without specifying id
  - replace instances of soon-to-be-deprecated `.toHaveErrorMessage` with `.toHaveAccessibleErrorMessage`

## QA steps
- [ ] Import the Input component from @canonical/react-components.
- [ ] Render the Input component without specifying an id.
- [ ] Click on the label.
- [ ] Make sure that input is in focus.

## Fixes

Fixes: https://github.com/canonical/react-components/issues/1011

### DIFF
--- a/src/components/Field/Field.test.tsx
+++ b/src/components/Field/Field.test.tsx
@@ -26,7 +26,7 @@ describe("Field ", () => {
         <input aria-errormessage="id-1" aria-invalid="true" />
       </Field>
     );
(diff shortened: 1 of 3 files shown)


ndv99 (approved):
LGTM

PR #1017, opened 2024-01-18
fix: move useLocation within useNotify hook to cater for latest react-router-dom limitations

mas-who:
## Done

- moved `useLocation` call from the `NotificationProvider` component into the `useNotify` hook. This circumvents the issue where react router no longer accepts usage of `useLocation` outside of the Router context.

## QA
-  I checked this by copying the change over to the lxd-ui repo, then QA'd there.


edlerd (approved):
LGTM, thanks for this fix for the react-router update.

PR #1018, opened 2024-01-18
fix: Confirmation modal can be very narrow #976

nadeem-R17:
## Done

- Given a minimum width to the confirmation modal so that width doesn't become narrow when there isn't a lot of content.

## Fixes

Fixes: #976 .

https://github.com/canonical/react-components/assets/90577359/50d41671-a028-4b71-8b90-f7f3e205c738

### DIFF
--- a/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -62,7 +62,9 @@ export const ConfirmationModal = ({
       }
       {...props}
     >
-      {children}
+      <div style={{ minWidth: '480px' }}>
+        {children}
+      </div>
     </Modal>
   );
 };


src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -62,7 +62,9 @@ export const ConfirmationModal = ({
       }
       {...props}
     >
-      {children}
+      <div style={{ minWidth: '480px' }}>
```
edlerd, line 65: This is problematic, if the screen size is lower than 480px.

nadeem-R17: Thank you for bringing that to my attention. I've addressed the issue by implementing an adjustable minWidth for the content, taking into consideration screen sizes lower than 480px. If you have any further suggestions or if additional changes are needed, please feel free to let me know.

https://github.com/canonical/react-components/assets/90577359/56f8c342-04b2-41c2-a424-3c100739f361

bartaz:
Thanks for your proposal @nadeem-R17, but we need to address it differently.

PR #1021, opened 2024-01-19
fix: remove invalid syntax for style imports

mas-who:
## Done

- fix issue with style imports with the `~` prefix not working with projects built with vite.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check that styling for the table pagination component still looks okay in storybook


bartaz (approved):
This seems to fix the issue. Styling still works in storybook.

It would be good if the CSS followed Vanilla naming convention better, but it's not scope of this PR, so lets leave it and fix later.

PR #1022, opened 2024-01-19
fix: remove react-router-dom dependency from NotificationProvider

mas-who:
## Done

- Removed useLocation call from NotificationProvider. This was causing issues with react-router-dom version being used in the project that imports this component.
- Removed react-router-dom as a dependency from the project. It was only being used in `NotificationProvider`.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- ensure CI tests still all pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "@testing-library/jest-dom": "5.17.0",
     "@testing-library/react": "14.1.2",
     "@testing-library/user-event": "14.5.2",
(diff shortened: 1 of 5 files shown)


edlerd (approved):
LGTM

PR #1034, opened 2024-01-31
feat: implement externally controlled mode for TablePagination [WD-8505]

mas-who:
## Done

- Introduced externally controlled mode for the `TablePagination` component
- added test for externally controlled mode for the `TablePagination` component. You can refer to the test for example usage in externally controlled mode.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check in projects that uses `TablePagination` if server pagination works.

### DIFF
--- a/src/components/TablePagination/TablePagination.tsx
+++ b/src/components/TablePagination/TablePagination.tsx
@@ -1,59 +1,19 @@
 import React, {
-  ChangeEvent,
-  Children,
   HTMLAttributes,
   PropsWithChildren,
-  ReactElement,
   ReactNode,
-  RefObject,
-  cloneElement,
-  useEffect,
-  useRef,
   useState,
 } from "react";
-import classnames from "classnames";
-import { usePagination } from "hooks";
-import Select from "components/Select";
-import TablePaginationControls from "./TablePaginationControls";
 import "./TablePagination.scss";
+import TablePaginationControls from "./TablePaginationControls";
+import {
+  DEFAULT_PAGE_LIMITS,
+  generatePagingOptions,
+  renderChildren,
+} from "./utils";
+import { usePagination } from "hooks";
 
-/**
- * Determine if we are working with a small screen.
- * 'small screen' in this case is relative to the width of the description div
- */
-const figureSmallScreen = (descriptionRef: RefObject<HTMLDivElement>) => {
-  const descriptionElement = descriptionRef.current;
-  if (!descriptionElement) {
-    return true;
-  }
-  return descriptionElement.getBoundingClientRect().width < 230;
-};
-
-/**
- * Iterate direct react child components and override the value of the prop specified by @param dataForwardProp
- * for those child components.
- * @param children - react node children to iterate
- * @param dataForwardProp - the name of the prop from the children components to override
- * @param data - actual data to be passed to the prop specified by @param dataForwardProp
- */
-const renderChildren = (
-  children: ReactNode,
-  dataForwardProp: string,
-  data: unknown[]
-) => {
-  return Children.map(children, (child) => {
-    return cloneElement(child as ReactElement, {
-      [dataForwardProp]: data,
-    });
-  });
-};
-
-const DEFAULT_PAGE_LIMITS = [50, 100, 200];
-const generatePagingOptions = (pageLimits: number[]) => {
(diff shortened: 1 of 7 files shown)


src/components/TablePagination/ClientPagination.tsx
edlerd, line 1: I have some doubts labelling it as client/server. In `Accordion`  we have an `externallyControlled` prop, which seems to fit here better. The duplication of the rendered components can be avoided by having a purely presentational (externally controlled) component, that we use in both cases. When externally controlled, we forward all props directly to it. When the state is handled locally, an additional component can do the state juggling and then forward to the same presentational component. Wdyt?

mas-who: Aaah nice! I did it this way so that I don't have to call `usePagination` in both scenarios. However, like you said, we could just use a wrapper component to handle state updates for locally controlled scenario. Thanks! will make the change.

src/components/TablePagination/TablePagination.tsx
```diff
@@ -85,93 +45,145 @@ export type Props = PropsWithChildren<{
...
+
+export type InternalControlProps = BasePaginationProps & {
+  /**
+   * Whether the component will be controlled via external state.
+   */
+  externallyControlled?: false;
+};
+
+export type Props = PropsWithChildren<
+  BasePaginationProps & (ExternalControlProps | InternalControlProps)
```
edlerd, line 92:
```suggestion
  ExternalControlProps | InternalControlProps
```
I think the `BasePaginationProps` are extended by both options.

src/components/TablePagination/TablePagination.tsx
```diff
@@ -85,93 +45,145 @@ export type Props = PropsWithChildren<{
...
+    paginate(1);
+    setInternalPageSize(pageSize);
   };
 
-  const totalPages = Math.ceil(data.length / pageSize);
-  const clonedChildren = renderChildren(children, dataForwardProp, pageData);
+  const clonedChildren = renderChildren(children, dataForwardProp, controlData);
   return (
     <>
       {position === "below" && clonedChildren}
```
edlerd, line 173: Would it be easier to gasp if we assign a `const controls = <TablePaginationcontrols ...` above the return block, and change this to something like below?

```
return (
  <>
    {position === "above" && controls}
    {clonedChildren}
    {position === "below" && controls}
```

Comment

edlerd (commented):
Thanks for applying the new structure, I think this is way better than before!

Two suggestions for further improvements below.

src/components/TablePagination/TablePagination.tsx
mas-who: good catch, I must be in a rabbit hole at that point :sweat:

src/components/TablePagination/TablePagination.tsx
mas-who: I agree, this is better and no longer inverted!

src/components/TablePagination/TablePagination.tsx
```diff
@@ -85,94 +45,150 @@ export type Props = PropsWithChildren<{
...
+        pageSizeFound = true;
+        break;
+      }
+    }
+
+    if (!pageSizeFound) {
+      throw new Error(
+        `pageSize must be a valid option in pageLimits, pageLimits is set to [${pageLimits}]`
+      );
+    }
```
edlerd, line 125: Shall we check for additional props as well? As in the `externallyControlled` mode, `currentPage`, `onPageChanged`, and other props are mandatory. Maybe we should throw if they are missing?

edlerd (approved):
LGTM thanks for making this nice :)
Just one more idea to a bit more defensive implementation, feel free to address this or not.

src/components/TablePagination/TablePagination.tsx
mas-who: the discriminating union typing for the component will result in Typescript build errors if `externallyControlled` is set to `true` but the required props are not passed. I think defensive code would be redundant in this case.

mas-who: i.e. `ExternalControlProps | InternalControlProps` in the props type definition.

Issue #1041, opened 2024-02-09
Slider input field has wrong width


lorumic:
This seems to be a problem introduced in the context of #1036 (version 0.49.0).

The slider input used to have a 5% width (verified checking out the commit tagged `v0.48.0`), and hence looked like this:

![Screenshot from 2024-02-09 15:41:16](https://github.com/canonical/react-components/assets/56583786/c650927f-54e5-4081-9297-57b41fe1afc5)

From version 0.49.0, it looks like this instead:

![image](https://github.com/canonical/react-components/assets/56583786/9cfd7550-5f5b-46c4-b5a1-d533e35f8cea)

@petermakowski Do you know why this is happening?

petermakowski:
Very strange as there's no CSS code that could affect this in this change. I'll investigate.

petermakowski:
@bartaz  Some background info I managed to figure out so far:
`%vf-input-elements` cannot be used without `@include vf-b-forms`, which includes global styles for input elements. 
https://github.com/canonical/vanilla-framework/blob/b4378c920f8bcb804db2ead64566ce3fa54636ae/scss/_base_forms.scss#L144

As soon as you add `@include vf-b-forms` to a component in order to extend `%vf-input-elements`, that as a result affects the order final CSS declarations and specificity which may alter styles in unrelated component.

## Screenshots

##### styles from base forms end up having higher specificity than those of a slider
![Google Chrome screenshot 001541@2x](https://github.com/canonical/react-components/assets/7452681/878e5e09-eeb1-4504-bfac-adabbaae47ad)

##### after removal of `@include vf-b-forms` import in `MultiSelect.scss` - styles declarations have expected order https://github.com/canonical/react-components/pull/1042
![Google Chrome screenshot 001543@2x](https://github.com/canonical/react-components/assets/7452681/51788a5b-5650-43e4-9333-e4bce6a0c1a0)

We could increase specificity of `.p-slider__input` alternatively, but the real fix would be to give an ability to _just_ import base input styles (`%vf-input-elements`) without global styles.

bartaz:
> We could increase specificity of `.p-slider__input` alternatively, but the real fix would be to give an ability to _just_ import base input styles (`%vf-input-elements`) without global styles.

@petermakowski Thanks for looking into this and explaining the context. I think the issue is that we don't include Vanilla base styles in React components and try to add styles on top of that. That will not work. 

[...]

PR #1042, opened 2024-02-09
fix: slider input field has wrong width #1041

petermakowski:
## Done

- fix: Slider input field has wrong width #1041

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Verify that bug in the linked issue has been fixed.

## Fixes

Fixes: # .

### DIFF
--- a/src/components/MultiSelect/MultiSelect.scss
+++ b/src/components/MultiSelect/MultiSelect.scss
@@ -1,5 +1,6 @@
 @use "sass:map";
 
+// stylelint-disable-next-line scss/load-no-partial-leading-underscore
 @import "~vanilla-framework/scss/_base_forms";
 @import "~vanilla-framework/scss/base_placeholders";
 @import "~vanilla-framework/scss/base_icon-definitions";
--- a/src/components/MultiSelect/MultiSelect.scss
+++ b/src/components/MultiSelect/MultiSelect.scss
@@ -1,9 +1,79 @@
 @use "sass:map";
-@import "vanilla-framework";
-@include vf-base;
-@include vf-p-lists;
+
+@import "~vanilla-framework/scss/_base_forms";
+@import "~vanilla-framework/scss/base_placeholders";
+@import "~vanilla-framework/scss/base_icon-definitions";
+
+@include vf-b-placeholders;
 
 $dropdown-max-height: 20rem;
+
+/**
+ Including base vanilla input elements declaration from:
+https://github.com/canonical/vanilla-framework/blob/b4378c920f8bcb804db2ead64566ce3fa54636ae/scss/_base_forms.scss#L14
+
+Workaround for: https: //github.com/canonical/react-components/issues/1041
+*/
+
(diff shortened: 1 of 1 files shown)


src/components/MultiSelect/MultiSelect.scss
```diff
@@ -1,10 +1,80 @@
 @use "sass:map";
-@import "vanilla-framework";
-@include vf-base;
-@include vf-p-lists;
+
+@import "~vanilla-framework/scss/_base_forms";
```
lorumic, line 3: Stylelint seems to not like the leading `_` in `_base_forms`. :slightly_frowning_face: 
In my opinion, we can safely ignore this. We are constrained by how this is named in Vanilla, so there's not much we can do about it.
```suggestion
// stylelint-disable-next-line scss/load-no-partial-leading-underscore
@import "~vanilla-framework/scss/_base_forms";
```

src/components/MultiSelect/MultiSelect.scss
```diff
@@ -1,10 +1,80 @@
...
+
+  &:active,
+  &:focus {
+    background-color: $colors--light-theme--background-active;
+  }
+
+  &[disabled],
+  &[disabled="disabled"] {
+    @extend %vf-disabled-element;
+  }
```
lorumic, line 75: Don't we need the same for the `readonly` attribute as well?
```suggestion
  &[readonly],
  &[readonly="readonly"] {
    @extend %vf-readonly-element;
  }

  &[disabled],
  &[disabled="disabled"] {
    @extend %vf-disabled-element;
  }
```

src/components/MultiSelect/MultiSelect.scss
```diff
@@ -1,10 +1,80 @@
...
 $dropdown-max-height: 20rem;
 
+/**
+ Including base vanilla input elements declaration from:
+https://github.com/canonical/vanilla-framework/blob/b4378c920f8bcb804db2ead64566ce3fa54636ae/scss/_base_forms.scss#L14
+
+Workaround for: https: //github.com/canonical/react-components/issues/1041
+*/
+
+%vf-readonly-element {
```
bartaz, line 18: Why do we need to copy so much from Vanilla here?
This is extremely risky and fragile, as these internals of Vanilla are likely to be changed and this copy will not reflect that.

And I'm not sure if I understand how these styles are related to completely separate Slider component being broken?

petermakowski: I'm fully aware it's very fragile, hence I clearly marked it as a workaround in the comment. This was meant to be a quick fix until we figure out a proper way of addressing this. I'll add more information to the linked issue.

petermakowski:
@bartaz Added a more detailed explanation in https://github.com/canonical/react-components/issues/1041#issuecomment-1938288610

PR #1044, opened 2024-02-14
fix: address a11y issues in input related components

mas-who:
## Done

- do not set `aria-describedby` attribute for the Input component if there is no help or success validation text.
- do not render label element for `CheckableInput` if input label is empty. This resolves the issue of potentially creating labels with no text content.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Adjust the `help` and `success` props for the Input component on Storybook and make sure that the `aria-describedby` attribute is not set if both props are undefined.
- Pass empty string for the `label` prop for `RadioInput` and `CheckboxInput` components and make sure there is no empty `label` element rendered.

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -100,10 +100,11 @@ const Input = ({
   const helpId = useId();
   const hasError = !!error;
 
+  const description = [help ? helpId : null, success ? validationId : null]
+    .filter(Boolean)
+    .join(" ");
   const commonProps = {
-    "aria-describedby": [help ? helpId : null, success ? validationId : null]
-      .filter(Boolean)
-      .join(" "),
+    "aria-describedby": !description ? undefined : description,
     "aria-errormessage": hasError ? validationId : null,
     "aria-invalid": hasError,
     id: inputId,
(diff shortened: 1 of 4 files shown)


src/components/CheckboxInput/CheckableInput/CheckableInput.tsx
```diff
@@ -14,7 +14,7 @@ export type Props = PropsWithSpread<
     /**
      * The label for the input element.
      */
-    label: ReactNode;
+    label?: ReactNode;
```
edlerd, line 17: I think the label is required and shouldn't be optional.

edlerd (changes requested):
Thanks for removing the empty `aria-describedby` attribute.
One concern below on making the label optional.

src/components/CheckboxInput/CheckableInput/CheckableInput.tsx
mas-who: yeah true, since we are already doing a falsey check for label it can remain as a required prop

edlerd (approved):
LGTM, thank you for the fixes :)

PR #1050, opened 2024-03-04
refactor: remove nanoid

huwshimi:
## Done

- Deprecate custom useId hook and use useId provided by React 18.


src/hooks/useId.ts
```diff
@@ -1,9 +1,16 @@
...
+import { IS_DEV } from "utils";
 
 /**
- * A hook that returns the same random ID string on every render.
- * Can be used as a value for HTML "id" attributes.
- * @returns random ID string
+ * @deprecated Code component is deprecated. Use CodeSnippet component or inline `<code>` instead.
  */
-export const useId = (): string => useRef(nanoid()).current;
+export const useId = () => {
```
vladimir-cucu, line 8: After a quick look at the imports, I'm not sure if this is used anywhere right now. Is there a reason why we still need this code?

vladimir-cucu (approved):
Looks good to me! :rocket:

src/hooks/useId.ts
huwshimi: This is just for any external projects that use this hook (I know some projects like maas-ui and juju-dashboard used to import it, though I'm not sure if they still do).

PR #1052, opened 2024-03-14
fix(MultiSelect): Apply a temporary fix for Vanilla includes in MultiSelect component

bartaz:
## Done

- Updates Vanilla to 4.9.0
- Removes problematic `@includes` (`vf-base` and `vf-b-forms`) and`@extend` uses from `MultiSelect.scss` and replaces them with copies of relevant parts of Vanilla.

This is a temporary solution until we find better ways to reuse bigger parts of Vanilla styling. `@extend` usage by itself is not huge issue, but it does silently create a duplicate of Vanilla styling.
Bigger problem was including `vf-base` and `vf-lists` (which contain required placeholders to extend) as they duplicate big chunks of Vanilla styling creating source order conflicts of old and new styles and breaking the styling of some other components.

### QA steps

- Check if MultiSelect renders and works as expected:
  - https://react-components-1052.demos.haus/?path=/docs/multiselect--docs
- Check if previously broken slider component renders as expected:
  - https://react-components-1052.demos.haus/?path=/story/slider--with-input

## Fixes

Fixes: #1041

### DIFF
--- a/src/components/Icon/Icon.stories.scss
+++ b/src/components/Icon/Icon.stories.scss
@@ -1,7 +1,4 @@
-@import "~vanilla-framework/scss/settings";
-@import "~vanilla-framework/scss/base_placeholders";
-@import "~vanilla-framework/scss/base_icon-definitions";
-@import "~vanilla-framework/scss/patterns_icons";
+@import "vanilla-framework";
 
 @include vf-b-placeholders;
 @include vf-p-icons-common;
--- a/src/components/MultiSelect/MultiSelect.scss
+++ b/src/components/MultiSelect/MultiSelect.scss
@@ -1,138 +1,15 @@
 @use "sass:map";
 @import "vanilla-framework";
+@include vf-b-placeholders; // Vanilla base placeholders to extend from
 
 $dropdown-max-height: 20rem;
 
-// XXX: Temporarly we are duplicating some of the placeholders defined in Vanilla
-// to avoid conflict of including the duplicate of base Vanilla in React components.
-// This is a temporary solution until we have a better way to handle this.
-
-// as in Vanilla _base_placeholders.scss
-%vf-hide-text {
-  overflow: hidden;
-  text-indent: 110vw;
-  white-space: nowrap;
-}
-
-// as in Vanilla _base_placeholders.scss
-%icon {
-  $vertical-offset: 0.5px;
-  @extend %vf-hide-text;
-  @include vf-icon-size($default-icon-size);
-  background: {
-    position: center;
-    repeat: no-repeat;
-  }
-
-  display: inline-block;
-  font-size: inherit; // allow icons to match size of parent text element, when set in em
-  margin: 0;
-  padding: 0;
-  position: relative;
-  vertical-align: calc(
-    $vertical-offset + 0.5 * $cap-height - 0.5 * $default-icon-size
-  );
-}
-
-// as in Vanilla _base_forms.scss
-%vf-disabled-element {
-  cursor: not-allowed;
-  opacity: $disabled-element-opacity;
-}
-
-// as in Vanilla _base_forms.scss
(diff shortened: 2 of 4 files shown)


bartaz:
Percy: https://percy.io/bb49709b/react-components/builds/33166994/

There are a lot of unrelated changes caused by Vanilla upgrade that includes new theming and new icon colours, etc.

There is also some weird behaviour in Percy that it shows new input border as lighter than before, which is weird because when checked in the browser it doesn't look lighter than before.

image

I'm ignoring it for now, assuming it's some Percy/Storybook glitch, but would be good to undestand the cause and if we can do anything about it, to avoid false negatives in future.

src/components/Icon/Icon.stories.scss
```diff
@@ -1,7 +1,4 @@
-@import "~vanilla-framework/scss/settings";
```
bartaz, line 1: Removing unnecessary individual imports.
It's all included in full Vanilla import and there is no downside to import it all (no CSS produced until anything is included).

src/components/MultiSelect/MultiSelect.scss
```diff
@@ -1,10 +1,137 @@
 @use "sass:map";
 @import "vanilla-framework";
-@include vf-base;
-@include vf-p-lists;
 
 $dropdown-max-height: 20rem;
 
+// XXX: Temporarly we are duplicating some of the placeholders defined in Vanilla
```
bartaz, line 6: In longer term we should either upstream this component to Vanilla or refactor some bits of Vanilla to be able to extend these parts that are needed here, without the need to include the whole base or list styles, because duplication of base form styles was the cause of the issue (broken slider styles overriden by duplicated input styles).

src/components/MultiSelect/MultiSelect.scss
```diff
@@ -97,12 +225,6 @@ $dropdown-max-height: 20rem;
   }
 }
 
-.multi-select__dropdown-item-description {
```
bartaz, line 100: @petermakowski This didn't seemed to be used anywhere, so I removed it. Is this needed?

src/components/MultiSelect/MultiSelect.scss
petermakowski: I'd suggest we link to a new issue in react-components where we explain this in detail (and move this comment there).

src/components/MultiSelect/MultiSelect.scss
petermakowski: Good spot. It's no longer used, safe to remove.

petermakowski (approved):

src/components/MultiSelect/MultiSelect.scss
bartaz: It felt more fitting on Vanilla itself: https://github.com/canonical/vanilla-framework/issues/5030

src/components/MultiSelect/MultiSelect.scss
```diff
@@ -1,11 +1,15 @@
...
-@include vf-p-lists;
+@include vf-b-placeholders; // Vanilla base placeholders to extend from
 
 $dropdown-max-height: 20rem;
 
 .multi-select {
+  // Scope Vanilla form includes to multi select component only
+  // to avoid overriding any Vanilla base styles
+  // XXX: This is a workaround for https://github.com/canonical/vanilla-framework/issues/5030
+  @include vf-b-forms;
```
bartaz, line 11: As for now `vf-b-forms` contains both placeholders that we want to extend here and default form styling (that we don't want to duplicate). So to avoid overriding default Vanilla we need to scope it within the component.

bartaz: @petermakowski I think I found a cleaner workaround for extending Vanilla styles in components.

Issue #957, opened 2023-08-10
Inputs have no id and are passed extra params


huwshimi:
The Input component use to create a default id if one wasn't passed as a prop so that the label `forId` was set to the input.

Also, the value passed to the `label` prop is passed down to the `<input>` so if you've passed JSX you get something like `<input label="[object Object]" />`. I don't think `label` is a valid attribute of an input so this is probably just a mistake.

huwshimi:
It looks like this was fixed by https://github.com/canonical/react-components/pull/1012.

huwshimi:
Oops, only the id part was resolved by the previous PR.

PR #1054, opened 2024-03-19
feat: upstream FormikField

huwshimi:
## Done

- upstream the FormikField component to make react-component's fields and Formik easier to work with. This component is already in use in maas-ui and react-components and we want to use it in another project.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the FormikField docs page in Storybook.
- Check that the docs make sense.

## Fixes

https://warthogs.atlassian.net/browse/WD-9813
Fixes: #957.

### DIFF
--- a/src/components/FormikField/FormikField.stories.tsx
+++ b/src/components/FormikField/FormikField.stories.tsx
@@ -1,0 +1,75 @@
+import React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
+
+import FormikField from "./FormikField";
+import Select from "../Select";
+import { Formik } from "formik";
+
+const meta: Meta<typeof FormikField> = {
+  title: "FormikField",
+  component: FormikField,
+  tags: ["autodocs"],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof FormikField>;
+
+export const Default: Story = {
+  render: () => (
+    <Formik initialValues={{ username: "" }} onSubmit={() => {}}>
+      <FormikField name="username" label="Username" type="text" />
+    </Formik>
+  ),
+};
+
+export const Fields: Story = {
+  parameters: {
+    docs: {
+      description: {
+        story: `
+Any React Components input can be provided to FormikField (e.g. Input, Textarea or Select) or you may provide a custom component.
+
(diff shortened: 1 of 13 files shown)


src/components/FormikField/FormikField.stories.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react";
```
vladimir-cucu, line 1: I think the docs doesn't render as expected. For instance:
- when pressing `Show code`, the entire code is presented, but I assume we only want the bit from `render` to be displayed.
- the default value of component prop in props table is not listed in DEFAULT column.

The first issue also appears in `Card` stories, which is one of the few other `.stories.tsx` docs. There are also some slight inconsistencies in presentation between `.stories.tsx` and `.stories.mdx` docs e.g. the Props header missing in `.stories.tsx` docs.

huwshimi: > I think the docs doesn't render as expected. For instance:
> 
> * when pressing `Show code`, the entire code is presented, but I assume we only want the bit from `render` to be displayed.

There's an open issue here, with a workaround that I've applied to both files: https://github.com/storybookjs/storybook/issues/22281.

> * the default value of component prop in props table is not listed in DEFAULT column.

It looks like the table couldn't handle the `Input` component as a default but this can be set via a jsdoc string so I've done that.

> 
> The first issue also appears in `Card` stories, which is one of the few other `.stories.tsx` docs. There are also some slight inconsistencies in presentation between `.stories.tsx` and `.stories.mdx` docs e.g. the Props header missing in `.stories.tsx` docs.

It looks like it was decided not to customise the page too much and stick with the CSF defaults: https://github.com/canonical/react-components/pull/1010#discussion_r1447061301.

vladimir-cucu (approved):
LGTM :+1:

src/components/FormikField/FormikField.stories.tsx
vladimir-cucu: Thanks for the explanation and links, and for also fixing the inconsistency in `Card` stories!

Issue #1032, opened 2024-01-30
What to do with propagation of click events in ConfirmationModal?


vladimir-cucu:
Currently, when we press the `Cancel`, `Confirm` or the close button in `ConfirmationModal` component, the click propagates further to underlying components. In Juju Dashboard we have a panel that triggers the opening of a `ConfirmationModal` that doesn't overlap the panel (see attached screenshot). When we press any of the aforementioned 3 buttons, a click outside the panel is triggered as well, which closes the panel. However, we want the click outside the panel to not be triggered. We could add `event.nativeEvent.stopImmediatePropagation()` in juju dashboard where necessary, but I was wondering if we could instead include this in the `Modal` component itself? We had a similar [issue](https://github.com/canonical/react-components/pull/974) with propagating the click of `Esc` button in `Modal` component and we fixed it by stopping the propagation of the event.

![Screenshot from 2024-01-30 13-27-43](https://github.com/canonical/react-components/assets/108150922/36425577-a433-465d-a1f4-d0ded7130c21)

vladimir-cucu:
Discussing these questions at the React Guild sounds like a good idea. I'll remember to add it to the notes before the meeting next Thursday.

bartaz:
As discussed in React Guild - we are fine with stopping propagation inside the modal by default.

If we ever have a case where we need the propagation, we can add it as an option.

vladimir-cucu:
I could take a look at fixing this issue if nobody began looking at it. Should we disable the event propagation by default and add an optional prop `stopEventPropagation = true` to the `Modal` component?

PR #1058, opened 2024-03-24
fix: stop click event propagation in Modal

vladimir-cucu:
## Done

- Stopped click event propagation when closing `Modal` and when closing or confirming `ConfirmationModal`.

## QA

- Check that `Modal` and `ConfirmationModal` renders as expected in storybook.
- (optional) For example, the stop click event propagation can be tested in Juju Dashboard when opening the charm actions panel of a charm and clicking on an action. A `ConfirmationModal` will be rendered in a `Portal`. Clicking the close button used to close the panel, but, using the changes in this PR, it doesn't.

## Fixes

- Fixes: #1032 
- https://warthogs.atlassian.net/browse/WD-9915

### DIFF
--- a/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -29,7 +29,7 @@ export type Props = PropsWithSpread<
     /**
      * Function to perform the action prompted by the modal.
      */
-    onConfirm: (e: MouseEvent<HTMLElement>) => void;
+    onConfirm: (event: MouseEvent<HTMLElement>) => void;
   },
   Omit<ModalProps, "buttonRow">
 >;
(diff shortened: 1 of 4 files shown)


src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -43,18 +43,32 @@ export const ConfirmationModal = ({
   onConfirm,
   ...props
 }: Props): ReactElement => {
+  const handleClick =
+    (action: Function | null | undefined) =>
```
huwshimi, line 47: I think this could be slightly improved by using a generic so the type of the function can be inferred.
```suggestion
  const handleClick =
    <A extends Function>(action: A) =>
```

src/components/Modal/Modal.tsx
```diff
@@ -121,18 +128,27 @@ export const Modal = ({
     shouldClose.current = false;
   };
 
-  const handleOverlayOnMouseDown = (event) => {
-    if (event.target === modalRef.current) {
+  const handleOverlayOnMouseDown = (e: React.MouseEvent<HTMLDivElement>) => {
+    if (e.target === modalRef.current) {
```
huwshimi, line 132: We generally try to avoid one character variable names.

huwshimi (commented):
I've QAed this with Juju Dashboard and it works!

src/components/Modal/Modal.tsx
vladimir-cucu: Changed every `e` to `event` in `Modal` and `ConfirmationModal`. Thanks for the review!

huwshimi (approved):
Looks good!

PR #1062, opened 2024-04-03
fix: tooltip open

edlerd:
## Done

- Fix tooltip open/close: When tooltip wraps multiple children, and we change which child is hovered, the tooltip should stay open.

see description video in https://github.com/canonical/anbox-cloud-dashboard/pull/732#pullrequestreview-1976424478


lorumic:
QAed locally using storybook.

- Before:

Peek 2024-04-03 22-54

- After:

Peek 2024-04-03 22-55

lorumic (approved):

Issue #1068, opened 2024-04-12
Textarea doesn't update value when it's an empty string


andesol:
Consider this simplified example:

```js
import { useState } from 'react';
import { Button, Textarea } from "@canonical/react-components";

function App() {
  const [data, setData] = useState("Default text");

  return (
      <div>
        <Button onClick={() => setData("")}>
          Show nothing
        </Button>
        <p>Current data is: {data}</p>
        <Textarea
          value={data}
          onChange={(e) => setData(e.target.value)}
        />
      </div>
    );
}
```

This is the default scenario:

![image](https://github.com/canonical/react-components/assets/20623084/9dd66d91-a8b7-499f-9874-240edf0fcde4)

I'd expect clicking the button to clear the value in the `Textarea`. However, while the state is correctly updated (to an empty string), the `Textarea` still shows the previous value.

![image](https://github.com/canonical/react-components/assets/20623084/9e7bd8b6-f9a9-4fa6-ab45-e74bc32a38ac)

andesol:
It looks like [this line](https://github.com/canonical/react-components/blob/main/src/components/Textarea/Textarea.tsx#L162) should be
```
props.value ?? innerValue
```

bartaz:
Good catch @andesol. Care to propose a PR? I guess it would be good to have a test case that catches this as well.

andesol:
@bartaz Indeed, here's the PR https://github.com/canonical/react-components/pull/1070

PR #1070, opened 2024-04-16
fix(textarea): reset value to an empty string

andesol:
## Done

- Allow setting the value to an empty string. Otherwise, the previous value isn't updated.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

You should be able to replicate this in a simplified example such as the one in the issue (#1068).

### Percy steps

No visual change

## Fixes

Fixes: #1068

### DIFF
--- a/src/components/Textarea/Textarea.test.tsx
+++ b/src/components/Textarea/Textarea.test.tsx
@@ -40,4 +40,15 @@ describe("Textarea ", () => {
     render(<Textarea help={help} />);
     expect(screen.getByRole("textbox")).toHaveAccessibleDescription(help);
   });
+
+  it("can change value to an empty string", () => {
+    let testValue = "test";
+    const { rerender } = render(<Textarea value={testValue} />);
+    const textarea = screen.getByRole("textbox");
+    expect(textarea).toHaveValue("test");
+
+    testValue = ""; // simulate an external change of the value
+    rerender(<Textarea value={testValue} />);
+    expect(textarea).toHaveValue("");
+  });
 });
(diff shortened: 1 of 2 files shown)


src/components/Textarea/Textarea.test.tsx
```diff
@@ -40,4 +40,13 @@ describe("Textarea ", () => {
...
     expect(screen.getByRole("textbox")).toHaveAccessibleDescription(help);
   });
+
+  it("can change value to an empty string", () => {
+    const { rerender } = render(<Textarea value="test" />);
+    const textarea = screen.getByRole("textbox");
+    expect(textarea).toHaveValue("test");
+
+    rerender(<Textarea value="" />);
+    expect(textarea).toHaveValue("");
```
lorumic, line 50: Should we make this even closer to the reality of a React state change by doing a variable change? Something like this:
```suggestion
    let testValue = "test";
    const { rerender } = render(<Textarea value={testValue} />);
    const textarea = screen.getByRole("textbox");
    expect(textarea).toHaveValue("test");

    testValue = ""; // simulate an external change of the value
    rerender(<Textarea value={testValue} />);
    expect(textarea).toHaveValue("");
```

andesol: Makes sense, thanks for checking!

andesol:
@lorumic Ready!

lorumic (approved):
LGTM :+1:

PR #1075, opened 2024-04-22
feat: allow loading and disable state for confirm button in ConfirmationModal

mas-who:
## Done

- Allow loading state for the confirm button in the modal
- Allow disabled state for the confirm button in the modal

### DIFF
--- a/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -3,6 +3,7 @@ import type { ReactNode } from "react";
 import { PropsWithSpread, ValueOf } from "types";
 import Button, { ButtonAppearance } from "components/Button";
 import Modal, { ModalProps } from "components/Modal";
+import ActionButton from "components/ActionButton";
 
 export type Props = PropsWithSpread<
   {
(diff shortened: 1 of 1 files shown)


src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -30,6 +31,14 @@ export type Props = PropsWithSpread<
...
      */
     onConfirm: (event: MouseEvent<HTMLElement>) => void;
+    /**
+     * Is the confirm button loading.
+     */
+    confirmButtonLoading?: boolean;
+    /**
+     * Disable the confirm button.
+     */
+    confirmButtonDisabled?: boolean;
```
edlerd, line 41: Maybe both can be `loading` and `disabled` to stay in line with how the action button component calls them? Or do you think those will be too ambiguous?

edlerd (approved):
LGTM, thanks for adding this. One suggestion to make the naming more consistent with existing components.

src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -30,6 +31,14 @@ export type Props = PropsWithSpread<
      * Function to perform the action prompted by the modal.
      */
     onConfirm: (event: MouseEvent<HTMLElement>) => void;
+    /**
+     * Is the confirm button loading.
+     */
+    confirmButtonLoading?: boolean;
+    /**
+     * Disable the confirm button.
```
lorumic, line 39: I would use the same descriptions that we have for the `ActionButton` here:
```suggestion
     * Whether the confirm button should be in the loading state.
     */
    confirmButtonLoading?: boolean;
    /**
     * Whether the confirm button should be disabled.
```

PR #1081, opened 2024-05-10
chore(deps): update to Storybook 8

huwshimi:
## Done

- Upgrade Storybook deps to v8.
- Run migrations.
- Convert docs to CSF v3.
- Update docs to use autodoc.
- Fix iframe in prod. Fixes: #1102.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go through and check that component docs look OK.


package.json
```diff
@@ -107,7 +107,7 @@
     "@types/react": "18.3.1",
     "@types/react-dom": "18.3.0",
     "postcss": "^8.3.11",
-    "jackspeak": "2.3.6"
+    "jackspeak": "2.1.1"
```
edlerd, line 110: Why reduce the version?

src/docs/getting-started.stories.mdx
edlerd, line 1: Shall we keep this in some form?

edlerd (approved):
Spot checking QA on the demo server seems fine.
I didn't manage to check all the individual components. I guess those are mostly 1:1 transitions from mdx to tsx. Thanks for this massive update!

package.json
huwshimi: The newer version doesn't work with Yarn 1.

src/docs/getting-started.stories.mdx
huwshimi: Oops thanks, this file was renamed but the new file was not committed due to the .gitignore rules. I've fixed this now.

edlerd (changes requested):
I am surprised all screenshots in percy are marked as "new"?
Also this one seems to be not all right:
image

huwshimi:
> I am surprised all screenshots in percy are marked as "new"?

I think this is because the Storybook URL scheme has changed.

> Also this one seems to be not all right: 

Every time I run it a different set of images fail with that error. I'll try again later when hopefully the server is more stable.

huwshimi:
I tried upgrading to the latest percy and storybook, but that didn't fix it.

huwshimi:
@edlerd I've increased the asset loading timeouts in CircleCI and now the snapshots are all working.

.percy.yml
```diff
@@ -2,3 +2,8 @@ version: 2
 snapshot:
   widths: [1280]
   enable-javascript: true
+discovery:
+  network-idle-timeout: 700
+  retry: true
+  launch-options:
+    timeout: 60000
```
edlerd, line 9:
```suggestion
    timeout: 60000

```

edlerd (approved):
LGTM, tinny nitpick below.
It might be better to stash the commits into a single one?

.percy.yml
huwshimi: Thanks! Fixed.

PR #1082, opened 2024-05-10
feat: allow focusing disabled buttons and displaying tooltips

petermakowski:
## Done
- feat: allow focusing disabled buttons and displaying tooltips
  - this allows to display associated tooltips via keyboard
  
### Rationale
Previously, `disabled` attribute was used which prevents focusing the button entirely. This meant that users navigating by keyboard could not focus the button and any associated tooltip would not be displayed.

By switching to aria-disabled, the button can still be focused while conveying its disabled state to assistive technologies. Submit action is prevented by using `.preventDefault`.

https://css-tricks.com/making-disabled-buttons-more-inclusive/
  
### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- View the "Disabled with tooltip" button example in Storybook
- Tab to the disabled button and verify the tooltip is displayed
- Hover over the disabled button to see the tooltip appear
- Click the disabled button and confirm the click is prevented

### Percy steps

- New "Disabled with tooltip" button example added

## Fixes

Fixes: https://warthogs.atlassian.net/browse/MAASENG-2122

### After
Google Chrome screenshot 001879@2x

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -88,21 +88,19 @@ const Button = <P,>({
     {
       "has-icon": hasIcon,
       "is-dense": dense,
-      "is-disabled": Component !== "button" && disabled,
+      "is-disabled": disabled,
       "is-inline": inline,
       "is-small": small,
     },
     className
   );
   const onClickDisabled = (e: MouseEvent) => e.preventDefault();
-  const disabledProp =
-    Component === "button" ? { disabled } : { "aria-disabled": disabled };
 
   return (
     <Component
       className={classes}
       onClick={disabled ? onClickDisabled : onClick}
-      {...disabledProp}
+      aria-disabled={disabled || undefined}
       {...buttonProps}
     >
       {children}
--- a/src/components/Button/Button.stories.mdx
+++ b/src/components/Button/Button.stories.mdx
@@ -1,6 +1,7 @@
 import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
 
 import Button, { ButtonAppearance } from "./Button";
+import Tooltip from "../Tooltip";
 
 <Meta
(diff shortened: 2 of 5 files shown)


src/components/Button/Button.tsx
```diff
@@ -88,21 +88,19 @@ const Button = <P,>({
...
   const onClickDisabled = (e: MouseEvent) => e.preventDefault();
-  const disabledProp =
-    Component === "button" ? { disabled } : { "aria-disabled": disabled };
 
   return (
     <Component
       className={classes}
       onClick={disabled ? onClickDisabled : onClick}
-      {...disabledProp}
+      aria-disabled={disabled ? disabled : undefined}
```
ndv99, line 103: Can this not just be `disabled` passed directly in? It's already a boolean/undefined value

bartaz:
We had some discussions around that in the past, with no exact conclusion from UX part what's the best approach: https://github.com/canonical/vanilla-framework/issues/4299 https://github.com/canonical/vanilla-framework/issues/4942

But seems like having tooltips on disabled buttons is something that we want, so this may be a way for implementing this.

Is there anything here that would be worth upstreaming to Vanilla? I guess there are no CSS changes needed, but I guess we could have a similar example and functionality documented there.

src/components/Button/Button.tsx
petermakowski: This is to explicitly skip this prop when false. This is similar to existing behaviour where `disabled={false}` results in `disabled` attribute not being rendered at all. I did this to reduce the impact of this change (such as failing snapshot tests).

ndv99 (approved):
LGTM :+1:

PR #1089, opened 2024-05-17
feat: add application layout side navigation components

huwshimi:
## Done

- Add SideNavigation and child components. These components are used by the application layout but can also be a used in other Vanilla layouts.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open the demo.
- Scroll down to the the SideNavigation docs and check they look OK.


src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,183 @@
...
+  index: number,
+  linkComponent?: Props<L>["linkComponent"],
+  dark?: boolean
+) => {
+  if (isReactNode(item)) {
+    return <SideNavigationItem key={index}>{item}</SideNavigationItem>;
+  }
+  if ("nonInteractive" in item) {
+    return (
+      <SideNavigationItem {...item} dark={item.dark ?? dark} key={index} />
```
edlerd, line 82: Can we rely on the parent for the `dark` property? I wonder if the item itself needs a notion of being dark or not -- maybe it can fully rely on its parent for this?
I can't think of a use case where some items should be dark and others or the parent not.

src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,183 @@
...
+  listClassName?: string,
+  linkComponent?: SideNavigationLinkProps["component"],
+  dark?: boolean
+) => {
+  return groups
+    ?.filter(Boolean)
+    .map((group: NavItemGroup<L> | NavGroup<L>, g) => {
+      let items: NavItem<L>[];
+      let props: HTMLProps<HTMLUListElement> = {};
+      if (typeof group === "object" && "items" in group) {
```
edlerd, line 112: This is a bit confusing to me. Why can the group come in so many different shapes? Should we have a more strict approach here, so the group is of a specific well-defined type, instead of giving options to be out of various shapes?

edlerd (approved):
LGTM, two comments below, none of them are blocking.

One other aspect to think about: Will this be possible to extend for collapsible sections? See the video below for how it looks like. No need to implement this right away, just making sure with the introduced structure we don't make it too hard to extend as a next version of this.

[Screencast from 22.05.2024 10:55:05.webm](https://github.com/canonical/react-components/assets/1155472/761bdfaa-7430-4dd4-b8ba-faa3f54cf101)

src/components/SideNavigation/SideNavigation.tsx
huwshimi: Yeah I'm not sure when this might be required, but I figured it can't hurt to override the property if someone specifically sets the property some items.

src/components/SideNavigation/SideNavigation.tsx
huwshimi: Yeah that's probably sensible. This type grew as I updated the side nav to support the different projects, but we could just use the object type which would support all cases e.g.
```typescript
    <SideNavigation
      items={[
        {
          className: "extra-ul-class",
          items: [
            {
              label: "Link one",
              href: "#",
            },
          ],
        },
        {
          items: [
            {
              label: "Link two",
              href: "#",
              icon: "user",
            },
          ],
        },
      ]}
    />
```

In this case, if you only had one ul you'd still have to provide the items in an object `items={[{ items: [...] }]}` but that's probably a reasonable trade-off to make the component generally easier to work with.

Does that sound better?

huwshimi:
> One other aspect to think about: Will this be possible to extend for collapsible sections? See the video below for how it looks like. No need to implement this right away, just making sure with the introduced structure we don't make it too hard to extend as a next version of this.

I suspect this should probably be a modification in Vanilla that we would then add to react-components.

Right now this is supported in the side nav as you can pass any `ReactNode` as an item, so you pass in a custom accordion component. For example this is an excerpt from how I handle the side nav accordion in lxd-ui here: https://github.com/canonical/lxd-ui/pull/768/files#diff-0a53ec42b8db78013ad213a9f952a9b25ef65c0ce9ee169bd1b65af00f245607R185

```typescript
      <SideNavigation<NavLinkProps>
        dark
        items={[
          {
            className: "sidenav-top-ul",
            items: isAuthenticated
              ? [
                  ...
                  {
                    icon: "connected",
                    label: "Networks",
                    onClick: softToggleMenu,
                    title: `Networks (${projectName})`,
                    to: `/ui/project/${projectName}/networks`,
                  },
                  <NavAccordion
                    baseUrl={`/ui/project/${projectName}/storage`}
                    title={`Storage (${projectName})`}
                    iconName="pods"
                    key="storage"
                    label="Storage"
                    onOpen={() => toggleAccordionNav("storage")}
                    open={openNavMenus.includes("storage")}
                  >
                  ...
```

src/components/SideNavigation/SideNavigation.tsx
edlerd: it might be better to separate the SideNavigation and the SideNavigationItems, so SideNavigation only accepts children, which can be SideNavigationItems wrapped in one or multiple uls. That way we keep the SideNavigation component concise and don't have to come up with a generic interface for the items.

huwshimi: The component supports `children` if you don't want to supply items, but you lose some benefits e.g. not being able to define the link component and dark/light mode in one place and have it passed to the nav items.

edlerd: I see, you can either pass children and take over or rely on a selection of props to have a more guided experience.The pattern seems to be newly introduced to react-components. When checking existing usage of `children` they seem to be the only option for components exposing them. 

I see the benefit of allowing flexibility with children and also allowing a shorter usage with the guided props. Though I am not a big fan of it, it makes the interface a bit unclear. When using children, some of the props will be ignored. When providing both children and guided props the behaviour becomes unclear as to what is actually going to be rendered. So maybe better to only have the flexible option with children? I am not opposing the idea completely, just my opinion at this poin. I would like to get other opinions on it.

huwshimi: I've updated this to only allow items to be nested arrays of objects, or instead you pass children to replace the contents.

edlerd (approved):

Issue #621, opened 2021-10-13
Filter panel opens when a chip is removed


ilayda21:
### Description of the bug
When a chip is removed from the search and filter component,  filter panel is displayed. 

### Steps to reproduce the behavior:
- Go to https://canonical-web-and-design.github.io/react-components/?path=/story/search-and-filter--default-story
- Add a filter.
- Remove the chip.
- Filter panel opens.

### Expected behavior
After deleting a chip, filter panel shouldn't open.

![1](https://user-images.githubusercontent.com/24393395/137158310-2f78b7f3-51da-4c39-bf13-99c6e822b428.gif)

marinacastejon:
We discussed about it in the squad. While it might make sense in some cases to encourage people to select another filter, in many cases removing filter should just update the information displayed (in our case a graph) without re-opening the filter panel again. In e-commerce this is the standard. In the IBM Carbon design system they also follow that, [see example](https://cloud.ibm.com/docs/search?q=cloud).
Therefore, we would like a version of this component that does not open the panel when removing the filter.
@bartaz @ziheliu214

bartaz:
It will be discussed in WG in canonical-web-and-design/vanilla-framework#4076

clagom:
- The current implementation of the component didn't take into consideration the UX specs: https://discourse.ubuntu.com/t/behaviour-delete-a-filter-chip/21546
In particular:
- The input field should include chips (currently in a different container, separated from the input field)
- Users should be able to use backspace to remove chips (currently is not possible because of the separation of the elements)
- If the focus is on the input field, it should open the panel underneath
- If there's no focus on the input field, the panel shouldn't open
- Clicking with the cursor on the "X" to remove a chip from the input field shouldn't therefore trigger the panel to open

WG: please refer to the UX specifications in order to iterate on the implementation of the component.
CC @bartaz @sowasred2012

ziheliu214:
> > Clicking with the cursor on the "X" to remove a chip from the input field shouldn't therefore trigger the panel to open
> 
> I believe that is opposite to the UX spec:
> 
> ![image](https://user-images.githubusercontent.com/1413534/138096699-a5e611fe-a117-4d70-ac38-92166e1776fd.png)

You are right, there should be a middle stage between the last 2 interaction stage - after removing the chip, the field remains inactive. Once on focus (with the blue outline) the panel is open.

clagom:
[...]

PR #1093, opened 2024-05-23
fix: search and filter panel no longer opens when dismissing a filter

jmuzina:
## Done

Fixes the search & filter component re-opening the filter panel when removing a filter. The component will now behave as it does in the [behaviour spec](https://discourse.ubuntu.com/t/behaviour-delete-a-filter-chip/21546)

### The problem
The search & filter component has an [on-click handler](https://github.com/canonical/react-components/blob/288b3856b3133e14b8ced3b5b830211403eb92bd/src/components/SearchAndFilter/SearchAndFilter.tsx#L239) that opens the filter panel. This is currently being triggered by any clicks within the search and filter element and its descendants. This means that when a user clicks a chip dismissal button, the event bubbles up to the search & filter component & shows the filter panel.

### Proposed solution
1. Added the dismissal event to the parameter list of the `Chip` component's `onDismiss` callback. 
2. Call `event.stopPropagation` in the `onDismiss` callback passed into the filter chips by the search & filter component

### Another possible solution
I initially tried another solution that removed the `onClick` handler from the search and filter component and moved it to the `onFocus` callback of the search input textbox. This also solves the issue, however it loses some functionality as clicking anywhere in the search and filter no longer opens the filter panel, and the only way to open the filter panel when it is overflowed is to click the CTA. So, I abandoned this approach for the approach in this PR.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

1. Open [Search & Filter component](https://react-components-1093.demos.haus/?path=/story/components-searchandfilter--with-existing-search-data)
2. Click the component to open the filter panel.
3. Add several filters.
4. Close the filter panel by clicking outside of the component.
5. Dismiss a filter by clicking its "x" button.
6. Observe that the filter has been removed and the filter panel is still closed.

### Percy steps

- No visual changes expected

## Fixes

Fixes: #621

### DIFF
--- a/src/components/SearchAndFilter/SearchAndFilter.tsx
+++ b/src/components/SearchAndFilter/SearchAndFilter.tsx
@@ -264,7 +264,11 @@ const SearchAndFilter = ({
               lead={chip.lead}
               value={chip.value}
               key={`search-${chip.lead}+${chip.value}`}
-              onDismiss={() => removeFromSelected(chip)}
+              onDismiss={(event) => {
+                // Prevent filter chip dismissals from bubbling up and triggering the parent onClick handler
+                event.stopPropagation();
+                removeFromSelected(chip);
+              }}
               selected={true}
               quoteValue={chip.quoteValue}
             />
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -35,7 +35,7 @@ export type Props = PropsWithSpread<
     /**
      * Function for handling dismissing a chip.
      */
-    onDismiss?: () => void;
+    onDismiss?: (event: MouseEvent<HTMLButtonElement>) => void;
     /**
      * Whether the chip is selected.
      */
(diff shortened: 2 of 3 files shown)


pastelcyborg (commented):
I think `stopPropagation` works well here - LGTM unless @bartaz has other thoughts.

src/components/SearchAndFilter/SearchAndFilter.tsx
```diff
@@ -261,7 +261,11 @@ const SearchAndFilter = ({
               lead={chip.lead}
               value={chip.value}
               key={`search-${chip.lead}+${chip.value}`}
-              onDismiss={() => removeFromSelected(chip)}
+              onDismiss={(event) => {
+                // Prevent filter chip dismissals from bubbling up and triggering the parent onClick handler
+                event.stopPropagation();
```
huwshimi, line 266: It might be a good idea to add a test for this change to catch this issue if someone updates this code.

jmuzina: Test has been added

huwshimi: Thanks!

dgtlntv (approved):
LGTM :)

edlerd:
Wow, this has been around for a long time. Great improvement, I like it :+1:

PR #1100, opened 2024-06-20
feat(accordion): allow to use a ReactNode as title for an accordion

lorumic:
## Done

- Changed the type of the `title` prop of the `AccordionSection` to be `ReactNode` instead of `string`. Since `ReactNode` is a superset of `string`, it is automatically backward-compatible and non-breaking.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check the newly added accordion example in Storybook.

### Percy steps

- One example was added to the accordion page of the Storybook, so that is likely to be flagged as a diff by Percy. Nothing else should be flagged.

### DIFF
--- a/src/components/Accordion/Accordion.stories.mdx
+++ b/src/components/Accordion/Accordion.stories.mdx
@@ -190,3 +190,55 @@ The expanded accordion section can be controlled by external state.
     {Template.bind({})}
   </Story>
 </Canvas>
+
+### Custom headings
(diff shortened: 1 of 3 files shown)


edlerd (approved):
LGTM :+1:

PR #1101, opened 2024-06-27
chore: Allow Contextual Menu Component to be positioned optimally [WD-10977]

Kxiru:
… vertically

## Done

- Added a scss file for Contextual Menu
- Implemented initial logic by @mas-who to vertically position the Menu optimally based on the viewport height.
- Then transferred this logic from AdjustVerticalHeight function in Contextual Menu, to updateVerticalPosition function in ContextualMenuDropdown.tsx such that the menu may be adjusted according to both the container height and the viewport height.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### Percy steps

N/A

## Fixes
- Contextual Menu dropdown now displays "upwards" if there is not enough space downwards.

image

### DIFF
--- a/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.test.tsx
+++ b/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.test.tsx
@@ -171,4 +171,180 @@ describe("ContextualMenuDropdown ", () => {
       ).toBe("right");
     });
   });
+
+  describe("dropdown menu vertical positioning", () => {
+    const renderMenu = (positionNode: HTMLElement, menuHeight: number) => {
+      // need to mock overflow property for scrollparent
+      jest
+        .spyOn(global, "getComputedStyle")
+        .mockImplementation((node: Element) => {
+          if (node.id === "scrollParent") {
+            return {
+              overflowY: "auto",
+              overflowX: "auto",
+            } as CSSStyleDeclaration;
+          }
+          return {} as CSSStyleDeclaration;
+        });
+
+      // render the component once to setup initial jsdom elements
+      const links = Array.from({ length: 10 }).map((_, i) => i);
+      const { rerender } = render(
+        <ContextualMenuDropdown
+          autoAdjust
+          setAdjustedPosition={jest.fn()}
+          isOpen
+          links={links}
+          positionNode={positionNode}
+        />
+      );
+
+      // get the dropdown menu dom element and set its height
+      // NOTE: we can only do this after the component has been rendered at least once
+      let dropdownNode = document.querySelector(
+        ".p-contextual-menu__dropdown"
+      ) as HTMLElement;
+
+      dropdownNode.getBoundingClientRect = jest
+        .fn()
+        .mockReturnValue({ height: menuHeight });
+
+      rerender(
+        <ContextualMenuDropdown
+          autoAdjust
+          setAdjustedPosition={jest.fn()}
+          isOpen
+          links={links}
+          positionNode={positionNode}
+        />
+      );
+    };
+
+    const assertMenuPosition = async (position: "above" | "below") => {
+      await waitFor(() => {
+        const dropdownNode = document.querySelector(
+          ".p-contextual-menu__dropdown"
+        ) as HTMLElement;
+
+        let condition = expect(dropdownNode);
+        if (position === "below") {
+          condition = condition.not as jest.JestMatchers<HTMLElement>;
+        }
+
+        condition.toHaveAttribute("style", "bottom: 0px;");
+      });
+    };
+
+    it("places menu below toggle if there is enough space in window and scroll parent", async () => {
+      global.innerHeight = 1000;
+
+      const scrollParent = document.createElement("div");
+      jest.spyOn(scrollParent, "getBoundingClientRect").mockReturnValue({
+        top: 0,
+        bottom: 0,
+        height: 1000,
+      } as DOMRect);
+      scrollParent.id = "scrollParent";
+
+      const positionNode = document.createElement("div");
+      jest.spyOn(positionNode, "getBoundingClientRect").mockReturnValue({
+        top: 500,
+        bottom: 480,
+        height: 20,
+      } as DOMRect);
+
+      scrollParent.appendChild(positionNode);
+      document.body.appendChild(scrollParent);
+
+      const menuHeight = 200;
+      renderMenu(positionNode, menuHeight);
+
+      await assertMenuPosition("below");
+
+      positionNode.remove();
+    });
+
+    it("places menu below toggle if there is not enough space in window or scroll parent, but there is more space below than above", async () => {
+      global.innerHeight = 600;
+
+      const scrollParent = document.createElement("div");
+      jest.spyOn(scrollParent, "getBoundingClientRect").mockReturnValue({
+        top: 180,
+        bottom: 400,
+        height: 20,
+      } as DOMRect);
+      scrollParent.id = "scrollParent";
+
+      const positionNode = document.createElement("div");
+      jest.spyOn(positionNode, "getBoundingClientRect").mockReturnValue({
+        top: 100,
+        bottom: 480,
+        height: 20,
+      } as DOMRect);
+
+      scrollParent.appendChild(positionNode);
+      document.body.appendChild(scrollParent);
+
+      const menuHeight = 200;
+      renderMenu(positionNode, menuHeight);
+
+      await assertMenuPosition("below");
+
+      positionNode.remove();
+    });
+
+    it("places menu above if there is not enough space in the window", async () => {
+      global.innerHeight = 10;
+
+      const positionNode = document.createElement("div");
+      jest.spyOn(positionNode, "getBoundingClientRect").mockReturnValue({
+        top: 0,
+        bottom: 50,
+        height: 50,
+      } as DOMRect);
+
+      jest.spyOn(document.body, "getBoundingClientRect").mockReturnValue({
+        height: 10,
+        bottom: 0,
+        top: 0,
+      } as DOMRect);
+      document.body.appendChild(positionNode);
+
+      const menuHeight = 200;
+      renderMenu(positionNode, menuHeight);
+
+      await assertMenuPosition("above");
+
+      positionNode.remove();
+    });
+
+    it("places menu above if there is enough space in the window, but not enough space in the scroll parent", async () => {
+      global.innerHeight = 500;
+
+      const scrollParent = document.createElement("div");
+      jest.spyOn(scrollParent, "getBoundingClientRect").mockReturnValue({
+        top: 0,
+        bottom: 0,
+        height: 300,
+      } as DOMRect);
+      scrollParent.id = "scrollParent";
+
+      const positionNode = document.createElement("div");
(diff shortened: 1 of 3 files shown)


huwshimi (commented):
It looks like this is already intended to be supported using the `autoAdjust` prop, however it currently takes into account the scroll area, not the window space.

If there's room in the scroll area below:

Screenshot 2024-06-28 at 9 28 14 AM

If there's no room below:

Screenshot 2024-06-28 at 9 28 34 AM

It looks like this could be fixed where the vertical space is calculated:

https://github.com/canonical/react-components/blob/f704e2ed5742fff7b648b9d02f64fe4d7630ad18/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx#L228

Kxiru:
Thanks @huwshimi for your review. Are you saying that the implemented logic would be better placed / integrated into updateVerticalPosition?

huwshimi:
> Thanks @huwshimi for your review. Are you saying that the implemented logic would be better placed / integrated into updateVerticalPosition?

Yes please!

huwshimi (changes requested):
This component already has support for adjusting the dropdown to be above or below, so there's no need to add additional state. Instead, you can just change the calculation from being relative to the scroll parent, to being relative to the window here:

https://github.com/canonical/react-components/blob/9668ce671a76d74cb3d7f1153011ad2d7c73062d/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx#L239

Kxiru:
> This component already has support for adjusting the dropdown to be above or below, so there's no need to add additional state. Instead, you can just change the calculation from being relative to the scroll parent, to being relative to the window here:
> 
> https://github.com/canonical/react-components/blob/9668ce671a76d74cb3d7f1153011ad2d7c73062d/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx#L239

Don't we want the component to be able to react according to BOTH the scroll parent and the window as well?

huwshimi:
> Don't we want the component to be able to react according to BOTH the scroll parent and the window as well?

Are you thinking of a scenario where the menu is at the bottom of a scrolling container, but that container is not at the bottom of the screen, so you want the menu to flip to above the button?

If that's the behaviour you want then you could add a second rect that is relative to the window and then adjust this calculation:

https://github.com/canonical/react-components/blob/main/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx#L250-L252

This would use both rects to calculate "dropdown fits below in scroll parent" AND "dropdown fits below in window", in pseudo code this could be something like:

```typescript
    setVerticalPosition(
      scrollParentSpaceBelow >= dropdownHeight  && windowSpaceBelow >= dropdownHeight || windowSpaceBelow > spaceAbove ? "bottom" : "top"
    );
```

Kxiru:
@huwshimi Hi there!
I have actioned your suggestions. Please let me know if this is in line with what you were thinking :).

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -12,6 +12,7 @@ import { useWindowFitment } from "hooks";
 import Button from "../../Button";
 import type { ButtonProps } from "../../Button";
 import type { WindowFitment } from "hooks";
+// import "./ContextualMenuDropdown.scss";
```
edlerd, line 15:
```suggestion
```

huwshimi (commented):
This is on the right track now, thanks for persisting. Right now the relation to the window isn't correct so the menu is appearing at the top when there's space below:

Screenshot 2024-07-15 at 9 38 21 AM

This PR could also use some tests, they might help during implementation to track down the issues I've mentioned.

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -243,12 +244,26 @@ const ContextualMenuDropdown = <L,>({
...
+    const windowRect = {
+      top: window.scrollY,
+      bottom: window.scrollX,
+      height: window.innerHeight,
+    };
+
+    const scrollParentSpaceBelow =
+      scrollableParentRect.height - relativeRect.bottom;
+    const scrollParentSpaceAbove = relativeRect.top;
     const dropdownHeight = relativeRect.height;
```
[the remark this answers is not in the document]
Kxiru: Thank you for your suggestions. Please review :).

Kxiru:
@huwshimi , thanks for our chat today! I've pushed the changes that we made and the written suggestions you provided. I'll let you know of my progress with creating a test.

mas-who:
@huwshimi @Kxiru regarding creating a test for this component, I believe since react testing library uses jsdom, it will be tricky to setup a test that relies on the component render cycle and perform assertion on the final element positions. See this [issue](https://github.com/testing-library/react-testing-library/issues/353#issuecomment-545231016) for example which has a suggestion of directly modifying dom node's prototype to mock the values but I don't think it would work for assertions if we want to check dom element position. I think maybe we can setup the test by moving `updateVerticalPosition` outside of the component and create a unit test for the function?

edlerd:
> @huwshimi @Kxiru regarding creating a test for this component, I believe since react testing library uses jsdom, it will be tricky to setup a test that relies on the component render cycle and perform assertion on the final element positions. See this [issue](https://github.com/testing-library/react-testing-library/issues/353#issuecomment-545231016) for example which has a suggestion to directly modify a dom node's prototype to mock the values but I don't think it would work for assertions if we want to check dom element position. I think maybe we can setup the test by moving `updateVerticalPosition` outside of the component and create a unit test for the function?

If I understand your problem correctly @mas-who @Kxiru you can have a look at the `updates the tooltip to fit on the screen` test in `Tooltip.test.jsx` for a solution of a very similar problem.

mas-who:
> > @huwshimi @Kxiru regarding creating a test for this component, I believe since react testing library uses jsdom, it will be tricky to setup a test that relies on the component render cycle and perform assertion on the final element positions. See this [issue](https://github.com/testing-library/react-testing-library/issues/353#issuecomment-545231016) for example which has a suggestion to directly modify a dom node's prototype to mock the values but I don't think it would work for assertions if we want to check dom element position. I think maybe we can setup the test by moving `updateVerticalPosition` outside of the component and create a unit test for the function?
> 
> If I understand your problem correctly @mas-who @Kxiru you can have a look at the `updates the tooltip to fit on the screen` test in `Tooltip.test.jsx` for a solution of a very similar problem.

Hmm not sure if it's the same problem, the calculation performed for positioning the dropdown isn't really based on event listeners. As far as I can tell, what we can assert for this component test case is that it's `top` and `bottom` attributes have been correctly calculated. Note this is due to inline style being [applied](https://github.com/Kxiru/react-components/blob/a135e875be8ec29d05723d20ce62f911c0d3afa9/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx#L311). To do that, we need to get the dropdown element from the dom after rendering the component, which would have zero dimensions as discussed in the reference issue. This would cause assertion to fail.

huwshimi:
To get this to work you'll need to mock the functions that don't have dimensions, like we do in this test: https://github.com/canonical/react-components/blob/2825eb98ed253825ef69e63c261c7c9605b15f4f/src/hooks/useWindowFitment.test.ts#L72-L78.

You'll need to do this for the scroll parent, the position node and the dropdown elements.

You'll also need to double render the component as RTL doesn't do full lifecycle rendering so we have to manually rerender the component to make the hooks run after the elements exist in the DOM:

https://github.com/canonical/react-components/blob/2825eb98ed253825ef69e63c261c7c9605b15f4f/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.test.tsx#L42-L44.

If that doesn't make sense let me know and I can give you a concrete example of how to construct the test.

Kxiru:
@huwshimi , @mas-who has pushed a commit that  recreates the test environment to check whether VerticalPosition resolves to top or bottom - and consequently, whether the styling attribute of the component Span has a bottom of 0, or not. Please could you let us know your thoughts of the test implementation?

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
```diff
@@ -243,27 +243,19 @@ const ContextualMenuDropdown = <L,>({
       bottom: toggleRect.bottom - scrollableParentRect.top,
     };
 
-    // Calculate the rect in relation to the Window
-    const windowRect = {
```
Kxiru: For context, this has been removed as the only attribute that is later referenced, is windowRect.height, i.e; window.innerHeight - as replaced/referenced in line 252.

huwshimi:
> @huwshimi , @mas-who has pushed a commit that recreates the test environment to check whether VerticalPosition resolves to top or bottom - and consequently, whether the styling attribute of the component Span has a bottom of 0, or not. Please could you let us know your thoughts of the test implementation?

This looks good! I think there are two other cases that need tests:
- that the menu is above if there is enough area in the window, but not in the scroll parent
- that the menu is below if there is not enough area in the window or scroll parent, but also that there is more space below than above.

I also think it could be worth splitting the current test into the two assertions to make it clear which mocks are required to make each test pass. It might be helpful to move some of the set up into a function to make the four tests cleaner.

mas-who:
> > @huwshimi , @mas-who has pushed a commit that recreates the test environment to check whether VerticalPosition resolves to top or bottom - and consequently, whether the styling attribute of the component Span has a bottom of 0, or not. Please could you let us know your thoughts of the test implementation?
> 
> This looks good! I think there are two other cases that need tests:
> 
> * that the menu is above if there is enough area in the window, but not in the scroll parent
> * that the menu is below if there is not enough area in the window or scroll parent, but also that there is more space below than above.
> 
> I also think it could be worth splitting the current test into the two assertions to make it clear which mocks are required to make each test pass. It might be helpful to move some of the set up into a function to make the four tests cleaner.

@huwshimi I've made those changes as mentioned :+1:

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.test.tsx
```diff
@@ -171,4 +171,180 @@ describe("ContextualMenuDropdown ", () => {
       ).toBe("right");
     });
   });
+
+  describe("dropdown menu vertical positioning", () => {
+    const renderMenu = (positionNode: HTMLElement, menuHeight: number) => {
+      // need to mock overflow propertty for scrollparent
```
huwshimi, line 177:
```suggestion
      // need to mock overflow property for scrollparent
```

huwshimi (approved):
Thanks for this @Kxiru and @mas-who this looks like it's working well now!

src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.test.tsx
mas-who: Thanks for the review @huwshimi !

edlerd (approved):
LGTM

PR #1105, opened 2024-07-15
ci(tics): Don't upload TICS report to artifacts

bartaz:
## Done

For security reasons we are removing the step that uploads TICS report as action artifact.


Lukewh (approved):
:+1: LGTM

PR #1111, opened 2024-08-09
fix: remove changing title in LoginPageLayout

vladimir-cucu:
## Done

- Don't change title when `LoginPageLayout` renders. This allow for a consistent behavior with other components and will fix bugs like the one bellow.
- Conditionally remove `is-paper` body class on unmount.

## Fixes

Fixes: https://github.com/canonical/identity-platform-admin-ui/issues/373
https://warthogs.atlassian.net/browse/WD-14064

### DIFF
--- a/src/components/ApplicationLayout/ApplicationLayout.test.tsx
+++ b/src/components/ApplicationLayout/ApplicationLayout.test.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { render, screen, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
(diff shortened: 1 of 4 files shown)


bartaz (approved):
Code looks good to me. I don't have a lot of context on how this would be used, but seems to do what intended.

huwshimi:
I wonder if this might not be the right approach. I'm not sure the layout components should be managing the window title as I can imagine different apps would want to manage the title through different implementations e.g. through the use of hooks.

Maybe the fix would be to not set the document title in `LoginPageLayout` and allow each application to manage the window title themselves. This would be a breaking change, but worth it I think.

I also noticed that the LoginPageLayout adds an `is-paper` body class but doesn't clean it up when unmounted so there's a couple of unintended side effects by using this component.

huwshimi (changes requested):
I wonder if this might not be the right approach. I'm not sure the layout components should be managing the window title as I can imagine different apps would want to manage the title through different implementations e.g. through the use of hooks.

Maybe the fix would be to not set the document title in `LoginPageLayout` and allow each application to manage the window title themselves. This would be a breaking change, but worth it I think.

I also noticed that the LoginPageLayout adds an `is-paper` body class but doesn't clean it up when unmounted so there's a couple of unintended side effects by using this component.

huwshimi (approved):
Thanks for updating this. I think it would be worth noting that this is a breaking change in the release: https://github.com/semantic-release/semantic-release?tab=readme-ov-file#commit-message-format.

PR #1113, opened 2024-08-19
feat: add MultiSelect variations

huwshimi:
## Done

- Add some additional props to customise the MultiSelect for ReBAC Admin.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -108,6 +108,18 @@ Then in `react-components` run:
 dotrun unlink-package
 ```
 
+## Linking from Vite
+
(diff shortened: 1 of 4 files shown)


vladimir-cucu (approved):
LGTM! Was wondering if it might make sense to add an additional example to the storybook with `listSelected = false`?

huwshimi:
> LGTM! Was wondering if it might make sense to add an additional example to the storybook with `listSelected = false`?

Maybe the listSelected is too small of a variation to warrant a separate example? The `listSelected` prop can be toggled in the controls section of the demo:

Screenshot 2024-08-20 at 9 34 15 AM

PR #1115, opened 2024-08-20
feat: support not blurring when searching

huwshimi:
## Done

- Update SearchBox to have a prop to disable blurring on search.
- Pass the search value to the onSearch prop.

## QA

- Go to [this demo](https://react-components-1115.demos.haus/?path=/story/components-searchbox--default&args=shouldBlurOnSearch:!false)
- In the controls tab set `shouldBlurOnSearch` to "false"
- Type something into the search box and press enter.
- The input should not blur.

### DIFF
--- a/src/components/SearchBox/SearchBox.tsx
+++ b/src/components/SearchBox/SearchBox.tsx
@@ -35,7 +35,7 @@ export type Props = PropsWithSpread<
     /**
      * A function that is called when the user clicks the search icon or presses enter
      */
-    onSearch?: () => void;
+    onSearch?: (inputValue: string) => void;
     /**
      * A function that is called when the user clicks the reset icon
      */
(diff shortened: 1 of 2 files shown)


src/components/SearchBox/SearchBox.tsx
```diff
@@ -35,7 +35,7 @@ export type Props = PropsWithSpread<
     /**
      * A function that is called when the user clicks the search icon or presses enter
      */
-    onSearch?: () => void;
+    onSearch?: (inputValue: string) => void;
```
vladimir-cucu, line 38: I think this might be a breaking change.

vladimir-cucu (approved):
Looks good! :rocket:

src/components/SearchBox/SearchBox.tsx
huwshimi: This is OK as it is backwards compatible. Existing `onSearch` functions like `() => "something"` or with a type of `() => void` will continue to work.

PR #1118, opened 2024-08-22
chore: forwardRef from SideNavigationLink component

aaryanporwal:
## Done

- use forwardRef in SideNavigationLink Component

### Storybook

To see rendered examples of all react-components, run:

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

Install the resulting tarball in your project with:

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

- List any expected visual change in Percy, or write something like "No visual changes expected" if none is expected.

## Fixes

Fixes: # .

PR #1123, opened 2024-09-02
fix: forwardRef from SideNavigationLink component

### LINKED ISSUE #1122: forwardRef error
After upgrading to 1.2.1 the following error occurs:

```
Warning: forwardRef render functions accept exactly two parameters: props and ref. Did you forget to use the ref parameter?
```

It looks like it was caused by this change: https://github.com/canonical/react-components/pull/1118.

aaryanporwal:
## Done

- Fixes #1122 
- [WD-14675](https://warthogs.atlassian.net/browse/WD-14675)

### Percy steps

- No visual changes expected

[WD-14675]: https://warthogs.atlassian.net/browse/WD-14675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/SideNavigation/SideNavigationLink/SideNavigationLink.tsx
+++ b/src/components/SideNavigation/SideNavigationLink/SideNavigationLink.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef } from "react";
+import React from "react";
 import classNames from "classnames";
 import type { HTMLProps } from "react";
 
@@ -7,7 +7,7 @@ import SideNavigationBase from "../SideNavigationBase";
 
 export type LinkDefaultElement = HTMLProps<HTMLAnchorElement>;
 
-export type Props<L = LinkDefaultElement> = Omit<
+export type Props<L = LinkDefaultElement, E = HTMLAnchorElement> = Omit<
   SideNavigationBaseProps<L>,
   "component"
 > & {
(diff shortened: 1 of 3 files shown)


huwshimi (changes requested):
It looks like most of the changes need to be reverted from the previous PR: https://github.com/canonical/react-components/pull/1118/files as removing the generic types has broken then ability to handle props when passing `component`.

For example when using this with a `Link` from React Router it would give an error for the `to` prop here as it can't infer the props of the `Link` component:

```typescript
<SideNavigationLink label="docs" component={Link} to="/docs" />
```

Unfortunately `forwardRef` does not support components with generic types, so instead  you have to manually pass a ref. We do that in a few places e.g. https://github.com/canonical/react-components/blob/f60d6bc1f2cb62436a3585c7827d89535d209013/src/components/Panel/Panel.tsx#L136

This issue could be caught in the future if we change this test: https://github.com/canonical/react-components/blob/f60d6bc1f2cb62436a3585c7827d89535d209013/src/components/SideNavigation/SideNavigationLink/SideNavigationLink.test.tsx#L34

To:

```typescript
it("can use a custom link component", () => {
  const Link = ({ to }: { to: () => void }) => (
    <button onClick={to}>link</button>
  );
  const label = "Test content";
  render(<SideNavigationLink label={label} component={Link} to={jest.fn()} />);
  expect(screen.getByRole("button", { name: label })).toHaveClass(
    "p-side-navigation__link",
  );
});
```

aaryanporwal:
> It looks like most of the changes need to be reverted from the previous PR: https://github.com/canonical/react-components/pull/1118/files as removing the generic types has broken then ability to handle props when passing `component`.
> 
> For example when using this with a `Link` from React Router it would give an error for the `to` prop here as it can't infer the props of the `Link` component:
> 
> ```ts
> <SideNavigationLink label="docs" component={Link} to="/docs" />
> ```
> 
> Unfortunately `forwardRef` does not support components with generic types, so instead you have to manually pass a ref. We do that in a few places e.g.
> 
> https://github.com/canonical/react-components/blob/f60d6bc1f2cb62436a3585c7827d89535d209013/src/components/Panel/Panel.tsx#L136
> 
> This issue could be caught in the future if we change this test:
> 
> https://github.com/canonical/react-components/blob/f60d6bc1f2cb62436a3585c7827d89535d209013/src/components/SideNavigation/SideNavigationLink/SideNavigationLink.test.tsx#L34
> 
> To:
> 
> ```ts
> it("can use a custom link component", () => {
>   const Link = ({ to }: { to: () => void }) => (
>     <button onClick={to}>link</button>
>   );
>   const label = "Test content";
>   render(<SideNavigationLink label={label} component={Link} to={jest.fn()} />);
>   expect(screen.getByRole("button", { name: label })).toHaveClass(
>     "p-side-navigation__link",
>   );
> });
> ```

Thank you Huw for pointing that out! I've updated the code and the test like you suggested. Can you please take a look again?

src/components/SideNavigation/SideNavigationLink/SideNavigationLink.tsx
```diff
@@ -16,22 +16,30 @@ export type Props<L = LinkDefaultElement> = Omit<
    * @default a
    */
   component?: SideNavigationBaseProps<L>["component"];
+  /**
+   * A ref to pass to the link element.
+   */
+  forwardRef?: React.Ref<HTMLAnchorElement> | null;
```
huwshimi, line 22: The element won't always be an anchor (if someone has passed the `component` prop) so this might need a new generic that could be set to `HTMLAnchorElement` by default.

huwshimi (commented):
This change also needs to be reverted https://github.com/canonical/react-components/pull/1118/files#diff-011b080bc51b47c868f250b3a34619b5f41e4c16c8e39a6636ccc719b6da8bd6L40

huwshimi (approved):
Thanks for this fix!

PR #1128, opened 2024-09-04
feat: Upstream Doughnut Chart to RC

Kxiru:
## Done

- Added the Doughnut Chart Component

## TODOs
- ~~Correct failing test that verifies whether a tooltip shows on hover.~~

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Run Storybook
- Navigate to Doughnut Chart component

### Percy steps

## Fixes

Fixes: N/A

### DIFF
--- a/src/components/DoughnutChart/DoughnutChart.scss
+++ b/src/components/DoughnutChart/DoughnutChart.scss
@@ -1,0 +1,30 @@
+@import "vanilla-framework";
+
+.doughnut-chart {
+  width: 6.5rem;
+
+  .doughnut-chart__tooltip {
+    display: block;
+  }
+
+  .doughnut-chart__tooltip > :only-child {
+    // Override the tooltip wrapper.
+    display: block !important;
+  }
+
+  .doughnut-chart__chart {
+    // Restrict hover areas to the strokes.
+    pointer-events: stroke;
+  }
+
+  .doughnut-chart__segment {
+    fill: transparent;
+
+    // Animate stroke size changes on hover.
+    transition: stroke-width 0.3s ease;
+  }
+}
+
+.doughnut-chart__legend {
+  list-style-type: none;
+}
--- a/src/components/DoughnutChart/DoughnutChart.test.tsx
+++ b/src/components/DoughnutChart/DoughnutChart.test.tsx
@@ -1,0 +1,76 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import React from "react";
+import DoughnutChart, { TestIds } from "./DoughnutChart";
+
+describe("DoughnutChart", () => {
+  const defaultProps = {
+    chartID: "test",
+    segmentHoverWidth: 10,
+    segmentThickness: 8,
+    size: 100,
+    segments: [
+      {
+        color: "#3498DB",
+        tooltip: "aaa",
+        value: 12,
+      },
+      {
+        color: "#E74C3C",
+        tooltip: "bbb",
+        value: 8,
+      },
+      {
+        color: "#F1C40F",
+        tooltip: "ccc",
+        value: 18,
+      },
+      {
+        color: "#2ECC71",
+        tooltip: "ddd",
+        value: 14,
+      },
+    ],
+  };
+
+  it("renders", () => {
+    render(<DoughnutChart {...defaultProps} />);
+    expect(screen.getByTestId("chart")).toBeInTheDocument();
+  });
+
+  it("displays the correct number of segments", () => {
+    render(<DoughnutChart {...defaultProps} />);
+    const segments = screen.getAllByTestId(TestIds.Segment);
+    expect(segments).toHaveLength(defaultProps.segments.length);
+  });
+
+  it("shows tooltips on hover", async () => {
+    const { container } = render(<DoughnutChart {...defaultProps} />);
+    const segments = screen.getAllByTestId(TestIds.Segment);
+
+    fireEvent.mouseOver(segments[0]);
+    fireEvent.click(container.firstChild.firstChild);
+
+    await waitFor(() => {
+      expect(screen.getByText("aaa")).toBeInTheDocument();
+    });
+  });
+
+  it("applies custom styles to segments", () => {
+    render(<DoughnutChart {...defaultProps} />);
+    const segment = screen.getAllByTestId(TestIds.Segment)[0];
+    expect(segment).toHaveStyle(`stroke: ${defaultProps.segments[0].color}`);
+    expect(segment).toHaveStyle(
+      `stroke-width: ${defaultProps.segmentThickness}`,
+    );
+  });
+
+  it("displays the label in the center if provided", () => {
+    render(<DoughnutChart {...defaultProps} label="Test Label" />);
+    expect(screen.getByTestId(TestIds.Label)).toHaveTextContent("Test Label");
+  });
+
+  it("does not display the label if not provided", () => {
+    render(<DoughnutChart {...defaultProps} />);
+    expect(screen.queryByText("Test Label")).not.toBeInTheDocument();
+  });
+});
--- a/src/components/DoughnutChart/DoughnutChart.tsx
+++ b/src/components/DoughnutChart/DoughnutChart.tsx
@@ -1,0 +1,232 @@
+import React, { FC, useRef, useState } from "react";
+import classNames from "classnames";
+import Tooltip from "components/Tooltip";
+import "./DoughnutChart.scss";
+
+export type Segment = {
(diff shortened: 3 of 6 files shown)


edlerd (commented):
This is a good start. Some ideas to improve:
- The linter has found one issue
- In storybook there is an error on the Docs and Template section for the component
- For the test you can start easy i.e. by rendering a chart with two equal segments and checking the radius is calculated correctly as 180 degrees. Or be creative for other test cases.

Kxiru:
Apologies, it seems as though my existing test suite and much of my changes did not push. Here is the latest version, and I will address any outstanding suggestions on top of this.

Kxiru:
@advl , could you give me some examples of the edge cases you are considering? :)
Also, I would be happy to implement some more graphs! I definitely think that some data visualisation components in React Components would go a long way for future projects. Two that we have perfected in the backlog include Multi-Bar (supporting Single bar) meters, and Horizontal bar charts.
image
image
image

We also have a legend for the doughnut chart which could be triggered and autogenerated by a prop?

image

Let me know your thoughts :).

src/components/DoughnutChart/DoughnutChart.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+    //     const tooltip = screen.getByTestId('tooltip-portal');
+    //     expect(tooltip).toBeInTheDocument();
+    //     expect(screen.getByRole('tooltip')).toHaveTextContent('aaa');
+    //   });
+
+    await userEvent.unhover(firstSegment);
+    await waitFor(() => {
+      expect(screen.queryByText("a")).not.toBeInTheDocument();
+    });
+  });
```
edlerd, line 68: I think you need `act` and maybe also the custom `userEventWithTimers` function for this to work. I.e. check this test from `Tooltip.test.tsx`

```
  it("updates the tooltip to fit on the screen", async () => {
    render(
      <Tooltip
        message="text that is too long to fit on the screen"
        position="right"
      >
        <button>Child</button>
      </Tooltip>
    );
    global.innerWidth = 20;
    await act(async () => {
      await userEventWithTimers.hover(
        screen.getByRole("button", { name: "Child" })
      );
      jest.runAllTimers();
    });
    expect(screen.getByTestId("tooltip-portal")).toHaveClass(
      "p-tooltip--btm-left"
    );
    expect(screen.getByTestId("tooltip-portal")).toHaveClass("is-detached");
  });
  ```

Comment (Kxiru) [reply src/components/DoughnutChart/DoughnutChart.test.tsx:68]: Heya David,
I actually attempted to implement this last week to no avail. Something that is interesting is that the console prints out the html structure of the doughnut chart and when you analyse this, the tooltip test-ids do not show. I then figured it should be necessary to add another test-id in the opening span of the tooltip, however this broke another test in the Tooltip test suite regarding the code changing and no longer matching a previous snapshot. Therefore, as altering the Tooltip component is not an option, and accessing the Tooltip as a sub-component is not an option, I do not believe this can be done at this stage.

image

The only thing that can be accessed from the section is the span which has a classname related to the tooltip (Which is also within the Tooltip component. I tried to put a test-id here), however as we cannot select by classname, this is inaccessible.

I wonder if the issue is about components not being registered as on hover, or whether Jest is simply not picking up on the Tooltip for one reason or another. In any case, I shall push my new (still failing) test for further review.

Comment (edlerd) [reply src/components/DoughnutChart/DoughnutChart.test.tsx]: This should work:
```

  it("shows tooltips on hover", async () => {
    const { container } = render(<DoughnutChart {...defaultProps} />);
    const segments = screen.getAllByTestId(TestIds.Segment);

    fireEvent.mouseOver(segments[0]);
    fireEvent.click(container.firstChild.firstChild);

    await waitFor(() => {
      expect(screen.getByText("aaa")).toBeInTheDocument();
    });
  });
  ```

Kxiru: @edlerd , This works! Interesting to note that onHover actions do not register for this element. Question though, the selector container.firstChild.firstChild, doesn't that point to the <style> tag at line 153 in DoughnutChart.tsx as opposed to a Tooltip?
I will push the revised test suite now.

jmuzina:
Looks good as-is, but I wonder if there is a way for us to make the chart a bit more accessible - screen readers and keyboard users may have a hard time navigating and understanding this chart. I'll take a brief look and let you know if there's anything that can be done quickly in scope of just upstreaming this

jmuzina:
> Looks good as-is, but I wonder if there is a way for us to make the chart a bit more accessible - screen readers and keyboard users may have a hard time navigating and understanding this chart. I'll take a brief look and let you know if there's anything that can be done quickly in scope of just upstreaming this

A quick solution would be to give the chart a title and description, that way screen readers announce the contents. You can do this by creating `<title>` and `<desc>`. 

```tsx
 <svg
    className="doughnut-chart__chart"
    id={id.current}
    viewBox={`0 0 ${canvasSize} ${canvasSize}`}
    data-testid={TestIds.Section}
    aria-labelledby={`${id.current}-chart-title ${id.current}-chart-desc`}
  >
    {label && <title id={`${id.current}-chart-title`}>{label}</title>}
    <desc id={`${id.current}-chart-desc`}>
      {segments.map(
        (segment) => {
          let description = "";
          if (segment.tooltip) description += `${segment.tooltip}: `;
          description += segment.value;
          return description;
        })
        .join(",")}
    </desc>
```

This causes a screen reader to announce the chart contents:
Screenshot 2024-09-13 at 11 44 53 AM

src/components/DoughnutChart/DoughnutChart.tsx
```diff
@@ -0,0 +1,212 @@
...
+    // Add this segment length to the running tally.
+    accumulatedLength += segmentLength;
+
+    return (
+      <circle
+        className="doughnut-chart__segment"
+        cx={radius - segmentWidth / 2 - hoverIncrease}
+        cy={radius + segmentWidth / 2 + hoverIncrease}
+        data-testid={TestIds.Segment}
+        key={i}
```
jmuzina, line 102:
```suggestion
        key={i}
        tabIndex={0}
        aria-label={`${tooltip}: ${value}`}
```
If you do something like this, the segments will also be tab-selectable and announce their contents.

jmuzina: It would also be nice to make the tooltip of a focused segment visible, so the tooltips appear on focus as well. Though that may be a bit more difficult and to be tackled as a separate issue.

src/components/DoughnutChart/DoughnutChart.scss
```diff
@@ -0,0 +1,30 @@
+@import "vanilla-framework";
+
+.doughnut-chart {
+  width: 6.5rem;
+
+  .doughnut-chart__tooltip {
+    display: block;
+  }
+
+  .doughnut-chart__tooltip > :only-child {
```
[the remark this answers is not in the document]
Kxiru: I can't say that I know exactly why this occurs, but when this section is removed, the chart fails to render upon hover. It simply just disappears :skull: , so this part prevents that from happening.
image

As for the Tooltip display:block, this can be removed, though I'm not sure if this is what you were referring to as it is not an inline-block statement.

Kxiru:
@jmuzina and @pastelcyborg , thank you for your reviews and suggestions so far. I have implemented these and await another once over whenever you are ready :).

src/components/DoughnutChart/DoughnutChart.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+    return (
+      <circle
+        className="doughnut-chart__segment"
+        cx={radius - segmentThickness / 2 - hoverIncrease}
+        cy={radius + segmentThickness / 2 + hoverIncrease}
+        data-testid={TestIds.Segment}
+        key={i}
+        tabIndex={0}
+        aria-label={`${tooltip}: ${value}`}
```
jmuzina, line 109: I've just noticed that this tooltip is optional, so if it is not provided the screen reader reads "undefined". This should be changed to only include the tooltip in the label if it is defined. Once that's done, this'll be good to go on the A11y side of things.

bartaz: @Kxiru Has this been addressed? Value of `aria-label` in situations when tooltip is not provided?

Kxiru: Just pushed the change! :)

jmuzina (approved):

PR #1133, opened 2024-09-18
fix: Update SideNavigationLink component's prop type

aaryanporwal:
## Done

- [x] Update SideNavigationLink component's prop type to handle props properly.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: # .

### DIFF
--- a/src/components/SideNavigation/SideNavigationLink/SideNavigationLink.tsx
+++ b/src/components/SideNavigation/SideNavigationLink/SideNavigationLink.tsx
@@ -1,29 +1,30 @@
 import React from "react";
 import classNames from "classnames";
-import type { HTMLProps, ReactNode } from "react";
+import type { HTMLProps } from "react";
 
 import type { SideNavigationBaseProps } from "../SideNavigationBase";
 import SideNavigationBase from "../SideNavigationBase";
+import { PropsWithSpread } from "types";
 
 export type LinkDefaultElement = HTMLProps<HTMLAnchorElement>;
 
-export type Props<L = LinkDefaultElement, E = HTMLAnchorElement> = Partial<
+export type Props<
+  L = LinkDefaultElement,
+  E = HTMLAnchorElement,
+> = PropsWithSpread<
+  {
+    /**
+     * The component or element to use for the link element e.g. `a` or `NavLink`.
+     * @default a
+     */
+    component?: SideNavigationBaseProps<L>["component"];
+    /**
+     * A ref to pass to the link element.
+     */
+    forwardRef?: React.Ref<E> | null;
+  },
   Omit<SideNavigationBaseProps<L>, "component">
-> & {
-  /**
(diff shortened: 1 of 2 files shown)


src/components/SideNavigation/SideNavigationLink/SideNavigationLink.tsx
```diff
@@ -14,6 +14,10 @@ export type Props<L = LinkDefaultElement, E = HTMLAnchorElement> = Partial<
    * The navigation item's label.
    */
   label: ReactNode;
+  /**
+   * The navigation item's onClick event handler.
+   */
+  onClick?: () => void;
```
edlerd, line 20: This looks wrong, Why add a prop, that is being ignored?
Should this be added to `SideNavigationBase` instead and also handled there?

aaryanporwal: When I add forwardRef to the usage in anbox which has onClick prop. I get the error: `Property 'onClick' does not exist on type 'IntrinsicAttributes & Partial<Omit<Props<HTMLSpanElement>, "component">> & { label: ReactNode; component?: ElementType | ... 1 more ... | undefined; forwardRef?: Ref<...> | undefined; }'. Did you mean 'onclick'?`
image

aaryanporwal: Should I add the onClick prop to SideNavBase?
Like this?

```
<Component ref={forwardRef} onClick={onClick} {...props}>
```

I don't get the above error when I add the onClick prop in SideNavBase.

huwshimi: These props should get inferred from the component/element type so this suggests a problem with the way `Props` is defined. It looks like changing `Props` to this fixes the issue:

```typescript
export type Props<
  L = LinkDefaultElement,
  E = HTMLAnchorElement,
> = PropsWithSpread<
  {
    /**
     * The navigation item's label.
     */
    label: ReactNode;
    /**
     * The component or element to use for the link element e.g. `a` or `NavLink`.
     * @default a
     */
    component?: SideNavigationBaseProps<L>["component"];
    /**
     * A ref to pass to the link element.
     */
    forwardRef?: React.Ref<E> | null;
  },
  Omit<SideNavigationBaseProps<L>, "component">
>;
```

Then you can remove the `onClick` you added.

aaryanporwal: Yes that's what I was suspecting. Thanks for the fix! It works now.

huwshimi (approved):
Looks good now, thanks for the improvement.

PR #1134, opened 2024-09-18
feat: add stepper and step components

anusha-c18:
## Done

- Added stepper wrapper component
- Added step component with options to:
   - add a title for the step 
   - enable/disable the step
   - select the step
   - display an optional label
   - display an optional link
   - display a progress line
   - display the index of the step
   - display an icon for the step
- Added tests for the step component
- Added tests for the stepper component 

The current stepper component is of the standard size outlined in the spec. If required, the large size can be added at a later time.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: # .

### DIFF
--- a/src/components/Stepper/Step/Step.scss
+++ b/src/components/Stepper/Step/Step.scss
@@ -1,0 +1,113 @@
+@import "vanilla-framework";
+
+.step-number {
+  border: 0.08rem solid black;
+  border-radius: 1rem;
+  height: 1.4rem;
+  line-height: 1.3;
+  margin-left: $sph--small;
+  margin-right: 0.1rem;
+  margin-top: 0.1rem;
+  text-align: center;
+  width: 1.4rem;
+}
+
+.step-number-disabled {
+  border: 0.08rem solid #757575;
+  color: #757575;
+}
+
+.step-content {
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  margin-left: $sph--small;
+}
+
+.step-enabled:hover {
+  cursor: pointer;
+  text-decoration: underline;
+}
+
+.step-disabled {
+  color: #757575;
+  pointer-events: none;
+}
+
+.step-status-icon {
+  height: 1.6rem;
+  margin-left: 0.4rem;
+  width: 1.6rem;
+}
+
+.step-selected {
+  background-color: var(--vf-color-background-alt);
+}
+
+.step-optional-content {
+  font-size: 12px;
+  max-width: 10rem;
+}
+
+.stepper-horizontal {
+  display: flex;
+
+  .p-inline-list__item {
+    margin: 0;
+  }
+
+  .step {
+    border-top: 0.2rem solid var(--vf-color-border-default);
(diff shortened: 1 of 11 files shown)


src/components/Stepper/Step/Step.scss
```diff
@@ -0,0 +1,115 @@
+@import "vanilla-framework";
+
+.p-list__item {
+  padding-bottom: 0;
+  padding-top: 0;
+}
+
+.p-inline-list {
+  display: flex;
+}
```
edlerd, line 10: Those look like standard vanilla classes, which we shouldn't override globally. 

I think best is 
1. to namespace all selectors in this file with a wrapper and 
2. target only custom classes

src/components/Stepper/Step/Step.tsx
```diff
@@ -0,0 +1,118 @@
...
+   */
+  enabled: boolean;
+  /**
+   * Optional value to highlight the selected step.
+   */
+  selected?: boolean;
+  /**
+   * Icon to display in the step. Specify "number" if the index should be displayed.
+   */
+  iconName: string;
```
edlerd, line 40: Should we reference the icon names `ValueOf<typeof ICONS> | string | "number"` here to allow autocomplete?

edlerd (commented):
QA LGTM.
The linter has some errors, one idea for improvement  and one concern regarding css namespacing.

src/components/Stepper/Step/Step.scss
anusha-c18: Agreed, I've shifted the css into the `stepper-vertical` and `stepper-horizontal` classes.

src/components/Stepper/Step/Step.tsx
anusha-c18: While having autocomplete would be nice, adding the general `string` type to the union expands the type, which prevents ts from providing autocomplete suggestions.

src/components/Stepper/Step/Step.scss
```diff
@@ -0,0 +1,113 @@
+@import "vanilla-framework";
+
+.p-inline-list__item {
+  margin: 0;
+}
```
edlerd, line 5: This is overriding a vanilla class on the root level. It should be namespaced.

edlerd (changes requested):

src/components/Stepper/Step/Step.scss
anusha-c18: Good catch - addressed the [changes](https://github.com/canonical/react-components/compare/ea71adb0101ff639008ef938152a337c3bf734bc..00fb6effdfa8ff1b63b7af74f3ee26d201ce47c7)

edlerd (approved):
QA and Code LGTM.

We need to have a look at percy and eventually approve the new screenshot introduced by this PR.

PR #1140, opened 2024-09-26
chore(ci): Remove unnecessary Percy label CI check

bartaz:
## Done

Now that Percy CI check is fully integrated we can remove the workflow that checks for a label.


jmuzina:
As this workflow runs on `pull_request_target`, I think you won't see the changes from this PR until this is merged into `main`. So the workflow ran, but it shouldn't run on future PRs.

jmuzina (approved):

jmuzina:
FYI, I added the Percy integration status check as a required PR check - it should pass here as soon as Percy finishes running.

PR #1143, opened 2024-09-30
feat(skip-link): add SkipLink component [WD-15080]

lorumic:
## Done

- Added new `SkipLink` component, based on Vanilla [`p-skip--link`](https://vanillaframework.io/docs/patterns/links#skip-link) pattern. Added tests and Storybook for it.
- Integrated the `SkipLink` component inside the `ApplicationLayout`, so that it can be automatically obtained by React applications using the `ApplicationLayout`.
- :warning: **If you are manually adding an `a` element with the `p-skip--link` class to your React application, and are also using the `ApplicationLayout` component, this can be a breaking change for you**.
Please make sure that you remove your manual addition of the skip link when upgrading react-components to the version that includes this change.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to the `SkipLink` page of Storybook.
- Click inside the container of the default example, to move the focus there.
- Hit "Tab", and check that the "Skip to main content" link appears.

### Percy steps

- There should be a new page in Storybook, for the `SkipLink` component.

## Fixes

Fixes: [WD-15080](https://warthogs.atlassian.net/browse/WD-15080)

[WD-15080]: https://warthogs.atlassian.net/browse/WD-15080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/SkipLink/SkipLink.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+/**
+ * This is a [React](https://reactjs.org/) component for the Vanilla [Skip link](https://vanillaframework.io/docs/patterns/links#skip-link) component.
+ */
+export const SkipLink = (): JSX.Element => {
```
mas-who, line 6: I think `href` should be a prop. The detail that the `href` reference must link up with the id of `AppMain`, and it has to be `main-content` is currently hidden inside this component. Having `href` would help especially In case if a downstream project decides to not use `ApplicationLayout` and use `AppMain` directly.

src/components/SkipLink/SkipLink.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { Meta, StoryObj } from "@storybook/react";
```
mas-who, line 1: it feels a little strange that the doc for this component is almost just a blank page. Should we add some text here to describe the relationship between this component and `AppMain`? Maybe something like the below:

```
This component is intended to be used alongside the AppMain component, with its href attribute pointing to the ID of AppMain.
```

Also, is it possible to add a `main` element to the doc with `id` linked up to `href` of `SkipLink`?

mas-who (commented):
This looks mostly good! Just some small suggestions about the documentation and component interface :)

src/components/SkipLink/SkipLink.stories.tsx
lorumic: Actually, there is no particular relationship between this component and `AppMain` - nor with a `main` HTML element. You're not constrained to use this together with `AppMain`, but rather, you can use it in whichever layout you prefer - it's completely flexible and up to your particular case. And this is also the reason why I wanted to keep the example as simple and implementation-agnostic as possible: I didn't want to generate confusion about what this specific component is about, nor dictate how to use it, but I get your point. I will add some basic layout in the example that clarifies what they user needs to do in order to see the skip link in action :)

src/components/SkipLink/SkipLink.tsx
lorumic: I see your point about customising the `href`, and it is a valid one indeed. I will keep `main-content` as a default though, since that's the standard that we use across projects.

If you use `AppMain` directly, you need to add `SkipLink` manually to your application. Currently, the only way to have it "for free" is by using `ApplicationLayout`, as it needs to be placed in a certain position of the layout to work. So if you don't use `ApplicationLayout` but only `AppMain`, you fall in the manual/customised case.

src/components/SkipLink/SkipLink.stories.tsx
mas-who: Yeah I agree with you, I guess my two comments is a little conflicting haha. I do think then having `href` as a prop for `SkipLink` makes sense to make it a bit more flexible. In terms of the documentation, I think maybe we can say it "could" be used with `AppMain` instead of it "must" if that makes sense?

src/components/SkipLink/SkipLink.tsx
mas-who: yeah having a default makes sense :+1:

mas-who (approved):
LGTM, thank you for the changes :)

PR #1157, opened 2024-10-30
chore(dependencies): Update all packages to latest version

mcslayer:
## Done

- Update all dependencies to the last version
Screenshot 2024-10-30 at 22 27 19
- After this update, I've got more lint errors. So I fixed them in components:
  - `ContextualMenu.stories.tsx` — problem with using nodes from the fill method without keys. Fixed by 3 separate paragraphs
  - `ModularTable.stories.tsx` — problem with using `div` in `p`. Fixed by adding a break to markdown for separate code block and paragraph
  - `SideNavigationItem.stories.tsx` — problem with using `li` in `li` and with icon's place. Fixed by using args from stories in items props
  - `Textarea.tsx` — problems with using `defaultValue` in spread `...props`, because after that we have controlled and uncontrolled textarea at the same time (value we get from state, defaultValue we get from `...props`). Fixed by using separate prop `defaultValue`.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: # .


bartaz:
Thank you, but we generally don't manage dependency updates manually. We rely on renovate to trigger update PRs for us.

But thanks for noticing that many of them were outdated. I poked on renovate to update them (https://github.com/canonical/react-components/pull/1159).

mcslayer:
:( okay

Anyway, I suggest taking a closer look at a few specific points:
- ContextualMenu.stories.tsx — problem with using nodes from the fill method without keys. Fixed by 3 separate paragraphs
- ModularTable.stories.tsx — problem with using div in p. Fixed by adding a break to markdown for separate code block and paragraph
- SideNavigationItem.stories.tsx — problem with using li in li and with icon's place. Fixed by using args from stories in items props
- Textarea.tsx — problems with using defaultValue in spread ...props, because after that we have controlled and uncontrolled textarea at the same time (value we get from state, defaultValue we get from ...props). Fixed by using separate prop defaultValue.

PR #1163, opened 2024-12-12
fix: avoid trapping tab key presses when the search and filter is closed

edlerd:
## Done

- avoid trapping tab key presses when the search and filter is closed

### Storybook

```shell
yarn start
```
### QA steps

- open the search and filter in storybook
- press tab and ensure you can tab over the closed search and filter with few tab key presses
- previously, all the hidden chips in the hidden search and filter panel would trap the tab key.


fasih-mehmood (approved):
LGTM, thank you!

PR #1168, opened 2025-01-06
feat(CustomSelect): Upstream CustomSelect component [WD-17720]

mas-who:
## Done

- Upstream `CustomSelect` component from lxd-ui

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check selecting an option
- Check search functionality
- Check height adjustment against viewport size
- Check keyboard navigation

### DIFF
--- a/src/components/CustomSelect/CustomSelect.stories.tsx
+++ b/src/components/CustomSelect/CustomSelect.stories.tsx
@@ -1,0 +1,143 @@
+import { Meta, StoryObj } from "@storybook/react/*";
+import CustomSelect from "./CustomSelect";
+import React, { ComponentProps, useState } from "react";
+import { CustomSelectOption } from ".";
+
+type StoryProps = ComponentProps<typeof CustomSelect>;
+
+const generateStandardOptions = (num: number): CustomSelectOption[] =>
+  Array(num)
+    .fill(null)
+    .map((_, i) => ({
+      value: `option-${i + 1}`,
+      label: `Option ${i + 1}`,
+      text: `Option ${i + 1}`,
+      disabled: false,
+    }));
+
+const generateCustomOptions = (): CustomSelectOption[] => {
+  return [
+    {
+      value: "smile",
+      label: <div>&#128512;</div>,
+      text: "Smile",
+      disabled: false,
+    },
+    {
+      value: "grin",
+      label: <div>&#128513;</div>,
+      text: "Grin",
+      disabled: false,
+    },
+    {
+      value: "cry",
+      label: <div>&#128557;</div>,
+      text: "Cry",
+      disabled: false,
+    },
+    {
+      value: "angry",
+      label: <div>&#128545;</div>,
+      text: "Angry",
+      disabled: false,
+    },
+    {
+      value: "sad",
+      label: <div>&#128546;</div>,
+      text: "Sad",
+      disabled: false,
+    },
+  ];
+};
+
+const Template = ({ ...props }: StoryProps) => {
+  const [selected, setSelected] = useState<string>(props.value || "");
+  return (
+    <CustomSelect
+      {...props}
+      value={selected}
+      onChange={(value) => setSelected(value)}
+    />
+  );
+};
+
+const meta: Meta<StoryProps> = {
+  component: CustomSelect,
+  render: Template,
+  tags: ["autodocs"],
+  args: {
+    name: "customSelect",
+    label: "Custom Select",
+    searchable: "auto",
+    initialPosition: "left",
+  },
+  argTypes: {
+    searchable: {
+      options: ["auto", "always", "never"],
+      control: {
+        type: "select",
+      },
+    },
+    initialPosition: {
+      options: ["left", "right"],
+      control: {
+        type: "select",
+      },
+    },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<StoryProps>;
+
+/**
+ * If `label` is of `string` type. You do not have to do anything extra to render it.
+ */
+export const StandardOptions: Story = {
+  args: {
+    options: generateStandardOptions(10),
+  },
+};
+
+/**
+ * If `label` is of `ReactNode` type. You can render custom content.
+ * In this case, the `text` property for each option is required and is used for display in the toggle, search and sort functionalities.
+ */
+export const CustomOptions: Story = {
+  args: {
+    options: generateCustomOptions(),
+  },
+};
+
+/**
+ * For each option, if `disable` is set to `true`, the option will be disabled.
+ */
+export const DisabledOptions: Story = {
+  args: {
+    options: generateStandardOptions(5).map((option, i) => ({
+      ...option,
+      disabled: i % 2 === 0,
+    })),
+  },
+};
+
+/**
+ * Search is enabled by default when there are 5 or more options.
+ */
+export const AutoSearchable: Story = {
+  args: {
+    options: generateStandardOptions(5),
+    searchable: "auto",
+  },
+};
+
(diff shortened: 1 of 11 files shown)


src/components/CustomSelect/CustomSelect.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+ * In this case, the `text` property for each option is required and is used for display in the toggle, search and sort functionalities.
+ */
+export const CustomOptions: Story = {
+  args: {
+    options: generateCustomOptions(),
+  },
+};
+
+/**
+ * For each option, if `disable` is set to `true`, the option will be disabled.
```
fasih-mehmood, line 114:
```suggestion
 * For each option, if `disabled` is set to `true`, the option will be disabled.
```

src/components/CustomSelect/CustomSelect.tsx
```diff
@@ -0,0 +1,203 @@
...
+    wrapperClassName?: ClassName;
+    // The styling for the select toggle button
+    toggleClassName?: ClassName;
+    // The styling for the select dropdown
+    dropdownClassName?: string;
+    // Whether the select is searchable. Option "auto" is the default, the select will be searchable if it has 5 or more options.
+    searchable?: "auto" | "always" | "never";
+    // Whether to focus on the element on initial render.
+    takeFocus?: boolean;
+    // Additional component to display above the dropdwon list.
```
fasih-mehmood, line 49:
```suggestion
    // Additional component to display above the dropdown list.
```

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
```diff
@@ -0,0 +1,374 @@
...
+
+export const sortOptions = (
+  a: CustomSelectOption,
+  b: CustomSelectOption,
+): number => {
+  // sort options alphabetically
+  const textA = getOptionText(a) || a.value;
+  const textB = getOptionText(b) || b.value;
+  return textA.localeCompare(textB);
+};
```
fasih-mehmood, line 166: I don't see this being used anywhere, can we remove it if it is not being used?

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
```diff
@@ -0,0 +1,374 @@
...
+      onClose();
+    }
+  };
+
+  const handleSearch = (value: string) => {
+    setSearch(value.toLowerCase());
+    // reset selected index when search text changes
+    setSelectedIndex(0);
+    optionsRef.current = [];
+  };
```
fasih-mehmood, line 289: How much work will it be to take an exposed `onSearch` prop and call it here with the value every time the search changes?

```suggestion
  const handleSearch = (value: string) => {
    setSearch(value.toLowerCase());
    
    if (onSearch) {
      onSearch(value);
      return;
    }
    
    // reset selected index when search text changes
    setSelectedIndex(0);
    optionsRef.current = [];
  };
```

This could be useful if we want to update the list of options asynchronously instead of simply filtering the existing options based on the search value. We could use it alongside a `searchExternallyControlled` or similar prop.

fasih-mehmood (approved):
Great work @mas-who, left a couple of comments. This looks and works great, thank you!

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
mas-who: good catch, I mostly copied over the code from lxd-ui, this function is used by a parent component there to sort the options. I think it makes sense to still keep it for `CustomSelect` and have it as a default sort. Then we can expose a prop `sortOptions` so that a parent component can implement it's own sort if desired. wdyt?

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
mas-who: Good suggestion, it's not that much effort and even so, if it's an good idea for improvement we should implement it :smile: will address this

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
fasih-mehmood: I don't think we should sort the options at all. It would be better to leave it to the dev to sort the options before passing them into this component if they want to. There can be cases where I, as a dev, will not want any sorting at all. We can still keep it as an option if you want.

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
fasih-mehmood: Awesome! Don't mind the return statement in the suggestion. I thought the "setSelectedIndex" was for updating the selected option but this is only for updating the keyboard navigation.

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
mas-who: You make a valid point, I am more thinking about the majority use case where this component would be used with simple text label options in which case I expect alphabetical sort would be most likely. Looking at the existing `Select` component, I see that sorting is also not implemented from within the component, I think let's go with your suggestion and remove this function.

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
mas-who: sure, I think the ordering of the code will need to shuffle a bit in the suggestion. `setSelectedIndex` and the reset for `optionsRef` should execute before the return, and `setSearch` should not execute if `onSearch` is defined since we do not want internal filtering to happen at all, the filtering will be delegated to the parent and passed down via the `options` prop.

Additionally, we can also expose `setSelectedIndex` via `onSearch` for the parent component to decide which option to highlight upon search. Do you think this is a plausible use case?

fasih-mehmood: We don't need to expose `setSelectedIndex`. If the options are externally controlled through search, they would most likely change every time the user searches something. if we just keep it as `setSelectedIndex(0)`, it will naturally highlight the top option until there is only one option.

fasih-mehmood: Yes, `setActiveOptionIndex` sounds good, or maybe we can do `setHighlightedOptionIndex`, since active can still mean selected. wdyt?

mas-who: Yeah initially thought about `setHighlightedOptionIndex` but felt it was a bit verbose. Happy to use that instead since it seems like a more popular choice :+1:

mas-who: I've made the adjustment, please have a look.

fasih-mehmood (approved):

src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx
fasih-mehmood: Looks great, thank you!

PR #1169, opened 2025-01-13
feat(MultiSelect): add scrollOverflow prop

yurii-vasyliev:
## Done

When used in certain cases with long lists, it would be useful to set the `scrollOverflow`.
I'd think we need this prop to be `true` all the time, but for the sake of compatibility with other projects, let's keep it false by default as it is right now.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: # .

### DIFF
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -31,6 +31,7 @@ export type MultiSelectProps = {
   dropdownFooter?: ReactNode;
   showDropdownFooter?: boolean;
   variant?: "condensed" | "search";
+  scrollOverflow?: boolean;
 };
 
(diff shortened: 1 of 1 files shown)


jmuzina:
When I set scrollOverflow to true, scroll down, and select an item, the list is scrolled back to the top. Is that intended behavior?

https://github.com/user-attachments/assets/67dff634-8584-4f0f-98c4-3cebc5fb3b8b

yurii-vasyliev:
@jmuzina , I don't think so, it seems like a bug. However, I think it's outside of the scope of this PR.

jmuzina (approved):

Issue #1170, opened 2025-01-16
Percy chunk loading error


mas-who:
When running PR tests, it seems like the percy job fails intermittently with chunk loading errors like the one show below. Rerunning the percy tests may result in the same error on different components. We should investigate and resolve this to prevent flaky tests.

![Image](https://github.com/user-attachments/assets/8a3a5f57-880f-444d-8f9b-8be9113159d5)

PR #1171, opened 2025-01-17
fix: reduce percy snapshot concurrency to resolve random test failures

mas-who:
## Done

- reduced percy snapshot concurrency, this should fix random test failures in percy
- based on discussion from this [issue](https://github.com/percy/percy-storybook/issues/1008)

## QA

- Ensure CI passes

## Fixes

Fixes: #1170

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -3,6 +3,7 @@ snapshot:
   widths: [1280]
   enable-javascript: true
 discovery:
+  concurrency: 2
   network-idle-timeout: 700
   retry: true
   launch-options:


.percy.yml
```diff
@@ -3,6 +3,7 @@ snapshot:
   widths: [1280]
   enable-javascript: true
 discovery:
+  concurrency: 2
```
jmuzina, line 6: It would be good to leave a comment here that links to the workaround we found or just explains why we are setting it to 2

jmuzina (commented):
Thanks for this! One small comment about documenting the fix.

.percy.yml
mas-who: thanks that's a good suggestion, I added a comment :)

jmuzina:
Changes from this were rolled into #1159

jmuzina (approved):

PR #1174, opened 2025-01-29
feat(deps): incremental major dependencies update

mas-who:
## Done

This PR is related to #1167 . It was a bit confusing for me to try get everything working in the renovate PR, so I decided to incrementally address updates in this branch.

- updated concurrently major version
- updated cypress major version
- updated eslint and stylelint major versions
- updated react to v19
- updated ubuntu to 24.04 in gh action
- removed jackspeak and strip-ansi resolutions

## Jira
- WD-18051

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Ensure CI passes
- Click around in story book and do spot checks for any regressions


package.json
```diff
@@ -112,18 +112,16 @@
     "react-useportal": "1.0.19"
   },
   "resolutions": {
-    "@types/react": "18.3.18",
-    "@types/react-dom": "18.3.5",
-    "postcss": "^8.3.11",
-    "jackspeak": "2.1.1",
-    "strip-ansi": "6.0.1"
+    "@types/react": "19.0.8",
```
bartaz, line 115: Let's try to remove those from resolutions and see what happens.

mas-who: removing the react related resolutions works, I'm testing to see if removing `postcss` resolution causes any issues.

mas-who: All resolutions have been removed, and CI is passing. Browsing through Storybook on the demo server didn’t reveal any issues. I suggest proceeding with the removal unless another reviewer recommends otherwise :thinking:

package.json
```diff
@@ -101,8 +101,8 @@
   "dependencies": {
     "@types/jest": "29.5.14",
     "@types/node": "20.17.13",
-    "@types/react": "18.3.18",
-    "@types/react-dom": "18.3.5",
+    "@types/react": "19.0.8",
+    "@types/react-dom": "19.0.3",
```
edlerd, line 105: Do we need the two versions here as well? I am not sure if this will block downstream projects to use React 18 and its types with no issues.

src/components/SearchBox/SearchBox.tsx
```diff
@@ -52,10 +52,6 @@ export type Props = PropsWithSpread<
      * Whether the search input should receive focus after pressing the reset button
      */
     shouldRefocusAfterReset?: boolean;
-    /**
-     * A ref that is passed to the input element.
-     */
-    ref?: string;
```
edlerd, line 58: I guess forwardRef injects this prop automatically? Probably still okay to mention it here, or why was it removed?

src/components/Stepper/Step/Step.scss
```diff
@@ -90,7 +90,7 @@
     padding-bottom: 0;
     padding-top: 0;
   }
-
+  /* stylelint-disable-next-line no-descending-specificity */
```
edlerd, line 93: Can we move the file contents around to reach ascending specificity? That way we would resolve the error.

edlerd (approved):
LGTM, some questions below.

package.json
mas-who: I think this is needed as these dependencies indicates what the current project must use. The downstream projects should be fine with the peer dependencies. This was also done in the [commit](https://github.com/canonical/react-components/commit/e1333ff5f7f0fee4a6530c8838dc52b4bc8f46ae) where the types were updated from v17 to v18.

src/components/SearchBox/SearchBox.tsx
mas-who: `ref` is reserved and shouldn't be part of the props interface for the component, that's why I removed it. Also it's weird that the `ref` prop (which is a string)  previously was overriding the actual `ref` prop for the `input` element down below.

src/components/Stepper/Step/Step.scss
bartaz: Does it throw an error or just a warning? We have it configured as warnings in Vanilla (https://github.com/canonical/vanilla-framework/blob/main/.stylelintrc.json#L36)

The issue is with SCSS nesting, and would be hard to fix without breaking things down into smaller chunks. Because we have:

1. `.stepper-horizontal .step`
2. `.stepper-horizontal :first-child .step`
3. `.stepper-vertical .step`
4. `.stepper-vertical :first-child .step`

So indeed, specificity of 2) `.stepper-horizontal :first-child .step` is higher than  of 3) `.stepper-vertical .step`. But to fix it we would have to move all the longer selectors (with `:first-child`) below shorter selectors, breaking the grouping created by nesting.

That's why we treat it as a warning only in Vanilla, because in many cases fixing it would lead to less readable code.

I'd suggest doing the same here (configure it as a warning). If it's already a warning, just ignore it (no need to disable the linter for it).

mas-who: Thanks for the tip @bartaz yes it's already configured as a warning so no need for this change. I will revert it :+1:

package.json
edlerd: I was concerned this might trickle down to the library that is built. And that might make problems for downstream applications when staying with react 18.

Back then when going from 17 to 18 we did not try to be backwards compatible and let clients stay on 17.

bartaz: I don't think we can have 2 versions here as that's this project dependencies, so they need to be specified.

We can have 2 supported versions in peer dependencies.

I don't remember it being a problem when migrating to 18. We updated react-component dependencies (and types) to 18, and allowed both 17 and 18 in peer dependencies. So hopefully this should work here as well.

Would it be possible to test it in one of the React projects? to switch react-components dependency to this branch and verify if it works?

I think notation was something like that:
`"react-components": "mas-who/react-components#update-deps-major"

mas-who: So having the peer dependency listed as `^18.0.0 || ^19.0.0` should allow downstream projects use v18 types without failing builds. The dependencies updated here is for this project only. Issues may arise when we are using new types introduced in v19 for react-components but a downstream project is still using v18 types as it satisfies the peer dependency. For the current PR, the only relevant typing change is replacing all occurrences of `JSX.Element` with `React.ReactElement` and the latter type exists prior to v19. So I believe nothing will break in downstream projects, and also I don't think we can remove these dependencies as it's needed for the current project.

mas-who: @bartaz yes I can test it with lxd-ui, I do remember previously I had some challenges to get local library referencing working but I'm sure I will figure it out :+1:

bartaz: @mas-who This is how I do it for Vanilla when I need it. Should hopefully work the same with react-components: https://github.com/canonical/snapcraft.io/pull/5000/commits/aa9f053b41e667bb7ed97192d9fa12e9ce310bfe#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R105

mas-who: thanks @bartaz I got it working, it was a good idea to test locally first because `React.ReactElement` from v19 types is stricter than the `JSX.Element` type in v18. This means downstream projects with v18 types will see type mismatch issues. Looking into a workaround for this

mas-who: So in react 19 the `JSX` namespace has been removed ([ref](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript)), and the suggestion is to replace `JSX` with `React.JSX` however it is not compatible with react 18 types and will cause downstream project to break. So we could keep the react types on v18 in this PR, I checked that works even if react is upgraded to v19. However, I think that adds little value because eventually a downstream project will need to upgrade react and its types. If react-components or a downstream project upgrade their react types from v18 to v19, the downstream project build will fail. Maybe we could release this PR as a breaking change and downstream projects would have to upgrade to react v19 to resolve the issues, wdyt @edlerd @bartaz ?

edlerd: Thanks for the analysis and finding a compatible way. As discussed, this sounds like a good solution. Please communicate this through the react channel on mattermost, so other projects are not surprised.

With that, this should be good to merge from my side.

Issue #1175, opened 2025-01-29
Link button example in Storybook uses `base` appearance instead of the `link` one


lorumic:
The "[Link button](https://canonical.github.io/react-components/?path=/story/components-button--link)" example in Storybook uses the `base` appearance instead of the `link` one. I believe this is accidental, e.g. as a result of an unfortunate copy-paste. Code reference for the link appearance: https://github.com/canonical/react-components/blob/main/src/components/Button/Button.tsx#L17

bartaz:
Yes, seems like an error. Should be easy to update.

PR #1176, opened 2025-01-30
fix(button): link button example in storybook was using base appearance

lorumic:
## Done

- Updated Storybook example for the "[Link button](https://react-components-1176.demos.haus/?path=/story/components-button--link)"
- Added the "Button as link" story at the bottom
- Added descriptions for all the button stories

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check the updated Storybook example for the "[Link button](https://react-components-1176.demos.haus/?path=/story/components-button--link)" - it should now look like a link, instead of a base button.
- Check that the storybook page for the `Button` component looks as expected

### Percy steps

- Confirm that the only visual changes are in the "Button" page of the Storybook.

## Fixes

Fixes: #1175

### DIFF
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -1,4 +1,4 @@
-import React, { HTMLProps } from "react";
+import React from "react";
 import { Meta, StoryObj } from "@storybook/react";
 
 import Tooltip from "components/Tooltip";
@@ -75,20 +75,16 @@ export const Link = {
 
   args: {
     children: "Link button",
-    appearance: "base",
-    element: "a",
-    href: "#test",
+    appearance: "link",
   },
 };
 
-export const LinkDisabled: Story<HTMLProps<HTMLAnchorElement>> = {
+export const LinkDisabled = {
   name: "Link disabled",
 
   args: {
     children: "Link button disabled",
-    appearance: "base",
-    element: "a",
-    href: "#test",
+    appearance: "link",
     disabled: true,
   },
 };
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { HTMLProps } from "react";
 import { Meta, StoryObj } from "@storybook/react";
 
 import Tooltip from "components/Tooltip";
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.stories.tsx
```diff
@@ -75,9 +75,7 @@ export const Link = {
 
   args: {
     children: "Link button",
-    appearance: "base",
-    element: "a",
-    href: "#test",
+    appearance: "link",
```
bartaz, line 78: Now looking at it, it got me thinking.

Maybe the intention was for the story to be about a link button (a link element), not about appearance.

So the existing example was likely correct. It was creating a button using `<a>` element instead of `button`.

But it is true, that because we have a "link" appearance it's quite confusing.


So maybe the fix is actually to add a new story about "link" appearance button, not to update the existing one.

It would be good to also add some kind of descriptions to the stories, so it's not so confusing what's the intention of the example.

lorumic: I think the existing one should be updated. All the examples around it are about the `appearance` property: default, base, positive, negative, brand - these are all possible appearance values, and the `link` appearance is the only one that is not represented by any example. The example with the `<a>` element was in the middle of them, so it looks out of place to me. Not sure why it was even added in the first place.

I can add back the previous example, with the `<a>` element, at the bottom of the list, if we really need to show such example, but for coherence with the examples around it, it should not be called "Link button". I'm open to suggestions for a better name, if we really cannot drop it altogether (which I still think would be the best choice here).

bartaz: Makes sense.

As for the `<a>` example I'd suggest to keep it (as you mentioned, makes sense to move it down). This is the only place where we seem to document the usage of `element` and `href` props. Maybe "Button as a link" for a title would be clear enough?

lorumic: Thanks for the suggestion. I think it's a good title. I've added the example back, plus the descriptions for all stories for `Button` [here](https://github.com/canonical/react-components/pull/1176/commits/23b9d1eab780c1b4df67436e8b11475dee7d1631).

bartaz (approved):
LGTM, good catch and thanks for quick fix!

PR #1183, opened 2025-02-18
refactor(Notification): Add type="button" and allow passing in ReactNode to Notification Actions

fasih-mehmood:
## Done

- Add `type="button"` prop to notification action buttons and dismiss button
- Allow passing in `ReactNode` to notification actions

### Percy steps

- No visual changes expected.

### DIFF
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -28,6 +28,7 @@ export const DefaultTitles = {
 type NotificationAction = {
   label: string;
   onClick: () => void;
+  disabled?: boolean;
 };
 
 /**
@@ -177,6 +178,7 @@ const Notification = ({
         <p className="p-notification__message">{children}</p>
         {onDismiss && (
           <button
+            type="button"
             className="p-notification__close"
             data-testid="notification-close-button"
             onClick={onDismiss}
@@ -199,11 +201,13 @@ const Notification = ({
             <div className="p-notification__actions">
               {actions.map((action, i) => (
                 <Button
+                  type="button"
                   appearance={ButtonAppearance.LINK}
                   className="p-notification__action"
                   data-testid="notification-action"
                   key={`${action.label}-${i}`}
                   onClick={action.onClick}
+                  disabled={action.disabled}
                 >
                   {action.label}
                 </Button>
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -1,5 +1,11 @@
 import classNames from "classnames";
-import React, { ElementType, Fragment, useEffect, useRef } from "react";
+import React, {
+  ElementType,
+  Fragment,
+  isValidElement,
+  useEffect,
+  useRef,
+} from "react";
 import type { HTMLProps, ReactNode } from "react";
 
 import Button, { ButtonAppearance } from "../Button";
@@ -211,7 +217,7 @@ const Notification = ({
                       {action.label}
                     </Button>
                   ) : (
-                    action
+                    isValidElement(action) && action
                   )}
                 </Fragment>
               ))}
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -1,15 +1,9 @@
 import classNames from "classnames";
-import React, {
-  ElementType,
-  Fragment,
-  isValidElement,
-  useEffect,
-  useRef,
-} from "react";
+import React, { ElementType, useEffect, useRef } from "react";
 import type { HTMLProps, ReactNode } from "react";
 
 import Button, { ButtonAppearance } from "../Button";
-import { IS_DEV } from "../../utils";
+import { IS_DEV, isReactNode } from "../../utils";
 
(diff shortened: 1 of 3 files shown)


src/components/Notification/Notification.tsx
```diff
@@ -28,6 +28,7 @@ export const DefaultTitles = {
 type NotificationAction = {
   label: string;
   onClick: () => void;
+  disabled?: boolean;
 };
```
huwshimi, line 32: Another option would be to allow passing any button props to these actions to make them more flexible.

```suggestion
type NotificationAction = {
  label: string;
  onClick: () => void;
} & Partial<ButtonProps>;
```

huwshimi (approved):
This looks good, I've made one suggestion, but not required.

src/components/Notification/Notification.tsx
fasih-mehmood: This was actually my initial plan, to replace the underlying button with an **ActionButton** and extend the Props. For my use case I needed access to both the `disabled` and `loading` props on an **ActionButton**. But David said that he wasn't sure about using an **ActionButton**. I can extend the **ButtonProps** and leave the underlying **Button** as is instead of using an **ActionButton**.

image
_Note: onClick would just be removed from the **Omit** group and from the **NotificationAction** itself since Button already has the prop. And we could wrap the Omit with a Partial_

What would you suggest?


Reference: https://chat.canonical.com/canonical/pl/ax8fwwcrz3fdupjaag6gt4fimh

huwshimi: Another option would be to allow `ReactNode` as well as `NotificationAction` so that you can pass an `ActionButton` or any other component:

```
actions?: (NotificationAction | ReactNode)[];
```

And then you'd also need to update where it does `actions.map` to only map to a button when it's a `NotificationAction`.

fasih-mehmood: That would be even better! Do we need an array of `ReactNode` though? Can we just do this and avoid mapping altogether if it's a ReactNode?

```ts
actions?: NotificationAction[] | ReactNode;
```

Also, I believe we would still keep the `type="button"` change and the `disabled` prop can be removed

huwshimi: > That would be even better! Do we need an array of `ReactNode` though? Can we just do this and avoid mapping altogether if it's a ReactNode?
> 
> ```ts
> actions?: NotificationAction[] | ReactNode;
> ```

True, though the benefit of allowing a mix will be that if you only want to use ReactNode for some actions and standard NotificationActions for others you won't have to pass in a button that matches what the component defines with the correct classes etc (for the NotificationActions).

> 
> Also, I believe we would still keep the `type="button"` change and the `disabled` prop can be removed

Yep that sounds good.

fasih-mehmood: Ah, fair point. I'll make the changes then. Thank you!

fasih-mehmood: Done

src/components/Notification/Notification.tsx
```diff
@@ -198,15 +199,21 @@ const Notification = ({
...
-                  appearance={ButtonAppearance.LINK}
-                  className="p-notification__action"
-                  data-testid="notification-action"
-                  key={`${action.label}-${i}`}
-                  onClick={action.onClick}
-                >
-                  {action.label}
-                </Button>
+                <Fragment key={i}>
+                  {typeof action === "object" && "label" in action ? (
```
huwshimi, line 203: This will also need to check that `action` exists, as `typeof null === "object"` but would then fail when it tried to check `"label" in null`.
```suggestion
                  {action && typeof action === "object" && "label" in action ? (
```

huwshimi (approved):
Just one small change needed.

src/components/Notification/Notification.tsx
fasih-mehmood: Can you take another look? I had to use `isValidElement` to make it work

src/components/Notification/Notification.tsx
```diff
@@ -198,15 +205,21 @@ const Notification = ({
...
+                      appearance={ButtonAppearance.LINK}
+                      className="p-notification__action"
+                      data-testid="notification-action"
+                      onClick={action.onClick}
+                    >
+                      {action.label}
+                    </Button>
+                  ) : (
+                    isValidElement(action) && action
+                  )}
```
huwshimi, line 221: I think you can swap this around to make it simpler if you use the `isReactNode` typeguard from src/utils.tsx. It'd probably also be a good idea to add a unit test for the ReactNode case.

```suggestion
                  {isReactNode(action) ? (
                    action
                  ) : (
                    <Button
                      type="button"
                      appearance={ButtonAppearance.LINK}
                      className="p-notification__action"
                      data-testid="notification-action"
                      onClick={action.onClick}
                    >
                      {action.label}
                    </Button>
                  )}
```

fasih-mehmood: Done, thanks!

PR #1187, opened 2025-03-20
feat(MultiSelect): Make sorting optional

ethanashaw:
## Done

- Add `isSortedAlphabetically?: boolean = true` to `MultiSelect`, which determines whether to sort the items by name.
- Add `hasSelectedItemsFirst?: boolean = true` to `MultiSelect`, which determines whether to sort selected items first.

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Ensure that `MultiSelect`s without the new fields have not changed their sorting.
- Ensure that `MultiSelect`s with `isSortedAlphabetically={false}` do not sort their items alphabetically.
- Ensure that `MultiSelect`s with `hasSelectedItemsFirst={false}` do not sort selected items first.

|image|
|-|

### Percy steps

- New story

### DIFF
--- a/src/components/MultiSelect/MultiSelect.test.tsx
+++ b/src/components/MultiSelect/MultiSelect.test.tsx
@@ -278,3 +278,46 @@ it("opens and closes the dropdown on click", async () => {
   await userEvent.click(screen.getByRole("combobox"));
   expect(screen.queryByRole("listbox")).not.toBeInTheDocument();
 });
+
+it("can render without sorting alphabetically", async () => {
+  const itemsUnsorted = [
+    { label: "item B", value: 2 },
+    { label: "item A", value: 1 },
+    { label: "other B", value: 3 },
+    { label: "other A", value: 4 },
+  ];
+
+  render(<MultiSelect items={itemsUnsorted} isSortedAlphabetically={false} />);
+  await userEvent.click(screen.getByRole("combobox"));
+  const list = screen.getByRole("list");
+
+  const expectedLabels = itemsUnsorted.map((item) => item.label);
+  await waitFor(() =>
+    within(list)
+      .getAllByRole("listitem")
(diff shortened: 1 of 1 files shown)


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -32,6 +32,8 @@ export type MultiSelectProps = {
   showDropdownFooter?: boolean;
   variant?: "condensed" | "search";
   scrollOverflow?: boolean;
+  sortedAlphabetically?: boolean;
+  sortedBySelection?: boolean;
```
edlerd: I think naming could be a bit clearer. For booleans an is or has prefix seems a good idea. How about `hasSelectedOptionsFirst`?

ethanashaw: Sounds good, I went with `hasSelectedOptionsFirst` and `isSortedAlphabetically`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -50,7 +52,8 @@ type MultiSelectDropdownProps = {
   onSelectItem?: (item: MultiSelectItem) => void;
   footer?: ReactNode;
   groupFn?: GroupFn;
-  sortFn?: SortFn;
+  sortFn: SortFn;
```
edlerd: This is exported, so it might be expected to be optional in downstream use cases. I also don't fully understand why this was changed to be required. Can this stay unchanged?

ethanashaw: Thanks for pointing this out, I didn't realize it was exported. I also reverted the default `sortFn` for `MultiSelectDropdown` to `sortAlphabetically`.

PR #1189, opened 2025-03-27
fix(MultiSelect): Add z-index to MultiSelect menu

fasih-mehmood:
## Done

- Set z-index on multi select menu so it renders above modal backdrop

### Drive-by:

- Fix typo in ModularTable documentation

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### Percy steps

- No visual changes expected

### DIFF
--- a/src/components/MultiSelect/MultiSelect.scss
+++ b/src/components/MultiSelect/MultiSelect.scss
@@ -15,7 +15,7 @@ $dropdown-max-height: 20rem;
   width: 100%;
 
   // Ensure the dropdown is above the modal backdrop
-  z-index: var(--menu-z-index, 151); 
+  z-index: var(--menu-z-index, 151);
 }
 
 .multi-select .p-form-validation__message {
--- a/src/components/MultiSelect/MultiSelect.scss
+++ b/src/components/MultiSelect/MultiSelect.scss
@@ -13,6 +13,9 @@ $dropdown-max-height: 20rem;
   margin-bottom: $input-margin-bottom;
   position: relative;
   width: 100%;
+
+  // Ensure the dropdown is above the modal backdrop
+  z-index: var(--menu-z-index, 151); 
 }
 
 .multi-select .p-form-validation__message {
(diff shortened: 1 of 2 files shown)


src/components/MultiSelect/MultiSelect.scss
```diff
@@ -13,6 +13,9 @@ $dropdown-max-height: 20rem;
   margin-bottom: $input-margin-bottom;
   position: relative;
   width: 100%;
+
+  // Ensure the dropdown is above the modal backdrop
+  z-index: var(--menu-z-index, 151);
```
edlerd, line 18: I missed the details on this yesterday in our react guild. We had the same issue with the `CustomSelect` component. It uses the `ContextualMenu` to open the dropdown, which internally uses `Portal`.

While the z-index introduced here fixes your issue at hand, this approach is still open to similar bugs. I.e. when using the component inside a wrapper that has overflow: hidden and with limited size of the wrapper. A reliable approach to overcome this is to use portal instead. That renders the dropdown on the root level of the dom tree, and we don't need the z-index anymore as a workaround.

fasih-mehmood: MultiSelect also uses `ContextualMenu` under the hood. Using `Portal` didn't fix it in our case because we render our Modals using `Portal` so we end up with 2 different portals and then the higher z-index wins which in our case is the Modal.

edlerd: Ah right, got it. I wasn't aware of the MultiSelect using the contextual menu with portal as well. I remember we had the same issue with the custom select and did some more research. Found [this approach](https://github.com/canonical/lxd-ui/pull/1002/files#diff-d2ebe81e706e39cec538cf3e23bd43d7a68c282306821cb8f95e5ed31ede02f3) to avoid static z indexes. The problem with those is they are highly dependent on each other. Maybe the other approach with a relative z index is more safe, what do you think?

fasih-mehmood: This looks great! Do you think we should use it directly in `ContextualMenu`? That way we won't have to handle it separately for MultiSelect.

edlerd: You are right, that is the better solution to raise it to the contextual menu.

fasih-mehmood:
Closed in favor of #1192

PR #1195, opened 2025-04-16
fix: dynamically set label to match id

anusha-c18:
## Done

- Fixes an accessibility issue in the `SearchBox` component by dynamically setting the label’s `htmlFor` to match the input id, when provided. 

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

### DIFF
--- a/src/components/SearchBox/SearchBox.tsx
+++ b/src/components/SearchBox/SearchBox.tsx
@@ -68,6 +68,8 @@ export type Props = PropsWithSpread<
 const SearchBox = React.forwardRef<HTMLInputElement, Props>(
   (
     {
+      name = "search",
+      id = "search",
       autocomplete = "on",
       className,
       disabled,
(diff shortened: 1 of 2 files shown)


src/components/SearchBox/SearchBox.tsx
```diff
@@ -108,14 +108,14 @@ const SearchBox = React.forwardRef<HTMLInputElement, Props>(
 
     return (
       <div className={classNames("p-search-box", className)}>
-        <label className="u-off-screen" htmlFor="search">
+        <label className="u-off-screen" htmlFor={props.id || "search"}>
```
jmuzina, line 111: If you're going to use a default for the `id` prop, it would be simpler and less repetitive to destructure it at the component signature level, and give it a default there - that way you don't have to duplicate the fallback to `"search"`.

```tsx
const SearchBox = React.forwardRef<HTMLInputElement, Props>(
  (
    {
      /// other destructured props....
      id="search"
      ...props
    }: Props,
    // ....
    <input id={id}>
```

Possibly out of scope of this PR: an even cleaner approach would be to default to an ID generated with `useId()` if no `id` prop is passed, instead of relying on the constant `"search"`. This would mitigate the possibility of multiple search boxes having the same ID, but I'm not sure if it's necessary as we don't often have multiple search boxes on the page at once.

anusha-c18: Thanks for the suggestion!
For now, I've given both the name and id defaults at the signature level [here](https://github.com/canonical/react-components/compare/88e0a03107f8d85dab2acb0408f932a7771c1e0d..f658806c771a6c3323dbeef0f25466b27b7f7154)

jmuzina (approved):
Thanks for the a11y fix :)

PR #1202, opened 2025-05-06
feat: add help text to switch component

anusha-c18:
## Done

- Add optional help text to switch component

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: #[WD-13578](https://warthogs.atlassian.net/browse/WD-13578)

[WD-13578]: https://warthogs.atlassian.net/browse/WD-13578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Stepper/Step/Step.scss
+++ b/src/components/Stepper/Step/Step.scss
@@ -85,7 +85,7 @@
   }
 }
 
-.stepper-vertical {
+.stepper-vertical:not(.stepper-horizontal) {
   .p-list__item {
     padding-bottom: 0;
     padding-top: 0;
--- a/src/components/Switch/Switch.tsx
+++ b/src/components/Switch/Switch.tsx
@@ -1,7 +1,8 @@
 import React from "react";
+import classNames from "classnames";
 import type { HTMLProps, ReactNode } from "react";
-
 import type { PropsWithSpread } from "types";
+import "./Switch.scss";
 
 export type Props = PropsWithSpread<
   {
@@ -13,6 +14,14 @@ export type Props = PropsWithSpread<
      * Whether the switch is disabled or not
      */
     disabled?: boolean;
+    /**
+     * Help text to show below the field.
+     */
+    help?: ReactNode;
+    /**
+     * Optional class(es) to pass to the help text element.
+     */
+    helpClassName?: string;
   },
   HTMLProps<HTMLInputElement>
 >;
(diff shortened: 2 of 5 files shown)


lorumic:
Looks like this one requires design input. Personally - and also judging from the screenshot added in the ticket, which unfortunately we don't have anymore as that part of the Anbox UI has changed - I think that the help text should be aligned with the main label, which does not currently seem the case:

image

anusha-c18:
The design team confirmed the same! Addressed the changes ac06596 @lorumic

src/components/Stepper/Step/Step.scss
```diff
@@ -91,6 +91,7 @@
     padding-top: 0;
   }
 
+  // stylelint-disable-next-line no-descending-specificity
```
lorumic, line 94: The reason why this is complaining here is that it assumes that one can have both the `stepper-vertical` and `stepper-horizontal` classes applied to the same element. Nothing would prevent that to happen currently, and in that case, the descending specificity issue would arise.
Since we know that is not the case - it's either one or the other - we can solve the warning more elegantly by changing line 88 to:
```
.stepper-vertical:not(.stepper-horizontal) {
```
And this will make the linter happy, without having to disable it.

src/components/Switch/Switch.tsx
```diff
@@ -23,20 +32,36 @@ export type Props = PropsWithSpread<
...
+        <span className="p-switch__label">{label}</span>
+      </label>
+      {help && (
+        <p
+          className={classNames(
+            "p-form-help-text",
+            "switch-help-text",
+            helpClassName,
+          )}
+          id={inputProps["aria-describedby"]}
```
lorumic, line 59: Do we need this? I think we can do without.
```suggestion
```

anusha-c18: This makes screen readers announce the help text when focusing on the switch input so it's better in terms of accessibility

lorumic: Ok, what about using the same approach that the `Input` component uses, i.e. using `useId`? https://github.com/canonical/react-components/blob/42c47733d1a7490c559cd7d26bc4c30807c40b8f/src/components/Input/Input.tsx#L104

src/components/Stepper/Step/Step.scss
anusha-c18: Awesome, thanks for the suggestion! Addressed the [changes](https://github.com/canonical/react-components/compare/49d4ed1a2d3ca9982371f1f9719330b9f0b99cba..721a126aeba642671a1b9b4146cf60e697e798fe)

lorumic (approved):
LGTM, thanks for adding this :rocket:

PR #1203, opened 2025-05-07
fix: remove information icon customization

anusha-c18:
## Done

- Remove custom info icon

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

### DIFF
--- a/src/components/Notifications/ToastNotification/Toast.scss
+++ b/src/components/Notifications/ToastNotification/Toast.scss
@@ -1,5 +1,3 @@
-@import "vanilla-framework/scss/settings";
-
 .toast-animate {
   position: relative;
   z-index: 101;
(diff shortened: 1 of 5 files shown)


src/components/Notifications/ToastNotification/Toast.scss
```diff
@@ -1,5 +1,4 @@
 @import "vanilla-framework/scss/settings";
```
edlerd, line 1: This import should also be removed. I think we should not have any vanilla imports.

edlerd (commented):
Thanks for bringing this up so quickly! I think some other import should be removed additionally.

src/components/Notifications/ToastNotification/Toast.scss
anusha-c18: Had to use the values of the variables directly after I removed this import 
Addressed the changes in 14d690f

edlerd (approved):
Thx for the fixes, LGTM

PR #1206, opened 2025-06-03
fix: reset filters when all filtered notifications are closed

anusha-c18:
## Done

- Fix an edge case where filters weren't reset when the filtered notifications were individually dismissed 

Before:

https://github.com/user-attachments/assets/344612bd-7b24-4727-a704-43f1404e544f

After:

https://github.com/user-attachments/assets/8d0994ad-a54a-4844-b867-fd97bad85b6a

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- select a filter
- dismiss all the notifications in that filter by clicking the close (X) icon within each notification
- confirm that the filters are reset when all the filtered notifications are closed

### Percy steps

- No visual changes expected

### DIFF
--- a/src/components/Notifications/ToastNotification/ToastNotificationList.tsx
+++ b/src/components/Notifications/ToastNotification/ToastNotificationList.tsx
@@ -8,7 +8,7 @@ import {
   ToastNotificationType,
 } from "./ToastNotificationProvider";
 import type { FC } from "react";
-import { useLayoutEffect, useRef, useState } from "react";
+import { useLayoutEffect, useRef, useState, useEffect } from "react";
 import { createPortal } from "react-dom";
 import Animate from "./Animate";
 import { usePrefersReducedMotion } from "../../../hooks";
@@ -200,6 +200,13 @@ const ToastNotificationList: FC<Props> = ({
   const filteredNotifications = hasFilters
     ? notifications.filter((notification) => filters.has(notification.type))
     : notifications;
+
+  useEffect(() => {
+    if (hasFilters && filteredNotifications.length === 0) {
+      // if there are no filtered notifications, reset the filters
+      setFilters(new Set());
+    }
+  }, [hasFilters, filteredNotifications]);
 
   // Don't assign alert role for notifications when expanded since we don't want
   // screen readers to announce every existing notification


src/components/Notifications/ToastNotification/ToastNotificationList.tsx
```diff
@@ -201,6 +201,14 @@ const ToastNotificationList: FC<Props> = ({
     ? notifications.filter((notification) => filters.has(notification.type))
     : notifications;
 
+  useEffect(() => {
+    if (hasFilters && filteredNotifications.length === 0) {
+      // if there are no filtered notifications, reset the filters
+      setFilters(new Set());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [filteredNotifications]);
```
lorumic: Not saying that this is better, but we could avoid silencing eslint (which I personally think we should not fall into the habit of doing) by removing the `hasFilters` check:
```suggestion
  useEffect(() => {
    if (filteredNotifications.length === 0) {
      // if there are no filtered notifications, reset the filters
      setFilters(new Set());
    }
  }, [filteredNotifications]);
```
~~I _think_ this should not cause an additional re-render when the Set is already empty, but I haven't tested it.~~ Nevermind, it _will_ trigger an additional re-render, because `Object.is(new Set(), new Set())` returns `false`, so the question is more: "do we care about that additional re-render?"

The alternative would be to have the default state set to `null` instead of the empty Set, so that we could do `setFilters(null)` here, and then React will recognise that the state is already `null` (`Object.is(null, null)` returns `true`), and so it won't re-render. However, this makes the state update much more cumbersome when you need to add your first filter, because you'd need to check if the `filters` state is `null`, and if it is initialise it like this: `setFilters([filter])`. So maybe it's not worth it, and keeping eslint silent is the best thing to do here.

Anyway, this is just a tiny non-blocking consideration, so I'll approve and leave the final decision up to you.

lorumic (approved):
A tiny, non-blocking idea.

src/components/Notifications/ToastNotification/ToastNotificationList.tsx
edlerd: In the current implementation, `hasFilters` is missing from the dependency list of the use effect.

edlerd (changes requested):
Thanks for finding this and picking it up to improve!

I think there is a dependency missing, see below.

src/components/Notifications/ToastNotification/ToastNotificationList.tsx
lorumic: I think the intention here is exactly to _not_ run this effect when `hasFilters` changes. Because if you're toggling the filter on and off (= this is when `hasFilters` changes) you will never have `filteredNotifications.length === 0` (otherwise you would not see the filter button at all). Also, the effect would be triggered twice when you manually dismiss the last filtered notification:
1. Because `filteredNotifications` changes
2. Because `hasFilters`, which is computed from `filteredNotifications`, changes consequently

So I think `hasFilters` should _not_ go in the dependency array.

edlerd: > I think the intention here is exactly to not run this effect when hasFilters changes. Because if you're toggling the filter on and off (= this is when hasFilters changes) you will never have filteredNotifications.length === 0 (otherwise you would not see the filter button at all).

without it we are prone to reading an outdated value.

lorumic: > without it we are prone to reading an outdated value.

I'm honestly not sure how that could happen, and QA confirms that there are no such issues.

Anyway, I have updated my comment above to explain better why we should not include `hasFilters` in the dependency array. I hope it's clearer now.

anusha-c18:
```
  // Only filter input data if there are filters present
  const filteredNotifications = hasFilters
    ? notifications.filter((notification) => filters.has(notification.type))
    : notifications;
```

I intentionally left out `hasFilters` from the dependencies. `filteredNotifications` already reflects the state of `hasFilters` and `notifications`, so including `hasFilters` separately in the dependency array is redundant. The effect only needs to run when the result of that filtering logic changes. Including `hasFilters` could cause unnecessary effect runs without changing the outcome.

edlerd: In the current form this is open for bugs where `hasFilters` is changed to `true` through a render, but the use effect still thinks it is `false`. You must define all reactive values inside of the dependency array of the use effect. See also https://react.dev/reference/react/useEffect#specifying-reactive-dependencies

edlerd: This bit explains it a bit more: https://react.dev/reference/react/useEffect#reading-the-latest-props-and-state-from-an-effect

lorumic: @anusha-c18 Let's maybe solve this easily by implementing my suggestion above, wdyt?

```suggestion
  useEffect(() => {
    if (filteredNotifications.length === 0) {
      // if there are no filtered notifications, reset the filters
      setFilters(new Set());
    }
  }, [filteredNotifications]);
```

anusha-c18: Thanks for the reference and suggestion! @edlerd @lorumic  
Since `filteredNotifications` defaults to all notifications when no filters are applied, removing the `hasFilters` check from the condition would result in both `ToastNotificationList` and `ToastNotificationProvider` trying to close the notification center, when all the notifications are dismissed. To ensure this only happens in the specific edge case when filters are applied but no matching notifications remain, I’ll keep `hasFilters` in the condition and include it in the dependency array.

lorumic (approved):

Issue #1109, opened 2024-07-25
Spinning animation in `ActionButton` ends too late


lorumic:
When using `ActionButton` with the `loading` and `disabled` props, the spinning animation ends later than when the variable regulating the loading behaviour is set to false:

```
            <ActionButton
              disabled={isLoading}
              loading={isLoading}
              className="u-no-margin--bottom"
              onClick={handleGenerate}
            >
              Request new URL
            </ActionButton>
```

![Peek 2024-07-25 20-30](https://github.com/user-attachments/assets/01ef5c81-7092-4b67-8abd-ac4544e01a3a)

As can be seen in the screen capture, the cursor gets switched back to the `pointer` value (= enabled state) while the spinning animation is still showing.

bartaz:
Not sure if that's the issue with the component itself, or the application, or React rendering.
All the button is doing is rendering itself based on props passed. So some more debugging may be needed to understand why they appear to change at different times.

PR #1207, opened 2025-06-06
fix: update duration after loading is false to calculate time spent f…

worapolw:
I found an issue from [issues-1109](https://github.com/canonical/react-components/issues/1109) which is a minor loading problem that occurs because of a wrong duration handling. So, I think this is an opportunity of me to fix this problem.

Please review my PR and don't hesitate to contact me if I did something wrong.

before:

https://github.com/user-attachments/assets/4bc5b298-8f41-4d11-a656-0744027d5059

after:

https://github.com/user-attachments/assets/1fa59076-74f3-49bf-86e1-b423b9baa212

## Done

- fix add time tracking to once loading is true and then do subtraction when loading is false to prevent unnecessary LOADER_MIN_DURATION in setTimeout before setShowLoader to false

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

-  try using ActionButton with a duration which is higher than LOADER_MIN_DURATION  and track of state of loading, it will show that state of user already done loading but in UI seem a bit slower

### Percy steps

- No visual changes expected

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,8 @@
 
 ## QA
 
+Pinging @canonical/react-library-maintainers for a review.
+
 ### Storybook
 
 To see rendered examples of all react-components, run:
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "tsc-alias": "1.8.10",
     "typescript": "5.7.3",
     "typescript-eslint": "8.24.1",
-    "vanilla-framework": "4.23.0",
+    "vanilla-framework": "4.24.1",
     "wait-on": "8.0.2",
     "webpack": "5.98.0"
   },
--- a/src/components/EventQueue/EventQueue.stories.tsx
+++ b/src/components/EventQueue/EventQueue.stories.tsx
@@ -1,0 +1,139 @@
+import React, { useState } from "react";
+import { Meta, StoryObj } from "@storybook/react";
+
+import { createEventQueue } from "./EventQueue";
+import Button from "../Button/Button";
+
+type Event = { message: string };
+
+const { EventQueueProvider, useEventQueue } = createEventQueue<Event>();
+
+const meta: Meta<typeof EventQueueProvider> = {
+  title: "Components/EventQueue",
+  component: EventQueueProvider,
+  tags: ["autodocs"],
+  parameters: {
+    docs: {
+      description: {
+        component: `
+This provides an event queue system for managing callbacks associated with
+asynchronous operations (e.g., API calls) in an application.
+
+It allows components to register success, failure,
+and optional finish handlers for a given operation ID, and later retrieve or remove them.
+
+This is useful for handling side effects when dealing
+with multiple operations that need to be tracked and updated independently.
+
+The \`createEventQueue\` function should be used to create a single provider and context that are shared throughout your application.
+The returned \`EventQueueProvider\` and \`useEventQueue\` hook should be exported
+from a shared module and reused across the app to ensure a single
+context instance is used.
+
+**Usage pattern:**
+
+\`\`\`tsx
+// eventQueue.ts
+export const { EventQueueProvider, useEventQueue } = createEventQueue<EventType>();
+
+// App.tsx
+import { EventQueueProvider } from "./eventQueue";
+
+<EventQueueProvider>
+  <App />
+</EventQueueProvider>
+
+// In any other component
+import { useEventQueue } from "./eventQueue";
+
+const eventQueue = useEventQueue();
+eventQueue.set(operationId, onSuccess, onFailure);
+\`\`\``,
+      },
+    },
+  },
+  decorators: [
+    (Story) => (
+      <EventQueueProvider>
+        <Story />
+      </EventQueueProvider>
+    ),
+  ],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof EventQueueProvider>;
+
+export const Default: Story = {
+  render: () => {
+    const EventQueueDemo = () => {
+      const queue = useEventQueue();
+      const operationId = "storybook-op-id";
+      const [logs, setLogs] = useState<string[]>([]);
+      const [isRegistered, setIsRegistered] = useState(false);
+
+      const appendLog = (msg: string) => setLogs((prev) => [...prev, msg]);
+
+      // This simulates the code that is executed when the asynchronous operation starts
+      const register = () => {
+        queue.set(
+          operationId,
+          (event) => appendLog(`Success: ${event.message}`),
+          (msg) => appendLog(`Failure: ${msg}`),
+          () => appendLog("Finished"),
+        );
+        appendLog("Registered callbacks");
+        setIsRegistered(true);
+      };
+
+      // This simulates the code that is executed when the asynchronous operation succeeds
+      const triggerSuccess = () => {
+        const event = queue.get(operationId);
+        if (event) {
+          event.onSuccess({ message: "This was successful!" });
+          event.onFinish?.();
+          queue.remove(operationId);
+          setIsRegistered(false);
+        }
+      };
+
+      // This simulates the code that is executed when the asynchronous operation fails
+      const triggerFailure = () => {
+        const event = queue.get(operationId);
+        if (event) {
+          event.onFailure("Something failed!");
+          event.onFinish?.();
+          queue.remove(operationId);
+          setIsRegistered(false);
+        }
+      };
+
+      return (
+        <div>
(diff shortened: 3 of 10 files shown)


src/components/ActionButton/ActionButton.tsx
```diff
@@ -71,12 +71,16 @@ const ActionButton = ({
...
+  const loadingTimeTracker = useRef<number>(0);
+  const startLoadTime = useRef<Date | undefined>(undefined);
 
   // Set up loader timer
   useEffect(() => {
     let loaderTimeout: number;
 
     if (loading) {
+      // Keep track of time when loading is occured.
+      startLoadTime.current = new Date();
```
edlerd: This will run twice, so not capture the exact event of loading start. Because after `loading` state changes to true, we hit this code branch. Then just below call `setShowLoader`, which might change the `showLoader` state, which then would re-trigger the use effect. Not a big issue though as the re-render should be fast.

edlerd (commented):
Thank you for the contribution! The change looks correct.

Can you please add tests for the corner cases?

One case is that loading is changing very quickly from false to true back to false. Then we should ensure the `LOADER_MIN_DURATION` is still applied.

Another case is for when we load longer than `LOADER_MIN_DURATION`, to ensure the visible loading state is not extended as it was before.

worapolw:
Hi @edlerd, thank you for reviewing my PR. 
I added tests for the edge cases and updated code as you suggested, please review my updated!

lorumic (changes requested):
@edlerd Please hold on, I am trying to clarify with @huwshimi why we need a `LOADER_MIN_DURATION` logic at all. In case it's agreed that we don't really need it, removing that instead of patching like it is currently being done in this PR could be a more desirable solution.

worapolw:
> @edlerd Please hold on, I am trying to clarify with @huwshimi why we need a `LOADER_MIN_DURATION` logic at all. In case it's agreed that we don't really need it, removing that instead of patching like it is currently being done in this PR could be a more desirable solution.

Ok, in this case, i'm going to let it be canonical team's decision about the next step of this PR.

lorumic:
> Ok, in this case, i'm going to let it be canonical team's decision about the next step of this PR.

Thanks for bearing with us @worapolw - we are currently checking with our design team whether we want to keep the `LOADER_MIN_DURATION` or get rid of it. If they tell us to keep it, I'll approve this PR and request another review from @edlerd since he reviewed it once. If instead we are told that it should be removed, I think you could maybe update this PR to change the logic so that there is no `LOADER_MIN_DURATION` anymore. I'll keep you updated with the outcome of the discussion.

lorumic:
> I will wait for the result of your team discussion and update code as the way canonical team want to go.

So the outcome is the following:
- We want to keep the `LOADER_MIN_DURATION` of 400ms, so the fix you applied @worapolw is relevant and you don't need to change that implementation.
- However, we should follow the design rule that "a button that is loading should be disabled" - which is what the original issue #1109 was all about - a button that becomes enabled too early (or rather, a loading animation that ends too late). This should never happen - not even when the request is completed in less than 400 ms.

So, to solve this, I think we can change the `disabled` condition of the `button` element inside `ActionButton` to be: `disabled={disabled || showLoader}`. Can you apply this little change @worapolw please? And also maybe cover it with the tests that you've added? Thanks :slightly_smiling_face:

src/components/ActionButton/ActionButton.tsx
```diff
@@ -139,7 +158,7 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
-      disabled={disabled}
+      disabled={disabled || showLoader}
```
edlerd:
```suggestion
      disabled={disabled === null ? showLoader : disabled}
```
I think this is too drastic, as it makes it impossible to enable a button if it is loading. There are use cases where we want to allow re-sending an action from a button in loading state. The current change would prevent the user from re-sending.

I suggest changing the disabled prop default to `null`. If disabled is not set by the caller, we can shadow the `showLoader` behavior. But if disabled is set by the caller, we should respect the value coming in.

edlerd (changes requested):
Thanks for adding the tests @worapolw ! And thanks @lorumic for clarifying with the design colleagues.

I have one last concern raised below, after that this is good to merge :)

src/components/ActionButton/ActionButton.tsx
```diff
@@ -71,12 +71,20 @@ const ActionButton = ({
...
 
   // Set up loader timer
   useEffect(() => {
     let loaderTimeout: number;
 
     if (loading) {
+      // add a condition to prevent double set startLoadTime
+      // when showLoader changes.
+      if (startLoadTime.current === undefined) {
+        // Keep track of time when loading is occured.
```
lorumic: Suggestion:
```suggestion
        // Keep track of the time when loading starts
```

src/components/ActionButton/ActionButton.tsx
```diff
@@ -86,13 +94,24 @@ const ActionButton = ({
     }
 
     if (!loading && showLoader) {
+      const now = new Date();
+      // calculate elasped time of loading from user and do subtraction to LOADER_MIN_DURATION,
```
lorumic: Suggestion:
```suggestion
      // calculate elapsed loading time and subtract it from LOADER_MIN_DURATION,
```

src/components/ActionButton/ActionButton.tsx
```diff
@@ -86,13 +94,24 @@ const ActionButton = ({
...
       loaderTimeout = window.setTimeout(() => {
+        loadingTimeTracker.current = 0;
+        startLoadTime.current = undefined;
         setShowLoader(false);
 
         if (success) {
           setShowSuccess(true);
         }
-      }, LOADER_MIN_DURATION);
+      }, loadingTimeTracker.current);
```
lorumic: Do we need to store `loadingTimeTracker` as a ref? Can't we simply put here the result of the calculation above? Something like:
```suggestion
      }, Math.max(LOADER_MIN_DURATION - loadingMilliseconds, 0));
```
Then of course we also wouldn't need to reset it to zero inside the timeout :)

lorumic: And maybe skip the timeout if the value is zero:
```
      const loadFinishHandler = () => {
        startLoadTime.current = undefined;
        setShowLoader(false);

        if (success) {
          setShowSuccess(true);
        }
      };

      const loadingTimeout = Math.max(LOADER_MIN_DURATION - loadingMilliseconds, 0);
      if (loadingTimeout > 0) {
        setTimeout(loadFinishHandler, loadingTimeout);
      } else {
        loadFinishHandler();
      }
```

lorumic (commented):
Thanks for the quick iterations @worapolw 
I had a deeper look at your changes and they mostly look OK - I just have a few ideas for minor improvements, let me know what you think about them.

worapolw:
> Thanks for the quick iterations @worapolw I had a deeper look at your changes and they mostly look OK - I just have a few ideas for minor improvements, let me know what you think about them.

Well, that's look better!

I cleaned up code as you suggested!

src/components/ActionButton/ActionButton.tsx
```diff
@@ -95,23 +94,29 @@ const ActionButton = ({
...
       const now = new Date();
-      // calculate elasped time of loading from user and do subtraction to LOADER_MIN_DURATION,
-      // also add an edge case when time diff is less than 0 to be 0.
+      // Keep track of time when loading is occured.
       const loadingMilliseconds: number =
         now.getTime() - (startLoadTime.current ?? now).getTime();
-      loadingTimeTracker.current = Math.max(
+
+      // also add an edge case when time diff is less than 0 to be 0.
+      const loadingTimeout = Math.max(
```
lorumic: Looks like something got jumbled a bit with the comments here:
```suggestion
      // calculate elapsed loading time
      const loadingMilliseconds: number =
        now.getTime() - (startLoadTime.current ?? now).getTime();

      // and subtract it from LOADER_MIN_DURATION,
      // also add an edge case when time diff is less than 0 to be 0.
      const timeoutDuration = Math.max(
```

lorumic: On a second thought, since we have the timeout itself called `loaderTimeout`, I think calling the duration of the timeout `loadingTimeout` is a bit confusing. Could you please rename this from `loadingTimeout` to `timeoutDuration`?

worapolw:
@lorumic Done! Thank you for your suggestion.

lorumic (approved):
Thanks for fixing this!

PR #1208, opened 2025-06-10
feat: add eventqueue to storybook

anusha-c18:
## Done

- Added eventqueue to storybook

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check the event queue story

### Percy steps

- No visual changes expected

## Fixes

Fixes: #[WD-22135](https://warthogs.atlassian.net/browse/WD-22135)

[WD-22135]: https://warthogs.atlassian.net/browse/WD-22135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/EventQueue/EventQueue.stories.tsx
+++ b/src/components/EventQueue/EventQueue.stories.tsx
@@ -1,0 +1,139 @@
+import React, { useState } from "react";
+import { Meta, StoryObj } from "@storybook/react";
+
+import { createEventQueue } from "./EventQueue";
+import Button from "../Button/Button";
+
+type Event = { message: string };
+
+const { EventQueueProvider, useEventQueue } = createEventQueue<Event>();
+
+const meta: Meta<typeof EventQueueProvider> = {
+  title: "Components/EventQueue",
+  component: EventQueueProvider,
+  tags: ["autodocs"],
+  parameters: {
+    docs: {
+      description: {
+        component: `
+This provides an event queue system for managing callbacks associated with
+asynchronous operations (e.g., API calls) in an application.
+
+It allows components to register success, failure,
+and optional finish handlers for a given operation ID, and later retrieve or remove them.
+
+This is useful for handling side effects when dealing
+with multiple operations that need to be tracked and updated independently.
+
+The \`createEventQueue\` function should be used to create a single provider and context that are shared throughout your application.
+The returned \`EventQueueProvider\` and \`useEventQueue\` hook should be exported
+from a shared module and reused across the app to ensure a single
+context instance is used.
+
+**Usage pattern:**
+
+\`\`\`tsx
+// eventQueue.ts
+export const { EventQueueProvider, useEventQueue } = createEventQueue<EventType>();
+
+// App.tsx
+import { EventQueueProvider } from "./eventQueue";
+
+<EventQueueProvider>
+  <App />
+</EventQueueProvider>
+
+// In any other component
+import { useEventQueue } from "./eventQueue";
+
+const eventQueue = useEventQueue();
+eventQueue.set(operationId, onSuccess, onFailure);
+\`\`\``,
+      },
+    },
+  },
+  decorators: [
+    (Story) => (
+      <EventQueueProvider>
+        <Story />
+      </EventQueueProvider>
+    ),
+  ],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof EventQueueProvider>;
+
+export const Default: Story = {
+  render: () => {
+    const EventQueueDemo = () => {
+      const queue = useEventQueue();
+      const operationId = "storybook-op-id";
+      const [logs, setLogs] = useState<string[]>([]);
+      const [isRegistered, setIsRegistered] = useState(false);
+
+      const appendLog = (msg: string) => setLogs((prev) => [...prev, msg]);
+
+      // This simulates the code that is executed when the asynchronous operation starts
+      const register = () => {
+        queue.set(
+          operationId,
+          (event) => appendLog(`Success: ${event.message}`),
+          (msg) => appendLog(`Failure: ${msg}`),
+          () => appendLog("Finished"),
+        );
+        appendLog("Registered callbacks");
+        setIsRegistered(true);
+      };
+
+      // This simulates the code that is executed when the asynchronous operation succeeds
+      const triggerSuccess = () => {
+        const event = queue.get(operationId);
+        if (event) {
+          event.onSuccess({ message: "This was successful!" });
+          event.onFinish?.();
+          queue.remove(operationId);
+          setIsRegistered(false);
+        }
+      };
+
+      // This simulates the code that is executed when the asynchronous operation fails
+      const triggerFailure = () => {
+        const event = queue.get(operationId);
+        if (event) {
+          event.onFailure("Something failed!");
+          event.onFinish?.();
+          queue.remove(operationId);
+          setIsRegistered(false);
+        }
+      };
+
+      return (
+        <div>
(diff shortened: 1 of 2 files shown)


src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+      operationId,
+      (event) => appendLog(`Success: ${event.message}`),
+      (msg) => appendLog(`Failure: ${msg}`),
+      () => appendLog("Finished"),
+    );
+    appendLog("Registered callback");
+    setIsRegistered(true);
+  };
+
+  const triggerSuccess = () => {
```
lorumic: Might be good to add a quick comment above this to explain what it represents?
```suggestion
  // This simulates the code that is executed when your asynchronous operation succeeds
  const triggerSuccess = () => {
```

src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+    const event = queue.get(operationId);
+    if (event) {
+      event.onSuccess({ message: "This was successful!" });
+      event.onFinish?.();
+      queue.remove(operationId);
+      setIsRegistered(false);
+    }
+  };
+
+  const triggerFailure = () => {
```
lorumic: And similarly here:
```suggestion
  // This simulates the code that is executed when your asynchronous operation fails
  const triggerFailure = () => {
```

src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+      queue.remove(operationId);
+      setIsRegistered(false);
+    }
+  };
+
+  return (
+    <div>
+      <div style={{ marginBottom: "0.5rem", display: "flex" }}>
+        <Button onClick={register} disabled={isRegistered}>
+          Register Callback
```
lorumic:
```suggestion
          Register Callbacks
```

src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const appendLog = (msg: string) => setLogs((prev) => [...prev, msg]);
+
+  const register = () => {
+    queue.set(
+      operationId,
+      (event) => appendLog(`Success: ${event.message}`),
+      (msg) => appendLog(`Failure: ${msg}`),
+      () => appendLog("Finished"),
+    );
+    appendLog("Registered callback");
```
lorumic:
```suggestion
    appendLog("Registered callbacks");
```

src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+const operationId = "storybook-op-id";
+
+const EventQueueDemo = () => {
+  const queue = useEventQueue();
+  const [logs, setLogs] = useState<string[]>([]);
+  const [isRegistered, setIsRegistered] = useState(false);
+
+  const appendLog = (msg: string) => setLogs((prev) => [...prev, msg]);
+
+  const register = () => {
```
lorumic:
```suggestion
  // This simulates the code that is executed when your asynchronous operation starts
  const register = () => {
```

src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+    ),
+  ],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof EventQueueDemo>;
+
+export const Default: Story = {
+  render: () => <EventQueueDemo />,
```
lorumic: This will hide the code example in the "Show code" view, which I think we should not do. We should provide a code example that is fully accessible to the storybook reader.

To follow the pattern we use in all the other stories, we should not define `EventQueueDemo` as the component here, but instead consider `EventQueueProvider` as our component in this context (even though *technically* it is not a "component" in the React sense), and move all the content of the `EventQueueDemo` function here. See before and after of doing this:

- Before (i.e. now)

image

- After

image

anusha-c18: I've updated the story to inline the logic within the render function using a nested component. This avoids the rules-of-hooks violation while the full code is visible in the show code view.

lorumic: Hmm, I'm not sure how what I proposed above violates the rules of hooks. I was suggesting to do exactly what you have done, but without wrapping everything inside the `EventQueueDemo` component. You can directly return the component from the `render` prop - I even tested it locally and did not experience any issue. Am I missing something?

src/components/EventQueue/EventQueue.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+              {logs.map((log, index) => (
+                <li key={index}>{log}</li>
+              ))}
+            </ul>
+          </div>
+        </div>
+      );
+    };
+
+    return <EventQueueDemo />;
```
lorumic, line 137: To be clear, this is what I have done locally and did not see any issue with it:
```suggestion
  render: () => {
    const queue = useEventQueue();
    const operationId = "storybook-op-id";
    const [logs, setLogs] = useState<string[]>([]);
    const [isRegistered, setIsRegistered] = useState(false);

    const appendLog = (msg: string) => setLogs((prev) => [...prev, msg]);

    // This simulates the code that is executed when the asynchronous operation starts
    const register = () => {
      queue.set(
        operationId,
        (event) => appendLog(`Success: ${event.message}`),
        (msg) => appendLog(`Failure: ${msg}`),
        () => appendLog("Finished"),
      );
      appendLog("Registered callbacks");
      setIsRegistered(true);
    };

    // This simulates the code that is executed when the asynchronous operation succeeds
    const triggerSuccess = () => {
      const event = queue.get(operationId);
      if (event) {
        event.onSuccess({ message: "This was successful!" });
        event.onFinish?.();
        queue.remove(operationId);
        setIsRegistered(false);
      }
    };

    // This simulates the code that is executed when the asynchronous operation fails
    const triggerFailure = () => {
      const event = queue.get(operationId);
      if (event) {
        event.onFailure("Something failed!");
        event.onFinish?.();
        queue.remove(operationId);
        setIsRegistered(false);
      }
    };

    return (
      <div>
        <div style={{ marginBottom: "0.5rem", display: "flex" }}>
          <Button onClick={register} disabled={isRegistered}>
            Register Callbacks
          </Button>
          <Button onClick={triggerSuccess} disabled={!isRegistered}>
            Trigger Success
          </Button>
          <Button onClick={triggerFailure} disabled={!isRegistered}>
            Trigger Failure
          </Button>
        </div>
        <div>
          <strong>Logs:</strong>
          <ul>
            {logs.map((log, index) => (
              <li key={index}>{log}</li>
            ))}
          </ul>
        </div>
      </div>
    );
```

src/components/EventQueue/EventQueue.stories.tsx
anusha-c18: You need to run the linter to see the rules-of-hooks violation. It doesn't recognize the render function as a valid React component so I wrapped it in the `EventQueueDemo` component.

lorumic: Ah, it's a linting error! I was checking in the browser's console and could not find anything.
Ok, cool, thanks for clarifying :) LGTM then :+1:

Issue #1188, opened 2025-03-24
Icon names in Icon props are out-of-date


jmuzina:
The Icon component's props define a set of expected icon names here:
https://github.com/canonical/react-components/blob/60dd4449c266b6d3d08dbb9ec6d4a164162a3944/src/components/Icon/Icon.tsx#L7

However, this is out-of-date with the current state of Vanilla.

<details>
<summary>
The following icons are missing (click to expand)
</summary>
```
success-grey
error-grey
chevron-right
chevron-left
question
applications
controllers
fullscreen
models
machines
pin
units
priority-critical
priority-high
priority-low
priority-medium
priority-negligible
priority-unknown
add-canvas
add-logical-volume
add-partition
back-to-top
begin-downloading
bundle
canvas
change-version
comments
conflict-grey
conflict-resolution-grey
conflict-resolution
conflict
connected
containers
copy-to-clipboard
disconnect
edit
export
exposed
filter
fork
get-link
halfscreen-bar
highlight-off
highlight-on
home
import
in-progress
inspector-debug
loading-steps
lock-locked-active
lock-locked
lock-unlock
maximise-bar
minimise-bar
mount-2
mount
open-terminal
plans
pods
power-error
power-off
power-on
profile
restart
revisions
security
settings
sort-both
sort-down
sort-up
starred
status-failed-small
status-in-progress-small
status-in-progress
status-queued-small
status-queued
status-succeeded-small
status-waiting-small
status-waiting
status
submit-bug
switcher-dashboard
switcher-environments
switcher
tag
task-outstanding
timed-out-grey
timed-out
topic
unit-pending
unit-running
unmount
unstarred
user-group
video-play
warning-grey
desktop
play
pause
stop
archive
arrow-bottom-left
arrow-bottom-right
arrow-down
arrow-left
arrow-right
arrow-top-left
arrow-top-right
arrow-up
blueprint
book
certificate
certification
cluster-host
contact
contextual-menu
cursor
file-blank
file
folder
gift
image
iso
location
log-out
map
notifications
private-key
profiles
repository
security-error
security-tick
security-warning
select-add
select-remove
select
single-host
snapshot
snooze
statistics
thumbs-down
thumbs-up
tidy
toggle-side-nav
turn-off-notification
upload
usb
website
credit-card
heading
quote
```
</details>

This is quite a few missing icons, and the type currently allows for other strings to be passed in, so I'm not sure how high-priority it is to add these icons to the props.
[...]

PR #1217, opened 2025-07-01
feat: support new chip variants (read-only, icon, badged), deprecate StatusLabel

jmuzina:
## Done

- React implementation of https://github.com/canonical/vanilla-framework/pull/5555 . Will not appear as expected until that is merged and released (expected Vanilla 4.25.0).
- Implements the new read-only, badged, and icon Chips.
- Implements some Chip states that have been in Vanilla for a while but have not been added to React Components - dense & inline
- Deprecates the Status Label component.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

1. Open [read-only chip](https://react-components-1217.demos.haus/?path=/story/components-chip--read-only). 
  a. Verify that it is completely non-interactive. 
    i. It should not be focusable. 
    ii. Clicking it or hovering it should not effect its state.
2. Open [dense chip](https://react-components-1217.demos.haus/?path=/story/components-chip--dense). Compare it to the default chip, and see that it has less padding.
3. Open [inline chip](https://react-components-1217.demos.haus/?path=/story/components-chip--inline). See that it is rendered inline with paragraph text. 
4. Open [icon chip](https://react-components-1217.demos.haus/?path=/story/components-chip--with-icon). See that the icon is displayed to the left of the Chip text.
5. Open [badge chip](https://react-components-1217.demos.haus/?path=/story/components-chip--with-badge). See that the badge is displayed at the right of the value, before the dismiss button. 

### Percy steps

Review new chip stories:
- ReadOnly
- Dense
- Inline
- WithIcon
- WithBadge

### DIFF
--- a/src/components/Chip/Chip.stories.tsx
+++ b/src/components/Chip/Chip.stories.tsx
@@ -43,3 +43,43 @@ export const Dismissible: Story = {
 
   name: "Dismissible",
 };
+
+export const Dense: Story = {
+  render: () => <Chip lead="Owner" value="Bob" isDense={true} />,
+  name: "Dense",
+};
+
+export const Inline: Story = {
+  render: () => (
+    <p>
+      This is text with an inline{" "}
+      <Chip value="chip" isDense={true} isInline={true} />
+    </p>
+  ),
+  name: "Inline",
+};
+
+export const ReadOnly: Story = {
+  render: () => <Chip lead="Owner" value="Bob" isReadOnly={true} />,
+  name: "Read-only",
+};
+
+export const WithIcon: Story = {
+  render: () => <Chip value="Users" iconName="user" />,
+  name: "With Icon",
+};
+
+export const WithBadge: Story = {
+  render: () => (
+    <Chip
+      lead="Owner"
+      value="Bob"
+      badge={
+        <span className="p-badge" aria-label="More than 2.5 billion items">
+          2.5B
+        </span>
+      }
+    />
+  ),
+  name: "With Badge",
+};
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -1,8 +1,9 @@
-import React from "react";
+import React, { type ReactNode } from "react";
 import { highlightSubString } from "../../utils";
 import type { KeyboardEvent, MouseEvent, HTMLProps } from "react";
 import { ValueOf, PropsWithSpread } from "types";
 import classNames from "classnames";
+import { ICONS } from "../Icon";
 
 export enum Label {
   Dismiss = "Dismiss",
@@ -22,6 +23,29 @@ export type Props = PropsWithSpread<
      */
     appearance?: ValueOf<typeof ChipType>;
 
+    /**
+     * Whether the chip is read-only.
+     * If true, the chip will not be interactive.
+     */
+    isReadOnly?: boolean;
+    /**
+     * A badge to display on the chip.
+     */
+    badge?: ReactNode;
+    /**
+     * The name of an icon to display on the chip.
+     */
+    iconName?: ValueOf<typeof ICONS> | string;
+    /**
+     * Whether the chip is dense.
+     * Reduces the height of the chip by removing padding.
+     */
+    isDense?: boolean;
+    /**
+     * Whether the chip is inline.
+     * If true, the chip will be displayed inline with other content (such as text).
+     */
+    isInline?: boolean;
     /**
      * The lead for the chip.
      */
@@ -70,6 +94,11 @@ const Chip = ({
   quoteValue,
   selected,
   subString = "",
+  isReadOnly = false,
+  isDense = false,
+  isInline = false,
+  iconName,
+  badge,
   value,
   ...props
 }: Props): React.JSX.Element => {
@@ -88,6 +117,7 @@ const Chip = ({
 
   const chipContent = (
     <>
+      {iconName && <i className={`p-icon--${iconName}`} />}
       {lead && <span className="p-chip__lead">{lead.toUpperCase()}</span>}
       <span
         className="p-chip__value"
@@ -95,6 +125,7 @@ const Chip = ({
           __html: quoteValue ? `'${chipValue}'` : chipValue,
         }}
       />
+      {badge && badge}
     </>
   );
 
@@ -102,11 +133,20 @@ const Chip = ({
     {
       [`p-chip--${appearance}`]: !!appearance,
       "p-chip": !appearance,
+      "is-dense": isDense,
+      "is-readonly": isReadOnly,
+      "is-inline": isInline,
     },
     props.className,
   );
 
-  if (onDismiss) {
+  if (isReadOnly) {
+    return (
+      <span {...props} className={chipClassName}>
+        {chipContent}
+      </span>
+    );
+  } else if (onDismiss) {
     return (
       <span {...props} className={chipClassName}>
         {chipContent}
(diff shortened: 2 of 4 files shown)


src/components/Chip/Chip.stories.tsx
```diff
@@ -43,3 +43,30 @@ export const Dismissible: Story = {
...
+  render: () => <Chip lead="Owner" value="Bob" isReadOnly={true} />,
+  name: "Read-only",
+};
+
+export const WithIcon: Story = {
+  render: () => (
+    <Chip lead="Deployment" value="Successful" iconName="success" />
+  ),
+  name: "With icon",
+};
```
edlerd: I realize this is still in draft state, but was curious about the new chip features. The icon looks slightly too high within the chip.

image

jmuzina: Yes, this is because Vanilla 4.25.0 hasn't been released, so the styling that aligns the icon is not yet in react components. 

You can expect it to look like this:

image

Here's the [design docs](https://docs.google.com/document/d/1HBZ_i7nI0FQbR9FJ_SE0x3Ax_BUi4DnBXMEhZPp2BiI/edit?tab=t.5qfuxdsojs30#heading=h.h2oevae7cg95) if you're curious :)

src/components/Chip/Chip.tsx
```diff
@@ -22,6 +22,29 @@ export type Props = PropsWithSpread<
...
+     */
+    isReadOnly?: boolean;
+    /**
+     * A badge to display on the chip.
+     */
+    badge?: ReactNode;
+    /**
+     * The name of an icon to display on the chip.
+     */
+    iconName?: string;
```
edlerd: Thanks for adding the chip with icon.

1. We used this before in LXD-UI. I realize the size of the icons is a bit smaller now. Is the small size intended to be aligned with the text? I might prefer the bigger icons, but not 100% sure on this one, and I am not a graphic designer. This used to be `background-size: contain`, which is now overridden.

2. I think the left margin of the icon in the chip should be removed. Can we change it in vanilla?


Before:

image

New:

image

suggested change with fewer left margin on the icon:

image

jmuzina: Referring this to @timea-udvarhelyi @dgtlntv

dgtlntv: Regarding the left margin of the icon its something I had already raised with @timea-udvarhelyi and I agree that we should remove it from the icon, it should only have a right margin.

Regarding the size of the icon I agree that it is quite small especially for more detailed icons it becomes quite hard to read. Something like 0.875rem (14px) feels more appropriate to me. Maybe @lyubomir-popov can give an opinion what the exact size of the icon should be?

jmuzina: Opened a [PR](https://github.com/canonical/vanilla-framework/pull/5574) that we can use to make these adjustments. I've already removed the left margin. Pending @lyubomir-popov thoughts on icon sizing we can also implement that change there. 

The chip icon is currently using some [common styling with the dismiss icon](https://github.com/canonical/vanilla-framework/blob/45b3421ba5aead447f3c0b6717713072b95d0292/scss/_patterns_chip.scss#L5-L15) - they both use the "small" icon size.

lyubomir-popov: My intuiton is that, since at paragraph font-size (1rem/16px), icons match that size (1rem again), going lower should happen in sync - 14px icon/14px font size.

Now, this strategy should only be used going downward - we shouldn't make icons next to h1 headings (outside chips) match the h1 font-size (2.5rem), as the icons in their current state are too chunky and low in detail to scale them up. But that precise feature of the icons means it is safe to go down to .875rem (14px) to accompany small text.

jmuzina: Icon size will stay at 14px for now as it aligns with current Vanilla convention of following the font size of the text next to it, as long as it is <= 16px (in this case the chip text, which is 14px)

If we want to change the icon size within the chip, that should go through the contribution process

edlerd:
```suggestion
    iconName?: ValueOf<typeof ICONS> | string;
```
This gives auto-completion, when using it.
Would be nice to extend the list of available strings in the `ICONS` type from `Icon.tsx` to include all the newly added icons.

src/components/Chip/Chip.tsx
```diff
@@ -22,6 +22,29 @@ export type Props = PropsWithSpread<
      */
     appearance?: ValueOf<typeof ChipType>;
 
+    /**
+     * Whether the chip is read-only.
+     * If true, the chip will not be interactive.
+     */
+    isReadOnly?: boolean;
```
edlerd, line 30: Not sure if the chip without border is a good way to signal it being read only. We tend to use the non-bordered chip for inline in LXD-UI, but still being interactive:

image

Maybe the read only variant can have a different style. Lower opacity or some other property changing to mark it as non-interactive. And maybe we should add the without border as a variant, like the is dense or is inline variant. Or maybe the is inline variant should automatically be without border. What do you think?

edlerd (approved):
Thanks for the iteration on the design! Code LGTM. One non-blocking idea for better typing, and another non-blocking suggestion for the styling.

src/components/Chip/Chip.tsx
jmuzina: Borderless (on top of no keyboard/mouse interactions) being used to indicate read-only comes from this [design spec](https://docs.google.com/document/d/1HBZ_i7nI0FQbR9FJ_SE0x3Ax_BUi4DnBXMEhZPp2BiI/edit?tab=t.5qfuxdsojs30#heading=h.co5ihleltffd).

The inline variant has already been in Vanilla for some time (just not explicitly documented in RC) - so we can't make it borderless by default without causing a visual change to people who are already using it - not sure if this would be considered breaking visually, wdyt @lyubomir-popov @dgtlntv ?

Am I to understand that you are using a chip styling override to remove the border for these cases, and aren't sure about if `is-readonly` is the right way to cover those use cases?

edlerd: > Am I to understand that you are using a chip styling override to remove the border for these cases, and aren't sure about if is-readonly is the right way to cover those use cases?

My concern is our UI already using a borderless chip which is interactive. Now, using borderless chips to indicate non-interactivity would be confusing.

edlerd: I am not sure borderless is a clear way to indicate no interactivity. Hence, my suggestion to change the read only design.

JanOstrowka: Thanks for the context, everyone! Just to clarify things a bit: the borderless chip style for read-only use was introduced recently after internal and external benchmarking, along with agreement from the design, visual, and design system teams. The idea is that a visible border signals interactivity, while no border means the chip is non-interactive or read-only. This follows a similar pattern to the old status label component, which was also non-interactive and had no border.

From the original screenshots, it looks like the borderless-but-interactive chip comes from a suggestion by @piperdeck, who proposed a custom component called an "entity chip". It follows a different interaction model than regular chips, and we’ve considered it to be a separate component that would require its own documentation and contribution. That’s why it hasn’t been included as part of the core chip component so far.

If the entity chip is still needed in LXD, I’d suggest going through the official contribution process so that its design and interaction patterns can be properly defined and documented. For now, it seems like a very specific component tied to LXD, and getting it reviewed would help clarify its place in the system and potentially open it up for broader use.

src/components/Chip/Chip.tsx
jmuzina: Updated the type to use the ICONS constant.

I'd propose that updating the ICONS constant be discussed separately in #1188

PR #1220, opened 2025-07-01
fix(step): use vanilla theme colors for stepper component

edlerd:
## Done

- fix(step) use vanilla theme colors for stepper component

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- check the stepper component in storybook

### Percy steps

- the selected colors on the stepper component should be a bit more visible, the background of the active step is more visible. the outline of an active step circle should be improved.

### DIFF
--- a/src/components/Stepper/Step/Step.scss
+++ b/src/components/Stepper/Step/Step.scss
@@ -1,7 +1,7 @@
 @import "vanilla-framework";
 
 .step-number {
-  border: 0.08rem solid black;
+  border: 0.08rem solid $colors--theme--border-high-contrast;
   border-radius: 1rem;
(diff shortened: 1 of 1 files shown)


jmuzina:
I wonder if we could also theme the `.progress-line` which is currently set to black - maybe `$colors--theme--border-high-contrast` would work?

Dark before/after:

image
image



Light before/after:

Screenshot from 2025-07-01 11-40-39
image

I'll check with a designer.

anusha-c18 (approved):
lgtm :)

PR #1228, opened 2025-07-15
chore: allow table pagination controls with unknown entries

Ninfa-Jeon:
## Done

- Added entry count types: `known` and `unknown`
- Allowed disabling of input field in `TablePaginationControls` only if the entry count type is `known`

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

- "No visual changes expected"

## Fixes

Fixes: # .


src/components/TablePagination/TablePaginationControls/TablePaginationControls.test.tsx
```diff
@@ -198,4 +198,26 @@ describe("<TablePaginationControls />", () => {
...
+        onPageChange={jest.fn()}
+        onPageSizeChange={jest.fn()}
+        entryCountType="unknown"
+      />,
+    );
+
+    const pageInput = screen.getByRole("spinbutton", {
+      name: Label.PAGE_NUMBER,
+    });
+    expect(pageInput).not.toBeDisabled();
```
anusha-c18:
```suggestion
    expect(pageInput).toBeEnabled();
```

src/components/TablePagination/TablePaginationControls/TablePaginationControls.test.tsx
```diff
@@ -198,4 +198,26 @@ describe("<TablePaginationControls />", () => {
     expect(pageInput).toHaveAttribute("min", "1");
     expect(pageInput).toHaveAttribute("max", "5");
   });
+
+  it("enables page input when there are known number of entries", () => {
```
anusha-c18: Shouldn't this be unknown?
```suggestion
  it("enables page input when there are unknown number of entries", () => {
```

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
```diff
@@ -46,6 +48,7 @@ export type Props = Omit<
   totalItems?: AllProps["totalItems"];
   visibleCount?: number;
   showPageInput?: boolean;
+  entryCountType?: EntryCountType;
```
anusha-c18: I don’t think we need a custom type here. Could we simplify this to a boolean instead?

```suggestion
  isTotalKnown?: boolean;
```

Feel free to improve the name if needed.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
anusha-c18, line 1: Unrelated to the changes in this PR, but I noticed the horizontal spacing looks off when `totalItems` isn’t defined:
Screenshot 2025-07-15 at 11 29 21 PM
Lmk if you’d like to address this, or I can create a follow up PR

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
jmuzina: Do we need a new attribute at all? Could we check for `totalItems` being undefined instead?

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
jmuzina, line 1: Could you also add a story for the TablePaginationControls component that shows pagination with an unknown number of items?

jmuzina, line 1: I think this may require an update to the `MainTable` component for this change to be fully usable.

`MainTable` embeds a paginator when `paginate` is true and there is more than one element in `rows`. It passes `rows.length` into the paginator. 

https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/components/MainTable/MainTable.tsx#L390-L398

If I understand your openfga use case correctly, you don't know how many total items there are, and setting `rows.length` to be the `totalItems` would be misleading, and would in fact cause your no-total-known case to never be rendered in `MainTable`, assuming we are considering this case to be triggered by `totalItems` being undefined.

Upon closer inspection of the Table code, I think we need to make some changes to the `usePagination` hook and the `MainTable` component to make this change ready.

1. Update `usePagination` so that `totalitems` may be undefined (currently it [cannot](https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/hooks/usePagination.ts#L24)). I wonder if there are any risks of `totalItems` possibly being undefined, we should be careful about this and rigorous with our tests.
2. Update `usePagination` so that `totalItems` is bound to the number of items, if the number of items is known - otherwise it should be undefined. Currently it is only ever `items.length`. 
    * This makes me rethink my [comment here](https://github.com/canonical/react-components/pull/1228/files#r2208485941) about simply using `undefined` to indicate an unknown number of rows case - this would leave `usePagination` in an incorrect and only partially useful state in these cases - the change should occur in the hook, not the pagination controls component.  I think the `isTotalKnown` suggested by @anusha-c18 could work well, but maybe it should also be included in `usePagination`, [here](https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/hooks/usePagination.ts#L14-L18), and users of `MainTable` should be able to also pass `isTotalKnown` which can then be passed to `usePagination`. 
3. `MainTable` should consume `totalItems` from `usePagination` (currently [it doesn't](https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/components/MainTable/MainTable.tsx#L337-L344)) and pass that to the paginator, rather than using `rows.length` as it does currently [here](https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/components/MainTable/MainTable.tsx#L396). 
4. `MainTable` should embed a paginator if `paginate` is true and `totalItems` is not zero, instead of [relying on `totalItems` being truthy](https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/components/MainTable/MainTable.tsx#L390), which would cause the paginatior to not be rendered if `totalItems` is undefined.

Sorry for the wordy response, I hope it is clear - let me know what you think!

jmuzina, line 1: Can you also add a story for `MainTable` with a paginated dataset with an unknown number of items?

jmuzina: Disregard this, I mixed up the `Pagination` component used by `MainTable` with the `TablePagination` component :sweat_smile:

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
```diff
@@ -103,7 +107,8 @@ const TablePaginationControls = ({
     onPageSizeChange(parseInt(e.target.value));
   };
 
-  const isInputDisabled = !totalPages || totalPages == 1;
+  const isInputDisabled =
+    entryCountType === "known" && (!totalPages || totalPages == 1);
```
jmuzina: On second look, I think we should just not render the input at all in case the total is unknown, following [Vanilla pagination guidelines](https://vanillaframework.io/docs/patterns/pagination#previous--next-buttons-only).

image

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
jmuzina: I QA'd this locally and found that when `totalItems` is undefined, the pagination summary /description still reads "Showing 10 out of undefined rows"

image

You will likely need to update [`getDescription`](https://github.com/canonical/react-components/blob/28031157ae158abe0da52bb6697b115c0e6257da/src/components/TablePagination/utils.tsx#L39-L67) to account for this.

Note, I raised an [issue](https://github.com/canonical/vanilla-framework/issues/5581) with the Design System Working Group proposing to formalize the behavior when the total is unknown to read like "Showing 10 rows", because I'm not sure if Design has a formal answer to that yet. If you need this shipped before they can provide input on it, I think it's fine to use `Showing x rows` in case the total is unknown - potentially imperfect is better than broken :)

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
anusha-c18: From the context I've gathered, this approach makes sense. If `totalItems` isn't defined, we could skip showing the number input.
That said, I'm curious, when the total number of items isn't known, were users actually able to fetch data by jumping to a specific page using the number input before [this change](https://github.com/canonical/react-components/commit/5e37c791299c9f7b558a65ad0923fee5ff1da1a2#diff-04e1b6c4a62d90c8efc9068cb59eb91ed91e861db90f572cdd88cf4eb0707dccR106)? @Ninfa-Jeon 
If yes, then maybe it's worth making the component more flexible. For example, we can show the number input by default and also provide a prop to hide it, as suggested by the vanilla guidelines.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
anusha-c18: Similar question here
When the total number of items isn't known, what was displayed in the pagination summary before [this change](https://github.com/canonical/react-components/commit/5e37c791299c9f7b558a65ad0923fee5ff1da1a2#diff-04e1b6c4a62d90c8efc9068cb59eb91ed91e861db90f572cdd88cf4eb0707dccR106)? @Ninfa-Jeon

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
huwshimi: > That said, I'm curious, when the total number of items isn't known, were users actually able to fetch data by jumping to a specific page using the number input before [this change](https://github.com/canonical/react-components/commit/5e37c791299c9f7b558a65ad0923fee5ff1da1a2#diff-04e1b6c4a62d90c8efc9068cb59eb91ed91e861db90f572cdd88cf4eb0707dccR106)? @Ninfa-Jeon

From memory the way this worked is that because of the limitations of the backend not knowing how many items there are, the input allowed you to move around more quickly then clicking next or back repeatedly if you knew that there were a certain number of items (e.g. if you're on page 10 you could enter 5 to jump back - you can also enter 100 and if there are no items we display a message). It's not an ideal experience but rather a workaround to improve the experience.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
Ninfa-Jeon: Simply checking for `totalItems` being undefined would enable the input for cases where there are no items on the page. This would especially be faulty in the majority cases where there is a known number of entries because allowing the user to jump through pages when there are no entries at all makes no sense. For our case, however, we just don't know the overall number of entries but the page might still exist for the user to jump to.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.test.tsx
Ninfa-Jeon: Nice find! I have changed another test using this check as well.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.test.tsx
Ninfa-Jeon: Oops.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
Ninfa-Jeon: Hi, I have addressed this issue. I have updated a few things here and there to make sure nothing breaks.

Ninfa-Jeon: Hi, I have added a story for the new case. I have also updated the description to support undefined `totalItems`. I'm afraid I don't have an answer to this @anusha-c18  as we have disabled descriptions for us. Enabling them locally gives me the following:
Screenshot 2025-07-16 at 9 35 29 AM

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
```diff
@@ -46,6 +46,7 @@ export type Props = Omit<
   totalItems?: AllProps["totalItems"];
   visibleCount?: number;
   showPageInput?: boolean;
+  isEntryCountUnknown?: boolean;
```
anusha-c18: Small nitpick, we could rename `isEntryCountUnknown` to `isTotalKnown` for clearer, positive naming. It aligns better with the existing `totalItems` prop and makes it slightly more readable.

src/components/TablePagination/utils.tsx
```diff
@@ -52,16 +52,17 @@ export const getDescription = ({
   if (description) {
     return description;
   }
+  const closing = totalItems ? ` out of ${totalItems}` : "";
 
   if (isSmallScreen) {
-    return `${visibleCount} out of ${totalItems}`;
+    return visibleCount + closing;
```
anusha-c18: maybe more consistent?
```suggestion
    return `${visibleCount}${closing}`;
```

src/components/TablePagination/TablePagination.scss
```diff
@@ -26,19 +26,23 @@
...
-    margin: 0 $spv--small 0 $spv--large;
+    margin: 0 $spv--large;
     min-width: 0;
     width: 4rem;
   }
 
+  .pagination-item-count {
+    margin: 0 $spv--large 0 0;
+  }
+
```
anusha-c18, line 45: This also needs to be handled for smaller screens, it is not responsive atm:


https://github.com/user-attachments/assets/321203cc-95fc-4220-8592-225395a7a1fc

anusha-c18: The same issue arises when the controls are known:

Before this change:

https://github.com/user-attachments/assets/f307a874-97dc-455f-9434-78468b447292

After this change:

https://github.com/user-attachments/assets/8a4783ea-9f59-4c64-aa94-026c0937320a

anusha-c18 (changes requested):
Thanks for the changes!
Found a couple of issues on QA and have a few more suggestions

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
anusha-c18: Thanks for the context! @huwshimi
I think we should keep the input visible for this use case instead of hiding it when `totalItems` is undefined @jmuzina

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
Ninfa-Jeon: Done! I have renamed it to `isTotalUnknown` as a user would only need to set this for unknown cases.

src/components/TablePagination/utils.tsx
Ninfa-Jeon: Done!

src/components/TablePagination/TablePagination.scss
Ninfa-Jeon: My bad, I didn't check it on smaller screens. It should look fine now. Thanks for bringing this up.

src/components/TablePagination/TablePaginationControls/TablePaginationControls.tsx
bartaz: I'm also unclear here why we need this flag.

It seems that the value of `totalItems` should be enough to know if total is known or not.
Is there a case where `totalItems` is `undefined`, but we would set `isEntryCountUnknown` to true?

anusha-c18: @Ninfa-Jeon

Ninfa-Jeon: Hi @jmuzina @anusha-c18 @bartaz I have removed the extra parameter. Upon checking code within the wrapper `TablePagination`, it is confirmed that for internally controlled `TablePaginationControls`, the value of `totalItems` will always be a number (as we are using `data.length` to get this value).

For cases other where it is externally controlled, if `totalItems` is received as `undefined` instead of `0`, the input will be enabled to cover cases of unknown entries.

This should be non-breaking for everyone using either the internally controlled version of this component or setting `totalItems` to a numerical value manually.

src/components/TablePagination/utils.tsx
jmuzina, line 1: The description reads "Showing x rows"

Screenshot 2025-07-31 at 18 27 44

Referencing [this comment](https://github.com/canonical/vanilla-framework/issues/5581#issuecomment-3077730006) by @dgtlntv:

> We think that option that shows just the amount of rows should be used in a case where we dont know if there are more rows.
We think we should do it similarly to Material indicating that there are more rows if we know there are more rows (but we dont know the exact number).

A screenshot from Material is attached:

image

Is it possible for us to contextually respond to whether a next page is available or not, so that if there are more pages possible in either prev or next, it says "showing x rows of more than x"?

Ninfa-Jeon: Hi @jmuzina I have added a couple of changes to the pagination description logic so that we can align with the approach you mention. For the estimated/partially known scenario, I don't think we have any way of knowing if there are items on the next page beforehand (unless it is a fully known entry count) but we can still infer that there are more entries than currently shown based on the current page number. Therefore, I used that to split the logic between the three scenarios. Thanks for your comment!

src/components/TablePagination/utils.tsx
```diff
@@ -42,28 +42,36 @@ export const getDescription = ({
...
   if (visibleCount === totalItems && visibleCount > 1) {
     return `Showing all ${totalItems} ${itemName}s`;
   }
 
-  return `Showing ${visibleCount} out of ${totalItems} ${itemName}${
-    totalItems !== 1 ? "s" : ""
-  }`;
+  return `Showing ${visibleCount} ${itemName}${
+    visibleCount !== 1 ? "s" : ""
+  }${closing}`;
```
jmuzina: This has caused a slight change in verbage in the description when the page limit is known, that was caught as a [visual diff in Percy](https://percy.io/bb49709b/web/react-components/builds/42135988/changed/2253370969?browser=firefox&browser_ids=70&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=overlay&viewMode=original&width=1280&widths=1280)

Before:
"Showing 1 out of 5 items"

After: 
"Showing 1 item out of 3"

Could you correct that case to avoid unexpected changes for existing users?

The case where the total is unknown or partially known can stay the same I think.

Ninfa-Jeon: I don't have access to view the Percy dashboard but I have reverted the logic for known case. Thanks!

anusha-c18 (approved):
Thanks for the fixes, Lgtm 👍💯

PR #1229, opened 2025-07-18
fix(sidepanel): default size of sidepanel to be standard, allow for narrow and wide sizes

edlerd:
## Done

- fix(sidepanel) default size of sidepanel to be standard, allow for narrow and wide sizes

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- check [sidepanel component](https://react-components-1229.demos.haus/?path=/docs/components-sidepanel--docs) default to be standard size, with options for narrow and wide

### Percy steps

- sidepanel width should be a bit more as the narrow is not the default anymore.

### DIFF
--- a/src/components/SidePanel/SidePanel.test.tsx
+++ b/src/components/SidePanel/SidePanel.test.tsx
@@ -31,7 +31,7 @@ it("displays as pinned", () => {
 });
 
 it("displays as wide", () => {
-  render(<SidePanel wide>SidePanel</SidePanel>);
+  render(<SidePanel width="wide">SidePanel</SidePanel>);
   expect(screen.getByText("SidePanel")).toHaveClass("is-wide");
 });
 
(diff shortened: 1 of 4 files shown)


src/components/SidePanel/SidePanel.test.tsx
```diff
@@ -31,7 +31,7 @@ it("displays as pinned", () => {
 });
 
 it("displays as wide", () => {
```
Ninfa-Jeon, line 33: can you add a case for `narrow` as well?

edlerd: It is already there :)

Ninfa-Jeon: oh, my bad. it wasn't specifically named "narrow" so I missed it. all good to merge!

PR #1241, opened 2025-08-07
feat(ToastNotifications): Allow id to be overwritten by notify functions

rumanowskak:
## Done

- Added id prop to `success`, `failure`, `info`, and `caution` methods from `ToastNotificationProvider`
- (driveby) Use `messageBuilder` functions when adding toast notifications
- (driveby) extract error message formatting from `failure` message builder
- (driveby) add button to storybook to add a warning notification

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### Percy steps

- Go to toast notifications
- Click "Add warning"
- Ensure a caution/warning notification is added to the list
- Ensure the counter for warnings is incremented by 1

### DIFF
--- a/src/components/NotificationProvider/messageBuilder.tsx
+++ b/src/components/NotificationProvider/messageBuilder.tsx
@@ -10,10 +10,15 @@ export const queue = (notification: NotificationType): QueuedNotification => {
   return { state: { queuedNotification: notification } };
 };
 
-export const info = (message: ReactNode, title?: string): NotificationType => {
+export const info = (
+  message: ReactNode,
+  title?: string,
+  actions?: NotificationAction[],
+): NotificationType => {
   return {
     message,
     title,
+    actions,
     type: NotificationSeverity.INFORMATION,
   };
 };
@@ -21,10 +26,12 @@ export const info = (message: ReactNode, title?: string): NotificationType => {
 export const success = (
   message: ReactNode,
   title?: string,
+  actions?: NotificationAction[],
 ): NotificationType => {
   return {
     message,
     title,
+    actions,
     type: NotificationSeverity.POSITIVE,
   };
 };
@@ -37,15 +44,33 @@ export const failure = (
 ): NotificationType => {
   return {
     actions,
-    message:
-      error && error instanceof Error ? (
-        <>
-          {message} {error.message}
-        </>
-      ) : (
-        message
-      ),
+    message: formatErrorMessage(message, error),
     title,
     type: NotificationSeverity.NEGATIVE,
   };
+};
+
+export const caution = (
+  message: ReactNode,
+  title?: string,
+  actions?: NotificationAction[],
+): NotificationType => {
+  return {
+    message,
+    actions,
+    title,
+    type: NotificationSeverity.CAUTION,
+  };
+};
+
+export const formatErrorMessage = (message?: ReactNode, error?: unknown) => {
+  if (error && error instanceof Error) {
+    return (
+      <>
+        {message} {error.message}
+      </>
+    );
+  } else {
+    return message;
+  }
 };
--- a/src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
+++ b/src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
@@ -3,7 +3,12 @@ import type {
   NotificationType,
 } from "components/NotificationProvider";
 import type { ValueOf } from "types";
-import { failure } from "components/NotificationProvider";
+import {
+  failure,
+  info,
+  caution,
+  success,
+} from "components/NotificationProvider";
 import { NotificationSeverity } from "components/Notifications";
 import ToastNotification from "./ToastNotification";
 import ToastNotificationList from "./ToastNotificationList";
@@ -28,22 +33,26 @@ interface ToastNotificationHelper {
     message: ReactNode,
     actions?: NotificationAction[],
     title?: string,
+    id?: ToastNotificationType["id"],
   ) => ToastNotificationType;
   info: (
     message: ReactNode,
     title?: string,
     actions?: NotificationAction[],
+    id?: ToastNotificationType["id"],
   ) => ToastNotificationType;
   failure: (
     title: string,
     error: unknown,
     message?: ReactNode,
     actions?: NotificationAction[],
+    id?: ToastNotificationType["id"],
   ) => ToastNotificationType;
   caution: (
     message: ReactNode,
     actions?: NotificationAction[],
     title?: string,
+    id?: ToastNotificationType["id"],
   ) => ToastNotificationType;
   clear: (notification?: ToastNotificationType[]) => void;
   toggleListView: () => void;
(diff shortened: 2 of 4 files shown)


src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
```diff
@@ -217,6 +225,7 @@ const ToastNotificationProvider: FC<PropsWithChildren<Props>> = ({
   };
 
   const clear = (notifications?: ToastNotificationType[]) => {
+    console.log(notifications);
```
edlerd:
```suggestion
```

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
```diff
@@ -3,19 +3,20 @@ import type {
...
 import { createContext, useContext, useEffect, useState } from "react";
 import React from "react";
+import { formatErrorMessage } from "components/NotificationProvider/messageBuilder";
 
 const HIDE_NOTIFICATION_DELAY = 5_000;
 
 export type ToastNotificationType = NotificationType & {
   timestamp?: ReactNode;
   id: string;
+  entityId?: string | number;
```
edlerd: I don't understand why the notification needs to have an entityId. The notification already has an ID that is generated internally that should allow targeting individual messages.

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
```diff
@@ -262,28 +271,38 @@ const ToastNotificationProvider: FC<PropsWithChildren<Props>> = ({
...
+    failure: (title, error, message, actions, entityId) =>
+      addNotification({
+        message: formatErrorMessage(message, error),
+        title,
+        error,
+        actions,
+        entityId,
+        type: NotificationSeverity.NEGATIVE,
+      }),
+    info: (message, title, actions, entityId) =>
```
edlerd: In my opinion, we should extend the failure helper with anything we need in here in `messageBuilder.tsx`. The other severity types like `info`, `success` and alike should rely on helpers over there as well instead of building the message ad hoc.

src/components/NotificationProvider/messageBuilder.tsx
```diff
@@ -37,15 +37,20 @@ export const failure = (
...
-      ) : (
-        message
-      ),
+    message: formatErrorMessage(message, error),
     title,
     type: NotificationSeverity.NEGATIVE,
   };
 };
+
+export const formatErrorMessage = (message?: ReactNode, error?: unknown) => {
```
edlerd, line 66: Good idea to extract this function.

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
rumanowskak: I did actually try using the built-in ID attribute, however this ID is an HTML element ID that is used for positioning an element - i didn't want to override this and break positioning logic

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
rumanowskak: I didn't do this initially since it would mean that I would have to add entityId to the base Notification as well - however, if we can do this, then I'm happy to refactor this section to use the helpers (and extend the helpers as needed)

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
edlerd: I meant the other way round. Rely on the ID the notification generates for itself internally, not overriding it.

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
edlerd: I think if we have good reason to have an entity ID on the notification, it should be added to the base notification. Though I currently don't understand yet why a notification should have an entity ID. If it turns out to be needed, maybe this should even be an entity ID list, as a notification could relate to multiple entities.

edlerd: Maybe we and others concerned can have a quick chat on this?

src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx
rumanowskak: For us it's not the greatest solution to be honest. We get ids from backend. So when we want to get rid of notifications and use dismiss endpoint from our backend we have to use the id provided by it. With your solution, we would have to map the generated ids, for our backend ids, which in our case, would be just a lot of extra boilerplate and would be easier for us if we could just pass an id.

edlerd: Thanks for explaining, I still don't fully understand. Why do you dismiss messages based on API responses?

rumanowskak: We get notifications from backend and then we create toast notification for each backend notification, when we dismiss notification we want to dismiss it also from our database by dismiss endpoint in backend, to do this we need id of notification that we got firstly from backend.

rumanowskak: this is what our response looks like - notifications are stored in a table in the back-end, and each has an ID 
image
We receive all the notifications as a list, then we parse that list and add each notification to the ToastNotificationProvider ("category" is the severity of the notification).
We can dismiss notifications by making a POST call to our back-end, with the ID of the notification we want to dismiss:
image
Then the notification won't show up in the response again if we ask for the notification list.
So if we keep the ID we receive from the backend in the state of the ToastNotificationProvider, then we can easily access it again when the `onDismiss` callback is called.

edlerd: This is the id of the notification itself, right? I was wrongly assuming `entityId` to be the id of a referenced entity.

If you have externally generated ids for notifications, I think we should allow overriding the `id` field of the notification. When set, this will opt out of the id generation in [ToastNotificationProvider.tsx](https://github.com/canonical/react-components/blob/main/src/components/Notifications/ToastNotification/ToastNotificationProvider.tsx#L207).

ndv99: I did try this in my other PR, but this was when I encountered the issue of IDs being used as HTML element IDs: https://github.com/canonical/react-components/blob/main/src/components/Notifications/ToastNotification/ToastNotificationList.tsx#L63-L85

Specifically line 73, where it's grabbing an element by ID using `notification.id`. I'm not sure about other back-ends, but for MAAS usually our IDs from the database are just whole numbers, which doesn't seem like a good candidate for a unique HTML element ID

edlerd: I think numbers are fine for html ids. Maybe you can prefix them with a string like "Notification-123" to ensure no other id clashes with the number.

ndv99: That sounds sensible to me. So we make `id` override-able, and prefix it with `notification-` before assigning it to an element, as well as when accessing elements by ID

edlerd: > That sounds sensible to me. So we make `id` override-able, and prefix it with `notification-` before assigning it to an element, as well as when accessing elements by ID

Great :)

I think the prefixing might be best to keep outside react components and in the client. So the ID is used as is by the `NotificationProvider`. What do you think?

ndv99: Let me make sure I understand - so the provider shouldn't do anything with the ID, except for generating it if one has been provided. And if one *is* provided, then it should be used as-is, right?

I think this makes sense to me, and then we let each project do prefixing however it suits them? Or perhaps provide a helper function for this as well?

edlerd: Yes, exactly 100% agree :)

ndv99:
> Thanks for the iteration, code looks good to me and very clean now, I like it. I think the simplicity of the result was worth the slight back and forth. :)

Thanks for all your feedback as well! I agree, I think we've landed on a good solution here (:

PR #1243, opened 2025-08-13
feat(SidePanel): side panel animates on open and close

ethanashaw:
## Done

- Side panel has animation with the new `isAnimated` prop. Use the `isOpen` prop instead of conditionally rendering `SidePanel`.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open and close the side panel

### Percy steps

- No visual changes expected

### DIFF
--- a/src/components/SidePanel/SidePanel.stories.tsx
+++ b/src/components/SidePanel/SidePanel.stories.tsx
@@ -261,7 +261,7 @@ export const Animated: Story = {
     parentId: "l-application-default",
     pinned: false,
     loading: false,
-    overlay: true,
+    overlay: false,
     width: "",
     isAnimated: true,
   },
--- a/src/components/SidePanel/SidePanel.tsx
+++ b/src/components/SidePanel/SidePanel.tsx
@@ -34,7 +34,7 @@ export type Props = {
   hasError?: boolean;
 
   /**
-   * Whether or not the side panel animates on open or close. This requires the side panel to be controlled by the `isOpen` prop instead of conditional rendering.
+   * Whether the side panel animates on open and close. This requires the side panel to be controlled by the `isOpen` prop instead of conditional rendering. If you use `overlay: true`, you must also include `overflow: hidden` on the document body.
    */
   isAnimated?: boolean;
 
@@ -132,9 +132,9 @@ const SidePanelComponent = ({
   }
 
   // Pinned side panels don't get animation in Vanilla
-  isAnimated &&= !pinned;
+  const isAnimatedFinal = isAnimated && !pinned;
 
-  if (!isOpen && !(isAnimated && hiding)) {
+  if (!isOpen && !(isAnimatedFinal && hiding)) {
     return null;
   }
 
@@ -144,7 +144,7 @@ const SidePanelComponent = ({
         <AppAside
           className={classnames("side-panel", className, {
             "is-overlay": overlay,
-            "slide-in": isAnimated,
+            "slide-in": isAnimatedFinal,
           })}
           collapsed={!isOpen}
           aria-label="Side panel"


src/components/SidePanel/SidePanel.tsx
```diff
@@ -53,6 +53,11 @@ export type Props = {
    */
   pinned?: boolean;
 
+  /**
+   * Whether to show the side panel or not.
+   */
+  show?: boolean;
```
edlerd: I think the only external interface change should be a new `isAnimated` prop.

I think we should not enable animations for everyone by default. Maybe better to keep animations disabled by default. Though this could be discussed as to what is the best default.

src/components/SidePanel/SidePanel.stories.tsx
edlerd, line 1: I think we should not change the stories to ensure backwards compatibility. Maybe add a new story for showing how it looks like with animations enabled?

src/components/SidePanel/SidePanel.tsx
```diff
@@ -108,20 +113,39 @@ const SidePanelComponent = ({
...
+
   return (
     <>
       {createPortal(
         <AppAside
           className={classnames("side-panel", className, {
             "is-overlay": overlay,
+            "slide-in": animated,
           })}
+          collapsed={!show}
```
edlerd: One issue with using only this is that if you have many side panels on a page that can be opened, they will all be in the DOM even though all are closed. I think the approach in `ToastNotification` is a little better of wrapping the rendered things in `Animate`, which in turn removes them from the DOM if they are not shown. This ensures the DOM is clean of (possibly many) side panels that are not opened.

edlerd (commented):
Thanks for raising the PR and initiating the work for this. Visually, this is really nice.

Some concerns on the approach and api/interface change below.

src/components/SidePanel/SidePanel.tsx
ethanashaw: @edlerd, the side panel is removed from the DOM through the `onTransitionEnd` prop, since the class added by `collapsed` (`is-collapsed`) uses a transition. It's a more declarative version of the approach in `ToastNotification`, but using a transition instead of an animation (to use the actual Vanilla styles and to avoid flickering when the animation ends).

src/components/SidePanel/SidePanel.tsx
ethanashaw: The `show` prop is needed to control rendering or it can't be animated on the way out, similar to `ToastNotification`. If we use the existing conditional rendering, it would immediately unmount.

Maybe I need to explain this in the description for the prop, because it kind of serves two purposes. If you don't assign a value to it, and instead use conditional rendering, you will not get any animation as before. If you do use it, you will get animation. We could add an `isAnimated` prop to separate this logic, as long as we explain that you have to use `show` too. What do you think?

edlerd: Yeah, maybe splitting is a better approach.

src/components/SidePanel/SidePanel.tsx
edlerd: I checked the demo server previously, and maybe I was just unlucky to pick the [pinned story](https://react-components-1243.demos.haus/?path=/story/components-sidepanel--pinned). When inspecting the DOM for the panel and closing it, it stays in the DOM.

Though you are right, all the other stories do unmount it from the DOM, so this concern is probably to be resolved. Well, at least reduced to the question of why the pinned story does not unmount.

src/components/SidePanel/SidePanel.tsx
edlerd: I think this should be wider communicated once it lands. We might want to refactor use cases of the component to use the `show` prop instead of conditional rendering in all the usages in projects. Maybe a better name for the prop is `isOpen`?

edlerd: And then we can have the second `isAnimated` prop, to drive that condition. I agree in the prop docs of animated we should mention that this only works with the other prop in tandem and not with conditional rendering that was previously used.

src/components/SidePanel/SidePanel.tsx
ethanashaw: It looks like pinned side panels don't get any animation in Vanilla, so it was failing to trigger `onTransitionEnd`. For this reason, I didn't add a story for pinned side panels with animation, but you can still mess with the props in the storybook to ensure it unmounts.

src/components/SidePanel/SidePanel.tsx
```diff
@@ -33,6 +33,16 @@ export type Props = {
    */
   hasError?: boolean;
 
+  /**
+   * Whether or not the side panel animates on open or close. This requires the side panel to be controlled by the `isOpen` prop instead of conditional rendering.
```
edlerd:
```suggestion
   * Whether the side panel animates on open or close. This requires the side panel to be controlled by the `isOpen` prop instead of conditional rendering.
```
A bit more concise :)

src/components/SidePanel/SidePanel.tsx
```diff
@@ -108,20 +118,40 @@ const SidePanelComponent = ({
...
+
   const container = document.getElementById(parentId) || document.body;
 
+  if (isOpen !== previousIsOpen) {
+    setPreviousIsOpen(isOpen);
+    setHiding(true);
+  }
+
+  // Pinned side panels don't get animation in Vanilla
+  isAnimated &&= !pinned;
```
edlerd: React props should be read-only. Better define a new variable.

src/components/SidePanel/SidePanel.stories.tsx
```diff
@@ -255,3 +253,17 @@ export const Wide: Story = {
...
+    hasError: false,
+    parentId: "l-application-default",
+    pinned: false,
+    loading: false,
+    overlay: true,
+    width: "",
+    isAnimated: true,
+  },
+  render: StoryExample,
+};
```
edlerd: I think there is the expectation of the caller to have `overflow:hidden` on the document body, so this doesn't come with a scrollbar during the animation. Though, I am not really sure how and where to put that information. Maybe on the is animated prop doc block?

edlerd (commented):
QA seems all good now. Some nitpicks on the code, then this should be good to merge from my side.

Issue #1245, opened 2025-08-14
The responsive behaviour of the `MainTable` wrapped inside a `ScrollableTable` is not equivalent to the default case


lorumic:
When wrapping a `MainTable` with the `responsive` prop inside a `ScrollableTable`, we get an inconsistent behaviour than the default one (i.e. the one that the `MainTable` has when it's not wrapped inside the `ScrollableTable`). Essentially, to use the space effectively, we should allow more than one item to sit in a row, when the width is enough.

Check these screenshots to see the difference in practice:

- `MainTable` with `responsive` prop not wrapped in a `ScrollableTable` (example from [Storybook](https://canonical.github.io/react-components/?path=/story/components-maintable--responsive)):

<img width="1320" height="387" alt="Image" src="https://github.com/user-attachments/assets/d31c776a-df1b-44d3-aea3-cade3eeeb454" />

- `MainTable` with `responsive` prop wrapped in a `ScrollableTable` (example from Storybook modified locally):

<img width="1320" height="581" alt="Image" src="https://github.com/user-attachments/assets/9c2c95d9-188e-4e52-8683-085e188769d2" />

- Practical example of how the scrollable table with a responsive main table inside should look like (taken from Anbox dashboard):

<img width="1194" height="960" alt="Image" src="https://github.com/user-attachments/assets/af35982a-4293-46b3-9d77-eb0bd1a2e0ba" />

PR #1247, opened 2025-08-15
fix: The responsive behaviour of the `MainTable` wrapped inside a `ScrollableTable` is not equivalent to the default case

lorumic:
## Done

- Fixes linked issue (#1245).
- Fixes a bug that causes padding to be added twice in `getElementAbsoluteHeight` helper function.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Go to "ScrollableTable" -> "Responsive" example in Storybook.
- Resize the window width and check that the issue described in #1245 is solved.

### Percy steps

The only visual change expected is the capture of two snapshots of the "Responsive" story for the `ScrollableTable`, at 768px and 1280px width.

## Fixes

Fixes: #1245

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -59,8 +59,7 @@ export const getElementAbsoluteHeight = (element: HTMLElement) => {
   }
   const style = window.getComputedStyle(element);
   const margin = toFloat(style.marginTop) + toFloat(style.marginBottom);
-  const padding = toFloat(style.paddingTop) + toFloat(style.paddingBottom);
-  return element.offsetHeight + margin + padding + 1;
+  return element.offsetHeight + margin + 1;
 };
 
 export const getAbsoluteHeightBelowById = (belowId: string): number => {
--- a/src/components/ScrollableTable/ScrollableTable.scss
+++ b/src/components/ScrollableTable/ScrollableTable.scss
@@ -45,12 +45,12 @@
         display: block;
       }
 
-      tbody tr {
-        display: block;
+      tbody {
+        display: grid;
       }
 
-      tbody tr:first-child {
-        margin-top: $spv--x-large;
+      tbody tr {
+        display: block;
       }
     }
   }
--- a/src/components/ScrollableTable/ScrollableTable.stories.tsx
+++ b/src/components/ScrollableTable/ScrollableTable.stories.tsx
@@ -104,5 +104,10 @@ export const Responsive: Story = {
         </>
       ),
     },
+    // Percy settings to capture at multiple widths
+    percy: {
+      responsiveSnapshotCapture: true,
+      widths: [768, 1280],
+    },
   },
 };


src/utils.ts
```diff
@@ -59,8 +59,7 @@ export const getElementAbsoluteHeight = (element: HTMLElement) => {
   }
   const style = window.getComputedStyle(element);
   const margin = toFloat(style.marginTop) + toFloat(style.marginBottom);
-  const padding = toFloat(style.paddingTop) + toFloat(style.paddingBottom);
-  return element.offsetHeight + margin + padding + 1;
+  return element.offsetHeight + margin + 1;
```
edlerd, line 62: Thanks for fixing this as a drive by!

edlerd (approved):
Code and QA LGTM. Thanks for finding the fix for this one!

lorumic:
> Maybe we could use a [story viewport configuration](https://storybook.js.org/docs/essentials/viewport#defining-the-viewport-for-a-story) to force that specific story to render in a smaller viewport

Thanks @jmuzina, that's exactly what I was looking for! The only thing that I don't like too much is that it basically appends a query parameter to the URL, which is kept for any other story you navigate to after.

In other words, when you open this new "Responsive (tablet)" story, and then navigate to another one (where maybe you don't want to see the tablet preview), you need to manually reset the viewport - either with the "Reset viewport" button or by removing the query parameter from the URL. It would have been nicer if instead it would have reset the viewport automatically upon navigation.

Anyway, for now I have added it, if it turns out to be an annoying behaviour I'm happy to revert :)

jmuzina:
> In other words, when you open this new "Responsive (tablet)" story, and then navigate to another one (where maybe you don't want to see the tablet preview), you need to manually reset the viewport - either with the "Reset viewport" button or by removing the query parameter from the URL. It would have been nicer if instead it would have reset the viewport automatically upon navigation.

I found [this comment on storybook](https://github.com/storybookjs/storybook/issues/27073#issuecomment-2225329662) which suggests you should be able to define the default viewport for default export (`meta`) as "reset"

Results:


https://github.com/user-attachments/assets/98fc5870-db42-4b81-92eb-a49c102946e7

lorumic:
> I found [this comment on storybook](https://github.com/storybookjs/storybook/issues/27073#issuecomment-2225329662) which suggests you should be able to define the default viewport for default export (`meta`) as "reset"

Excellent, thanks Julie! :100:

jmuzina:
hmm. The viewport in the browser is tablet-sized now, but the percy test is still being taken at desktop width - so we're still not solving the root problem here.

We can use [percy config options](https://www.browserstack.com/docs/percy/integrate/storybook#enable-responsive-dom-capture-for-storybook-components) to change the widths the story is tested at. Maybe you could get rid of the "Responsive (tablet)" specific story, and just capture the "Responsive" story at two widths, like this?

```tsx
export const Responsive: Story = {
  args: {
    belowIds: ["footer"],
  },
  render: (args) => StoryExample({ ...args, responsive: true }),
  parameters: {
    docs: {
      page: () => (
        <>
          <Title />
          <Subtitle />
          <Description />
        </>
      ),
    },
    // The viewport setting here is optional, it depends what you want the default view to be for the story on the browser
    viewport: { defaultViewport: "tablet" },
   // Capture at multiple widths
    percy: {
      responsiveSnapshotCapture: true,
      widths: [768, 1280],
    },
  },
};
```

lorumic:
> We can use [percy config options](https://www.browserstack.com/docs/percy/integrate/storybook#enable-responsive-dom-capture-for-storybook-components) to change the widths the story is tested at. Maybe you could get rid of the "Responsive (tablet)" specific story, and just capture the "Responsive" story at two widths, like this?

That's even better, added it :+1:

jmuzina (approved):
good to go - thanks for the fix and coming along with me on the percy configuration journey there 🤠

PR #1251, opened 2025-08-27
fix(`LoginPageLayout`): no longer sets theme to `is-paper`

ethanashaw:
## Done

`LoginPageLayout` no longer updates the body theme with an effect, leaving it to the responsibility of the library consumer. The current behavior updates the body to use `is-paper`, but this unexpected, difficult to trace, and often undesirable. For example, we have to include this effect in our app to undo the behavior, otherwise dark mode doesn't work:
```ts
useLayoutEffect(() => {
  document.body.classList.remove("is-paper");
}, []);
```

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### Percy steps

- `LoginPageLayout` has a plain background instead of a paper background.

### DIFF
--- a/src/components/LoginPageLayout/LoginPageLayout.test.tsx
+++ b/src/components/LoginPageLayout/LoginPageLayout.test.tsx
@@ -1,8 +1,7 @@
-import React, { useLayoutEffect, useState } from "react";
+import React from "react";
 import { render, screen, within } from "@testing-library/react";
 
 import LoginPageLayout from "./LoginPageLayout";
-import userEvent from "@testing-library/user-event";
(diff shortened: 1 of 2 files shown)


edlerd (approved):
Code LGTM.

I also agree with the changes, but this is a breaking change. So maybe should be released as a new major version. There are quite a few projects using it as to [this search](https://github.com/search?q=org%3Acanonical+%22%3CLoginPageLayout%22&type=code).

PR #1258, opened 2025-09-25
feat(MultiSelect): implement help

kimanhou:
## Done

- feat(MultiSelect): implement help as a helper text underneath the component

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Check that MultiSelect with help displays a helper text
- Check that MultiSelect without help remains unchanged

### Percy steps

### DIFF
--- a/src/components/MultiSelect/MultiSelect.stories.tsx
+++ b/src/components/MultiSelect/MultiSelect.stories.tsx
@@ -131,3 +131,14 @@ export const Disabled: Story = {
     disabled: true,
   },
 };
+
+export const HelpText: Story = {
+  args: {
+    ...CondensedExample.args,
+    help: (
+      <span>
+        This is a help text, that should appear underneath the component.
+      </span>
+    ),
+  },
+};
--- a/src/components/MultiSelect/MultiSelect.test.tsx
+++ b/src/components/MultiSelect/MultiSelect.test.tsx
@@ -321,3 +321,40 @@ it("can render without sorting selected items first", async () => {
       ),
   );
 });
+
+it("can render help", async () => {
+  const { container } = render(
+    <MultiSelect
+      items={items}
+      selectedItems={[items[1]]}
+      help={<span>This is a help text</span>}
+    />,
+  );
+
+  const helpTextList = container.querySelectorAll(".p-form-help-text");
+  expect(helpTextList).toHaveLength(1);
+  const helpText = helpTextList[0];
+  expect(helpText).toBeVisible();
+});
+
+it("doesn't render help", async () => {
+  const { container } = render(
+    <MultiSelect items={items} selectedItems={[items[1]]} />,
+  );
+  expect(container.querySelectorAll(".p-form-help-text")).toHaveLength(0);
+});
+
+it("can add additional classes to help", () => {
+  const { container } = render(
+    <MultiSelect
+      items={items}
+      selectedItems={[items[1]]}
+      help={<span>This is a help text</span>}
+      helpClassName="additional-help-text-class"
+    />,
+  );
+  expect(container.querySelectorAll(".p-form-help-text")).toHaveLength(1);
+  expect(container.querySelector(".p-form-help-text")).toHaveClass(
+    "additional-help-text-class",
+  );
+});
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -1,4 +1,5 @@
 import type { ReactNode } from "react";
+import classNames from "classnames";
 import React, { useEffect, useId, useMemo, useRef, useState } from "react";
 
 import "./MultiSelect.scss";
@@ -16,7 +17,8 @@ export type MultiSelectProps = {
   disabled?: boolean;
   error?: string;
   selectedItems?: MultiSelectItem[];
-  help?: string;
+  help?: ReactNode;
+  helpClassName?: string;
   label?: string | null;
   listSelected?: boolean;
   onDeselectItem?: (item: MultiSelectItem) => void;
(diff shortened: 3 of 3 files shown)


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -16,7 +16,7 @@ export type MultiSelectProps = {
   disabled?: boolean;
   error?: string;
   selectedItems?: MultiSelectItem[];
-  help?: string;
+  help?: ReactNode;
```
kimanhou, line 20: I changed the type from `string` to `ReactNode` to match the other components, that also have help as props.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -268,102 +269,105 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
...
+          header={dropdownHeader}
+          updateItems={updateItems}
+          onSelectItem={onSelectItem}
+          onDeselectItem={onDeselectItem}
+          footer={footer}
+          sortFn={isSortedAlphabetically ? sortAlphabetically : () => 0}
+          hasSelectedItemsFirst={hasSelectedItemsFirst}
+        />
+      </ContextualMenu>
+      {help ? <p className="p-form-help-text">{help}</p> : null}
```
kimanhou: I would have preferred the helper text to appear closer to the component. But I left it as is to match with Field (`margin-top: -0.5rem`)

anusha-c18:
```suggestion
      {help && (<p className="p-form-help-text">{help}</p>)}
```

anusha-c18: This could also use a `helpId`.

src/components/MultiSelect/MultiSelect.tsx
anusha-c18, line 20: Could we also pass a `helpClassName` prop for more flexibility?

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -131,3 +131,14 @@ export const Disabled: Story = {
     disabled: true,
   },
 };
+
+export const WithHelp: Story = {
```
anusha-c18:
```suggestion
export const HelpText: Story = {
```

src/components/MultiSelect/MultiSelect.tsx
anusha-c18, line 1: We should also add a test for the help text.

src/components/MultiSelect/MultiSelect.tsx
kimanhou: Any className could be passed onto the ReactNode. Is there a use case where this couldn't work without a helpClassName ?

anusha-c18: >  Is there a use case where this couldn't work without a helpClassName ?

Yes, if I am only passing a string.

edlerd: I think we should not add support for custom classes on the help text. If you really have to customize, then you can pass in a help value as react node. It seems better to not extend the interface too much. So we keep the application designs aligned and avoid too much customization.

src/components/MultiSelect/MultiSelect.tsx
edlerd: I think the standard spacing is okay, we should avoid customizations and follow the vanilla standards as much as possible.

src/components/MultiSelect/MultiSelect.tsx
anusha-c18: We’ve consistently exposed a `helpClassName` prop in components like `Field`, `Input`, and `Switch`.

I don’t see why `MultiSelect` should be treated differently. Adding it here keeps the API consistent and avoids introducing a one-off rule where this component requires a `ReactNode` instead.

kimanhou: Don't Field, Input and Switch also require a `ReactNode` as help ?

edlerd: > We’ve consistently exposed a `helpClassName` prop in components like `Field`, `Input`, and `Switch`.

Oh true, I wasn't aware of this. I see and indeed we can't do the same with a wrapper on the `help` value using it as a react node. For instance the `TextInput` story on input is passing in the `u-no-margin--bottom` to the help class that won't work when applied to a wrapper.

So yes, I think we should also add the `helpClassName` and forward it to the help element to allow directly manipulating it.

anusha-c18: > Don't Field, Input and Switch also require a ReactNode as help ?

Yes, exactly :) 
We have the flexibility to customize it with custom components but also pass a `helpClassName` when you're passing a string and on the parent paragraph element when you're passing a node.

src/components/MultiSelect/MultiSelect.test.tsx
```diff
@@ -321,3 +321,36 @@ it("can render without sorting selected items first", async () => {
...
+
+it("can render help", async () => {
+  const { container } = render(
+    <MultiSelect
+      items={items}
+      selectedItems={[items[1]]}
+      help={<span>This is a help text</span>}
+    />,
+  );
+  expect(container.querySelectorAll(".p-form-help-text")).toHaveLength(1);
```
anusha-c18: I think for this specific test, we should assert that the help text we passed is actually visible, not just that a node with the `.p-form-help-text` class exists.

anusha-c18 (changes requested):
Thanks for addressing the changes. One last suggestion, and we're good to merge!

anusha-c18 (approved):
Thanks for making the changes, LGTM 👍

PR #1269, opened 2025-10-17
feat(Modal): assign initial focus

ethanashaw:
## Done

`Modal` will focus on the close button instead of choosing the first focusable element.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Ensure that `Modal` focuses on close button, or the modal itself if there is no close button.

### Percy steps

- One new test

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -1,6 +1,6 @@
 import classNames from "classnames";
+import React, { useId, useRef, useEffect } from "react";
 import type { HTMLProps, ReactNode, RefObject } from "react";
-import React, { useEffect, useId, useRef } from "react";
 import { ClassName, PropsWithSpread } from "types";
 
 export type Props = PropsWithSpread<
@@ -21,6 +21,10 @@ export type Props = PropsWithSpread<
      * Function to handle closing the modal.
      */
     close?: () => void | null;
+    /**
+     * The element that will be focused upon opening the modal.
+     */
+    focusRef?: RefObject<HTMLElement | null>;
     /**
      * The title of the modal.
      */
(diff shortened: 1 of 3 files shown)


edlerd:
It was me changing the modal focus on open to the first focusable element instead of the close button. My assumption was that the first element is more useful as a user is morel likely interacting with that element. Say an input in the modal is more useful to have focus. If the user were to close the modal, they could simply tab to the close button with very few tabs, assuming modals don't have a lot of content. And our modal traps the focus, so it won't be too many tab presses to reach the close button.

Why do you think focusing the close button is preferred @ethanashaw ?

src/components/Modal/Modal.tsx
```diff
@@ -97,30 +103,26 @@ export const Modal = ({
...
       focusableElementSelectors,
     );
-    let focusIndex = 0;
-    // when the close button is rendered, focus on the 2nd content element and not the close btn.
-    if (hasCloseButton && focusableModalElements.current.length > 1) {
-      focusIndex = 1;
-    }
-    focusableModalElements.current[focusIndex]?.focus({ preventScroll: true });
-  }, [hasCloseButton]);
+  }, [focusRef]);
```
edlerd, line 118:
```suggestion
  }, [focusRef, closeButtonRef, modalRef]);
```
Should we add all dependencies here?

ethanashaw: I think there's no need, since `closeButtonRef` and `modalRef` won't change between renders, or trigger re-renders. `focusRef` is only needed because the parent component could pass a different ref on a render.

ethanashaw:
@edlerd, do you want this to be a breaking change or just a regular `feat`?

PR #1278, opened 2025-11-21
fix(tooltip): fix wrong initial and focus tooltip positions in followMouse mode

bmichalowski:
## Done

The tooltip did not handle the initial position correctly in followMouse mode.
There was also an issue with listening to mouse move events only when the tooltip was open.

## QA

tooltip_follow-mouse-mode_missing-tooltip-on-focus
tooltip_follow-mouse-mode_prev-posirion-bug

### QA steps

Bug 1: Tooltip is not visible on keyboard focus in followMouse mode
1. Given the tooltip is in followMouse mode
2. And the user has focused the element with the tooltip using the TAB key
3. When the element receives focus
4. Then the tooltip should be visible

Bug 2: Tooltip appears at previous cursor location instead of current
1. Given the tooltip is in followMouse mode
2. And the user hovers over an element on the left side
3. And then moves the cursor away from the element
4. When the user hovers over an element on the right side
5. Then the tooltip should appear next to the current cursor position
6. And not at the previous location
         

### Percy steps

No visual changes expected


src/components/Tooltip/Tooltip.tsx
```diff
@@ -307,6 +307,18 @@ const Tooltip = ({
...
 
+  const handleFocus = () => {
+  if (followMouse) {
+    if (wrapperRef.current) {
+      // set initial position for the tooltip
+      setPositionStyle(
+        getPositionStyle(adjustedPosition, wrapperRef.current)
+      );
+    }
+  }
```
edlerd:
```suggestion
    if (followMouse && wrapperRef.current) {
      // set initial position for the tooltip
      setPositionStyle(getPositionStyle(adjustedPosition, wrapperRef.current));
    }
    openPortal();
```
A bit simpler and please use correct spacing and formatting.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -253,7 +253,7 @@ const Tooltip = ({
     mouseHandler,
     "mousemove",
     true,
-    followMouse && isOpen,
+    followMouse,
```
edlerd:
```suggestion
  useListener(wrapperRef.current, mouseHandler, "mousemove", true, followMouse);
```
This whole block should be in a single line now.

bartaz:
@bmichalowski Are you interested in continuing working on the issue?

bmichalowski:
@bartaz Yes, I'd be happy to finish it. However, I need a helping hand with Percey. Visually, nothing changes, so I think it should be marked as non-required. However, it starts and crashes during execution.

PR #1291, opened 2026-01-09
fix(modal): fix bug where the list of focus-able elements was not re-computed when the state of certain children has changed [AC-4105]

Stefan3002:
## Done

- Fix bug where the list of focus-able elements was not re-computed when the state of certain children has changed
- This was done by using a Mutation Observer that checks for specific changes in the following attributes in the children of the modal : ``` attributeFilter: ["disabled", "tabindex", "contentEditable"], ```

## QA

- Just use the modal component with a disabled button and start tabbing: you must not reach the button
- Now enable the button programatically and start tabbing: you should reach the enabled button (previously impossible)
- For example: Create a disabled button and an input field. If you fill the input field with a specific string, you must enable the button. Now you should also be able to tab to the button 
- Simple example:

image
image

## Fixes

Fixes: # [AC-4105](https://warthogs.atlassian.net/browse/AC-4105)

[AC-4105]: https://warthogs.atlassian.net/browse/AC-4105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -67,14 +67,14 @@ export const Modal = ({
 
   const modalRef: RefObject<HTMLDivElement> = useRef(null);
   const closeButtonRef: RefObject<HTMLButtonElement> = useRef(null);
-  const focusableModalElements = useRef(null);
   const handleTabKey = (event: React.KeyboardEvent<HTMLDivElement>) => {
-    if (focusableModalElements.current.length > 0) {
-      const firstElement = focusableModalElements.current[0];
+    const focusableModalElements = modalRef.current.querySelectorAll(
+      focusableElementSelectors,
+    );
+    if (focusableModalElements.length > 0) {
+      const firstElement = focusableModalElements[0];
       const lastElement =
-        focusableModalElements.current[
-          focusableModalElements.current.length - 1
-        ];
+        focusableModalElements[focusableModalElements.length - 1];
 
       if (!event.shiftKey && document.activeElement === lastElement) {
         (firstElement as HTMLElement).focus();
@@ -111,10 +111,6 @@ export const Modal = ({
     } else {
       modalRef.current.focus();
     }
-
-    focusableModalElements.current = modalRef.current.querySelectorAll(
-      focusableElementSelectors,
-    );
   }, [focusRef]);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


jmuzina:
QA Demo will be available shortly [here](https://react-components-1291.demos.haus/) (@Stefan3002 check your email for a GitHub team invite link, which will allow demos to be automatically spun up for you in future)

src/components/Modal/Modal.tsx
```diff
@@ -112,11 +118,24 @@ export const Modal = ({
...
     }
 
-    focusableModalElements.current = modalRef.current.querySelectorAll(
-      focusableElementSelectors,
-    );
+    updateFocusable();
   }, [focusRef]);
 
+  useEffect(() => {
+    const observer = new MutationObserver(() => {
```
edlerd: While this works, it adds a lot of complexity. I think we might better keep it simple and remove the `focusableModalElements` cache. Instead, compile the list of `focusableModalElements` on the fly whenever `handleTabKey` executes. Wdyt?

src/components/Modal/Modal.tsx
edlerd, line 1: Please, can you add a test to `Modal.test.tsx` with a reproducer of the bug that fails on main and is fixed with your changeset?

src/components/Modal/Modal.tsx
Stefan3002: Sure that sounds good, but I am thinking about the performance implications: recomputing it everytime we press tab vs recomputing it only when something has actually changed. 

I am fine either way, but leaning towards using the observer 😁

edlerd: I don't think performance is a reasonable concern. We seem to be talking about 7000 operations per millisecond with query selectors, according to [this blog](https://gomakethings.com/javascript-selector-performance/).

Stefan3002: Sure, updated :smile:

src/components/Modal/Modal.tsx
Stefan3002: Done!

Stefan3002: I used a native button because the ```Button``` component does not set disabled=true and thus is not removed from the focus list, which is exactly what we wanted to begin with :smile:

PR #1294, opened 2026-01-13
fix(modal): fix bug where the action button would set disabled to true

Stefan3002:
## Done

- Removed disabled property from the action button
- This makes it align with the regular ```Button``` by not setting disabled=true but rather using ```aria-disabled```

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Use a ```ActionButton``` component and disable it. It should still be focus-able via keyboard

### DIFF
--- a/src/components/ActionButton/ActionButton.test.tsx
+++ b/src/components/ActionButton/ActionButton.test.tsx
@@ -1,5 +1,5 @@
 import React, { act } from "react";
-import { render, screen } from "@testing-library/react";
+import { fireEvent, render, screen } from "@testing-library/react";
 
 import ActionButton, {
   Label,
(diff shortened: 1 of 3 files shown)


src/components/ActionButton/ActionButton.tsx
```diff
@@ -165,8 +165,8 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
-      disabled={disabled === null ? showLoader : disabled}
       ref={ref}
+      aria-disabled={disabled || undefined}
```
edlerd: We should also disable it when loading.

Stefan3002: Yes, added :smile:

src/components/ActionButton/ActionButton.tsx
```diff
@@ -165,8 +173,9 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
-      disabled={disabled === null ? showLoader : disabled}
       ref={ref}
+      onClick={disabled && !loading ? onClickDisabled : onClick}
+      aria-disabled={disabled || loading || undefined}
```
edlerd:
```suggestion
      aria-disabled={disabled === null ? showLoader : disabled}
```
Can we keep the old logic here? It has a slightly different behavior when the parent sets the `disabled` prop.

src/components/ActionButton/ActionButton.tsx
```diff
@@ -165,8 +173,9 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
-      disabled={disabled === null ? showLoader : disabled}
       ref={ref}
+      onClick={disabled && !loading ? onClickDisabled : onClick}
```
edlerd: I think you also want to prevent onclick from firing if loading is true. Maybe this should be in sync with the aria disabled below, so we'd best introduce a const for it.

Stefan3002: Oops, got the ! wrong :sweat_smile:

Stefan3002: How about now?

src/components/ActionButton/ActionButton.tsx
```diff
@@ -155,8 +161,11 @@ const ActionButton = ({
     },
   );
   const showIcon = showLoader || showSuccess;
+  const isReady = !disabled && !loading;
```
edlerd:
```suggestion
  const isDisabled = !disabled && !loading;
```

edlerd: I am not sure if naming is better, as we then have "isDisabled" and "disabled". But isReady seems also confusing.

src/components/ActionButton/ActionButton.tsx
```diff
@@ -165,8 +174,9 @@ const ActionButton = ({
   return (
     <button
       className={buttonClasses}
-      disabled={disabled === null ? showLoader : disabled}
```
edlerd, line 168: Let's preserve this exact logic. Slightly changing it above with the new "isReady" const.

edlerd (approved):
Thanks for the adjustments. Code LGTM.

Tests and snapshots need some iteration. and we should do QA with the demo server up.

src/components/ActionButton/ActionButton.tsx
Stefan3002: Done

src/components/ActionButton/ActionButton.test.tsx
```diff
@@ -72,37 +80,53 @@ describe("ActionButton", () => {
...
     act(() => {
       jest.advanceTimersByTime(1);
     });
     // the ActionButton should finish its loading job
     expect(screen.getByLabelText(Label.SUCCESS)).toBeInTheDocument();
-    expect(screen.getByRole("button")).toBeEnabled();
+    const buttonAfterSuccess = screen.getByRole("button");
+    expect(buttonAfterSuccess).not.toHaveAttribute("aria-disabled", "true");
+    fireEvent.click(buttonAfterSuccess);
+    expect(onClick).toHaveBeenCalledTimes(0);
```
edlerd: Shouldn't this be aria-disabled false and called once? As this is after loading finished.

src/components/ActionButton/ActionButton.test.tsx
```diff
@@ -72,37 +80,53 @@ describe("ActionButton", () => {
...
       jest.advanceTimersByTime(LOADER_MIN_DURATION + 3000);
     });
     rerender(<ActionButton success>Click me</ActionButton>);
     // the ActionButton should finish its loading job at LOADER_MIN_DURATION + 3s
     expect(screen.getByLabelText(Label.SUCCESS)).toBeInTheDocument();
-    expect(screen.getByRole("button")).toBeEnabled();
+    const buttonAfterLongLoad = screen.getByRole("button");
+    expect(buttonAfterLongLoad).not.toHaveAttribute("aria-disabled", "true");
+    fireEvent.click(buttonAfterLongLoad);
+    expect(onClick).toHaveBeenCalledTimes(0);
```
edlerd: Same as above, here the button should be enabled again, and the click handler triggered.

src/components/ActionButton/__snapshots__/ActionButton.test.tsx.snap
```diff
@@ -14,6 +14,7 @@ exports[`ActionButton matches loading snapshot 1`] = `
 
 exports[`ActionButton matches success snapshot 1`] = `
 <button
+  aria-disabled="false"
```
edlerd: Shall we omit this if it is false and have a bit more clean dom?

Stefan3002: I get an error saying it does not match the snapshot if I remove it :shrug:

edlerd: I meant changing the implementation so it never renders false, but undefined instead.

Stefan3002: Maybe like so?
```
aria-disabled={isDisabled || undefined}
```

src/components/ActionButton/ActionButton.test.tsx
Stefan3002: Yes, you are right, somehow overlooked it

src/components/ActionButton/ActionButton.test.tsx
Stefan3002: Fixed

Issue #1248, opened 2025-08-18
`ContextualMenu`: Dropdown content not reachable by keyboard


fasih-mehmood:
## Description:

When a `ContextualMenu` is opened, the dropdown content is not correctly focused when the user presses the Tab key. This prevents users from navigating the dropdown's menu items using the keyboard.

This issue appears to be caused by the component's use of a `Portal`. The `Portal` renders the dropdown content at the end of the body tag to prevent it from being clipped by overflowing parent containers. However, this placement also moves the dropdown out of its original position in the document's accessibility tree, causing it to receive tab focus only after all other tabbable elements on the page.

![Image](https://github.com/user-attachments/assets/e1e28ffb-190f-4d1c-82da-e194911e876e)

## Steps to Reproduce:

1. Open a ContextualMenu on a [page](https://canonical.github.io/react-components/?path=/docs/components-contextualmenu--docs) with other tabbable elements (e.g., buttons, input fields, links).
2. Press the `Tab` key repeatedly.
3. Observe that the focus cycles through all other tabbable elements on the page before finally reaching the dropdown content.

## Expected Behavior:
After the ContextualMenu is opened, pressing `Tab` should move the focus directly to the first tabbable element within the dropdown content (e.g., the first menu item).

## Proposed Solution:
Consider implementing a focus management solution that programmatically moves the focus to the dropdown's first item when the menu is opened and back to the last focused item when the menu is closed.

Safvan-tsy:
Hey @bartaz @fasih-mehmood,

I’m interested in working on this issue. If nobody is currently working on it, please feel free to assign it to me.

PR #1298, opened 2026-01-14
fix(ContextualMenu): Context menu focuses first item on opening

jmuzina:
## Done

- Updated the `ContextualMenu` to focus the first non-disabled child item after opening. Then the user can tab through the items to access them via keyboard.
- Adds a keyboard focus trap to the contextual menu
- Updates the `MultiSelect` component to opt-out of the ContextualMenu autofocusing the first element, as this would make text search difficult (focus should move to the text input, not the first option in that case)

## Future work

There is still more work that can be done to make this component more accessible & WCAG compliant. WCAG specifies that context menus should be navigable with Arrow keys, and Tab should close the context menu and focus the next item after it. This PR at least makes the context menu navigable with Tab - more work will need to be done to comply with [this spec](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) to add more compliant keyboard functionality and to return focus to the element that triggered the menu (see [this issue](https://github.com/canonical/react-components/issues/1305))

I've left these changes for later task as they'll take a lot more work and I wanted to get the component at least minimally accessible in the short-term (a screen reader user raised with WPE that a context menu in 360 is almost impossible to navigate because focus does not move to it on opening and there is no focus trap)

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open the [`ContextualMenu` story](https://react-components-1298.demos.haus/iframe.html?globals=&args=&id=components-contextualmenu--toggle&viewMode=story).
- Use your keyboard to open the dropdown. 
- See that focus has moved to the first link. 
- Use tab and see that focus moves to the next link.
- Keep pressing tab and see that focus only moves within the context menu (focus wraps from the end to the front).
- Press escape and verify the context menu still closes. Focus should return to the button that opened the menu.
- Verify that no functional regressions have happened in the `MultiSelect` component, which consumes `ContextualMenu`.

### Percy steps

- No visual changes expected

## Fixes

Fixes: #1248

### DIFF
--- a/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
+++ b/src/components/ContextualMenu/ContextualMenuDropdown/ContextualMenuDropdown.tsx
@@ -8,7 +8,7 @@ import type { ButtonProps } from "../../Button";
 import type { WindowFitment } from "hooks";
 
 export enum Label {
-  Dropdown = "submenu",
+  Dropdown = "sub",
 }
 
 /**
--- a/src/components/ContextualMenu/ContextualMenuDropdown/__snapshots__/ContextualMenuDropdown.test.tsx.snap
+++ b/src/components/ContextualMenu/ContextualMenuDropdown/__snapshots__/ContextualMenuDropdown.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`ContextualMenuDropdown  renders 1`] = `
 <span
   aria-hidden="true"
-  aria-label="submenu"
+  aria-label="sub"
   class="p-contextual-menu__dropdown"
   data-testid="dropdown"
   role="menu"


edlerd:
Did you check the CustomSelect component with this change? I suspect it might need to opt out as well -- at least if the search variant of it is in place.

fasih-mehmood:
> > I'm not able to close the dropdown using the keyboard
> 
> @fasih-mehmood  Which story are you trying this out on, and which browser? In Chromium on the ContextualMenu toggle story, pressing escape closes the dropdown for me
> 

@jmuzina I tried the ContextualMenu Toggle story on Chrome.

ethanashaw (approved):
You might consider cleaning up some of the effects using this pattern here, although it's not significantly better: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

I think a better option would be to actually assign focus as a part of the event handlers that toggle `isOpen`. I'll leave it up to you, I'm fine with the current code if you'd rather leave it.

src/components/ContextualMenu/ContextualMenu.tsx
```diff
@@ -199,23 +213,116 @@ const ContextualMenu = <L,>({
...
+   * @returns The dropdown element or null if it does not exist.
+   */
+  const getDropdown = useCallback(() => {
+    if (typeof document === "undefined") return null;
+    /**
+     * This is Using `document` instead of refs because `dropdownProps` may include a ref,
+     * while `dropdownId` is unique and controlled by us.
+     */
+    return document.getElementById(dropdownId);
+  }, [dropdownId]);
```
ethanashaw, line 227: We might be able to get away without using an ID here if we use the `dropdownProps.ref` directly. I haven't tested this, but here's an example from Gemini:
```tsx
import { useRef, useEffect } from 'react';

// No forwardRef wrapper needed in React 19!
function MyInput({ ref: externalRef, ...props }) {
  const internalRef = useRef(null);

  // Callback ref to sync both internal and external refs
  const mergedRef = (node) => {
    // 1. Assign to our internal ref for logic within this component
    internalRef.current = node;

    // 2. Assign to the ref passed by the parent
    if (typeof externalRef === 'function') {
      externalRef(node);
    } else if (externalRef) {
      externalRef.current = node;
    }
  };

  useEffect(() => {
    // Now you can use the node internally
    if (internalRef.current) {
      internalRef.current.style.backgroundColor = 'lightyellow';
    }
  }, []);

  return <input {...props} ref={mergedRef} />;
}
```
It's a bit messy because the `dropdownProps.ref` could be a callback or an object, but it seems like it would work if you want to avoid an ID here.

If we do this, we might also be able to consolidate a lot of these callbacks and effects into a single callback ref.

jmuzina:
> I have 1 bit of feedback from an a11y perspective: You don't need to include the word menu in the label as it's being read as "sub menu menu" I'm assuming this is because you have also set the menu role. Hope this feedback helps and let me know if you need more information.

@paul-geoghegan Do you think it would be OK for the label to just be "sub"? Do you know if this reading (sub menu menu) is consistent across screen readers - if there is some screen reader that doesn't read the role "menu" after the label "sub" then it might seem confusing to just open an element labeled "sub". Let me know if you have any other suggestions for how this should be read as well - thanks!

paul-geoghegan:
> @paul-geoghegan Do you think it would be OK for the label to just be "sub"? Do you know if this reading (sub menu menu) is consistent across screen readers - if there is some screen reader that doesn't read the role "menu" after the label "sub" then it might seem confusing to just open an element labeled "sub". Let me know if you have any other suggestions for how this should be read as well - thanks!

I think so, often enough controls are just marked as buttons so it would be sub menu button and maybe have an aria-haspopover but if your intent is to have it have the menu role then yes sub should probably be the label. I can't think of any screen readers that would ignore the menu label.

PR #1301, opened 2026-01-16
feat(prefixedIpInput): add prefixedInput, prefixedIpInput WD-32589

omarelkashef:
## Done

- Added prefixedInput and prefixedIpInput components

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- In Storybook, check prefixedInput and prefixedIpInput components.

### Percy steps

UI should look like in LXD UI, see screenshots below
image
image

### DIFF
--- a/src/components/PrefixedInput/PrefixedInput.scss
+++ b/src/components/PrefixedInput/PrefixedInput.scss
@@ -1,0 +1,22 @@
+@import "vanilla-framework";
+
+.prefixed-input {
+  position: relative;
+
+  .prefixed-input__input {
+    padding-top: 0.25rem;
+  }
+
+  .prefixed-input__text {
+    padding-left: $spv--large;
+    padding-top: 0.3rem;
+    pointer-events: none;
+    position: absolute;
+  }
+
+  &--with-label {
+    .prefixed-input__text {
+      top: 2.5rem;
+    }
+  }
+}
--- a/src/components/PrefixedInput/PrefixedInput.stories.tsx
+++ b/src/components/PrefixedInput/PrefixedInput.stories.tsx
@@ -1,0 +1,62 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import PrefixedInput from "./PrefixedInput";
+
+const meta: Meta<typeof PrefixedInput> = {
+  component: PrefixedInput,
+  tags: ["autodocs"],
+};
+
+export default meta;
+
+type Story = StoryObj<typeof PrefixedInput>;
+
+export const Default: Story = {
+  args: {
+    immutableText: "https://",
+    placeholder: "example.com",
+  },
+};
+
+export const WithLabel: Story = {
+  args: {
+    immutableText: "https://",
+    label: "Website URL",
+    placeholder: "example.com",
+  },
+};
+
+export const Disabled: Story = {
+  args: {
+    immutableText: "@",
+    label: "Username",
+    placeholder: "username",
+    disabled: true,
+  },
+};
+
+export const WithError: Story = {
+  args: {
+    immutableText: "https://",
+    label: "Website URL",
+    placeholder: "example.com",
+    error: "Invalid URL format",
+  },
+};
+
+export const WithHelpText: Story = {
+  args: {
+    immutableText: "User ID:",
+    label: "User Identifier",
+    placeholder: " Enter user ID",
+    help: "This will be used to identify your account",
+  },
+};
+
+export const Required: Story = {
+  args: {
+    immutableText: "https://",
+    label: "Website URL",
+    placeholder: "example.com",
+    required: true,
+  },
+};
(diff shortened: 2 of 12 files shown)


src/components/PrefixedInput/PrefixedInput.stories.tsx
edlerd, line 1: The two new stories should appear as components. Currently they are in their own category:

image

src/components/PrefixedIpInput/PrefixedIpInput.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const IPv4WithValue: Story = {
+  name: "IPv4 with value",
+  args: {
+    cidr: "192.168.1.0/24",
+    ip: "192.168.1.100",
+    name: "ip-address",
+    label: "IP Address",
+    onIpChange: (ip) => console.log("IP changed:", ip),
+  },
+};
```
edlerd, line 61: We should allow changing the value in the story view. Currently this is not possible and any inputs are discarded

image

src/components/PrefixedIpInput/PrefixedIpInput.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const IPv6WithValue: Story = {
+  name: "IPv6 with value",
+  args: {
+    cidr: "2001:db8::/32",
+    ip: "2001:db8::1",
+    name: "ipv6-address",
+    label: "IPv6 Address",
+    onIpChange: (ip) => console.log("IP changed:", ip),
+  },
+};
```
edlerd, line 107: Styling is slightly off with the last two colons in this view. They should both be on the same height.

image

src/components/PrefixedIpInput/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const lengths = [15, 11, 7, 3]; // Corresponding to 0-3 immutable octets
+    return lengths[immutableOctetsLength];
+  };
+  const maxLength = isIPV4 ? getIPv4MaxLength() : editable.length;
+  const placeholder = props.disabled ? "" : editable;
+  const setIp = (editableValue: string) => {
+    const fullIp = editableValue
+      ? isIPV4
+        ? `${immutable}.${editableValue}`
+        : `${immutable}${editableValue}`
```
edlerd, line 60: As discussed last week, this should add padding for IPv6.

If the address is shortened and the user entered an editable value that is not full length, then this should be `${immutable}:${editableValue}`.

If they entered a value that uses all fields (the parts between colons), or if they already entered a leading colon in the editable, then we should not add the additional colon.

src/components/PrefixedInput/PrefixedInput.scss
```diff
@@ -0,0 +1,18 @@
+@import "vanilla-framework";
+
+.prefixed-input {
```
edlerd, line 3: In Percy, there is too much spacing between the prefix and the editable part for the ip input. Though I can't reproduce it locally with Firefox and the dev server running.

image

src/index.ts
edlerd, line 1: We should also export the helper functions from the ip utils.

src/components/PrefixedIpInput/PrefixedIpInput.tsx
omarelkashef: As discussed, the current behavior will be left and the consuming app can choose to inject the colon for the shortened IPV6 notation if needed.

src/components/PrefixedInput/PrefixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+      const prefixWidth = prefixElement.getBoundingClientRect().width;
+      inputElement.style.paddingLeft = `${prefixWidth}px`;
+    }
+  }, []);
+
+  useLayoutEffect(() => {
+    updatePadding();
+
+    // Listen for window resize events (includes zoom changes)
+    window.addEventListener("resize", updatePadding);
```
edlerd: This listener needs to be removed on unmount.

Better use the `useListener(window, updatePadding, "resize", true);` hook that is local tooling doing the unmount plus throttling already.

omarelkashef: can useListener be called inside a callback?

edlerd: It would need to move outside to the main render function. It is a hook itself and we can't establish a hook inside another hook.

src/components/PrefixedIpInput/PrefixedIpInput.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    error: { control: "text" },
+    help: { control: "text" },
+    disabled: { control: "boolean" },
+    required: { control: "boolean" },
+  },
+  parameters: {
+    percy: {
+      waitForTimeout: 1000,
+    },
+  },
```
edlerd:
```suggestion
```
This doesn't fix the alginment. Lets remove it then.

PR #1313, opened 2026-02-11
fix: prevent nested portal conflicts between ConfirmationButton and ConfirmationModal

rubinaga:
## Done

- `ConfirmationButton` now passes its `Portal` to `ConfirmationModal` via `portalRenderer` instead of wrapping it directly.
- `ConfirmationModal` now prioritizes `portalRenderer`, then `renderInPortal`, then inline render.
- Added regression tests for confirm click flow, ignored portal overrides, and portal precedence behavior.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Open a ConfirmationButton example and click the trigger button.
- Verify the modal opens and stays open (does not immediately close).
- Click the modal confirm button and verify the confirm callback fires exactly once.
- Verify Esc/outside-click close behavior still works in ConfirmationButton flow.
- Render ConfirmationModal standalone with default props and verify it renders inline (not in portal).
- Render standalone ConfirmationModal with renderInPortal={true} and verify it renders in a portal.
- Render ConfirmationModal with both renderInPortal={true} and a custom portalRenderer and verify the custom renderer is used.

### DIFF
--- a/src/components/ConfirmationButton/ConfirmationButton.test.tsx
+++ b/src/components/ConfirmationButton/ConfirmationButton.test.tsx
@@ -156,4 +156,53 @@ describe("ConfirmationButton ", () => {
     expect(shouldShowModal).toHaveBeenCalled();
     expect(screen.getByText("Confirm")).toBeInTheDocument();
   });
+
+  it("executes onConfirm when clicking the modal confirm button", async () => {
+    const onConfirm = jest.fn();
+    render(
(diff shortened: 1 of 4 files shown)


edlerd (approved):
LGTM.

PR #1314, opened 2026-02-11
feat(tooltip): when tooltip position is left or right, auto adjust its position

edlerd:
## Done

- feat(tooltip) when tooltip position is left or right, auto adjust its position considering the above or below space being too small

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- check tooltip positioning in storybook

### Percy steps

- no change expected

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -135,13 +135,38 @@ const getPositionStyle = (
 export const adjustForWindow = (
   position: Position,
   fitsWindow: WindowFitment,
+  messageHeight: number = 0,
 ): Position => {
   let newPosition: string = position;
   if (!fitsWindow.fromLeft.fitsLeft && newPosition === "left") {
     newPosition = "top-right";
   }
+  if (
+    fitsWindow.fromTop.spaceAbove < messageHeight / 2 &&
+    newPosition === "left"
+  ) {
+    newPosition = "btm-left";
+  }
+  if (
+    fitsWindow.fromBottom.spaceBelow < messageHeight / 2 &&
+    newPosition === "left"
+  ) {
+    newPosition = "top-left";
+  }
   if (!fitsWindow.fromRight.fitsRight && newPosition === "right") {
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -242,7 +267,8 @@ const Tooltip = ({
 
   const onUpdateWindowFitment = useCallback(
     (fitsWindow: WindowFitment) => {
-      setAdjustedPosition(adjustForWindow(position, fitsWindow));
+      const messageHeight = messageRef.current.getBoundingClientRect().height;
```
Onibenjo: can we add a check for the ref here

```js
 if (!messageRef.current) {
        return;
      }
```

Onibenjo: before accessing it incase it's null

Onibenjo (commented):
a null check comment
asides thata LGTM

src/components/Tooltip/Tooltip.tsx
edlerd: good catch! adding the check below.

Onibenjo (approved):

src/components/Tooltip/Tooltip.tsx
```diff
@@ -242,7 +267,9 @@ const Tooltip = ({
 
   const onUpdateWindowFitment = useCallback(
     (fitsWindow: WindowFitment) => {
-      setAdjustedPosition(adjustForWindow(position, fitsWindow));
+      const messageHeight =
```
Onibenjo, line 270: I see the check was reverted or so 
@edlerd

Onibenjo: nvm

edlerd: Check is still there with the `?.`, just had to fix the formatting for the linter :)

PR #1339, opened 2026-04-01
fix(Modal): allow Escape to close ContextualMenu before closing Modal

### LINKED ISSUE #1305: `ContextualMenu` cannot be closed with ESCAPE key when inside a Modal
The ContextualMenu, when rendered inside a modal (see [story](https://canonical.github.io/react-components/?path=/story/components-contextualmenu--inside-modal)) cannot be closed with the Escape key. In other ContextualMenu stories, pressing Escape closes the menu; however this is not the case inside Modals. 

### To reproduce

1. Open a page with a contextual menu rendered inside of a modal.
2. Open the contextual menu.
3. Press Escape.
4. Notice that the contextual menu does not close.

### Expected behavior

Pressing the escape button when inside a focus trapped element (whether it be a modal, context menu, etc) causes the focus-trapped element to be dismissed (and ideally, for focus to be returned to the element that was triggered to open the focus trap per [WCAG](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html)). 

The above likely implies we would need some sort of global focus stack so that we can dismiss the right element and return focus to the correct element whenever focus-trap elements are closed (otherwise components that register Escape-to-close keylisteners will enter a race condition). I'm not sure if we have such a solution in place at the moment.

koushik717:
## Problem

Fixes #1305

When a `ContextualMenu` is open inside a `Modal`, pressing Escape closes the Modal immediately rather than the menu first. This breaks the expected layered-dismiss UX and is an accessibility issue.

## Root cause

`Modal.tsx` called `stopImmediatePropagation()` unconditionally. Because `ContextualMenu` renders via a Portal (a DOM sibling, not a React child), its own `document` keydown listener was silenced before it could fire.

## Fix — component-agnostic global LIFO escape-key stack

Following @edlerd's exploration in #1305, this implements the **global handler-stack** approach used by every major overlay library (Radix UI, Headless UI, Chakra UI).

### New: `src/hooks/useEscapeStack.ts`

A module-level LIFO array of callbacks. A single `document` keydown listener calls **only the top-of-stack handler** and then calls `stopImmediatePropagation()` — no component needs to know about any other.

```ts
// low-level imperative API (used internally)
const unregister = pushEscapeHandler(() => closeMyOverlay());

// React hook API (for consumers)
useEscapeStack(() => closeMyOverlay(), { isActive: isOpen });
```

### `useOnEscapePressed` — delegates to `pushEscapeHandler`

No API change; existing call-sites are unaffected.

### `Modal` — uses `useOnEscapePressed`

The Escape branch is removed from Modal's manual keydown map. Modal now registers via the stack, so it naturally yields to any overlay opened after it.

### `usePortal` — registers on the stack only while open

A new `useEffect([isOpen, closeOnEsc, …])` registers the handler when the portal opens and pops it when it closes. The stack always reflects the current visual overlay depth.

### Result

The most recently opened overlay handles Escape first, regardless of component type, DOM position, or whether it is a consumer-defined overlay (consumers can call `useEscapeStack` or `pushEscapeHandler` to join the same stack).

Two-step dismiss for Modal + ContextualMenu:
1. **First Escape** → ContextualMenu closes (top of stack)
2. **Second Escape** → Modal closes (now top of stack)

## Changes

- `src/hooks/useEscapeStack.ts` — New: LIFO stack + `useEscapeStack` hook + `pushEscapeHandler`
- `src/hooks/useEscapeStack.test.ts` — New: unit tests for stack ordering and propagation
- `src/hooks/useOnEscapePressed.ts` — Delegates to `pushEscapeHandler`
- `src/components/Modal/Modal.tsx` — Uses `useOnEscapePressed`; Escape removed from keydown map
- `src/components/Modal/Modal.test.tsx` — Updated + regression test for #1305
- `src/external/usePortal.ts` — Registers on escape stack only while portal is open
- `src/hooks/index.ts` / `src/index.ts` — Export new hook and utility

## Testing

```
yarn lint  ✓  (0 errors)
yarn test  ✓  (837 tests, 103 suites)
```

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -2,6 +2,7 @@ import classNames from "classnames";
 import React, { useId, useRef, useEffect } from "react";
 import type { HTMLProps, ReactNode, RefObject } from "react";
 import { ClassName, PropsWithSpread } from "types";
+import { useOnEscapePressed } from "hooks";
 
 export type Props = PropsWithSpread<
   {
@@ -88,33 +89,7 @@ export const Modal = ({
     }
   };
 
-  const handleEscKey = (
-    event: KeyboardEvent | React.KeyboardEvent<HTMLDivElement>,
-  ) => {
-    // If there is an open contextual menu dropdown inside (or alongside) this
-    // modal, let the Escape event propagate so that the menu's own keydown
-    // handler can close it first. The modal should only intercept Escape when
-    // no child overlay is open. This fixes the bug where a ContextualMenu
-    // rendered inside a Modal cannot be closed with the Escape key because
-    // stopImmediatePropagation() was called unconditionally, preventing the
-    // menu's document-level keydown listener from ever firing.
-    const hasOpenDropdown = !!document.querySelector(
-      '.p-contextual-menu__dropdown[aria-hidden="false"]',
-    );
-    if (hasOpenDropdown) {
-      return;
-    }
-    if ("nativeEvent" in event && event.nativeEvent.stopImmediatePropagation) {
-      event.nativeEvent.stopImmediatePropagation();
-    } else if ("stopImmediatePropagation" in event) {
-      event.stopImmediatePropagation();
-    } else if (event.stopPropagation) {
-      event.stopPropagation();
-    }
-    if (close) {
-      close();
-    }
-  };
+  useOnEscapePressed(() => close?.(), { isEnabled: !!close });
 
   useEffect(() => {
     if (focusRef?.current) {
@@ -127,14 +102,10 @@ export const Modal = ({
   }, [focusRef]);
 
   useEffect(() => {
-    const keyListenersMap = new Map([
-      ["Escape", handleEscKey],
-      ["Tab", handleTabKey],
-    ]);
-
     const keyDown = (event: KeyboardEvent) => {
-      const listener = keyListenersMap.get(event.code);
-      return listener && listener(event);
+      if (event.code === "Tab") {
+        handleTabKey(event as unknown as React.KeyboardEvent<HTMLDivElement>);
+      }
     };
 
     document.addEventListener("keydown", keyDown);
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -91,6 +91,19 @@ export const Modal = ({
   const handleEscKey = (
     event: KeyboardEvent | React.KeyboardEvent<HTMLDivElement>,
   ) => {
+    // If there is an open contextual menu dropdown inside (or alongside) this
+    // modal, let the Escape event propagate so that the menu's own keydown
+    // handler can close it first. The modal should only intercept Escape when
+    // no child overlay is open. This fixes the bug where a ContextualMenu
+    // rendered inside a Modal cannot be closed with the Escape key because
+    // stopImmediatePropagation() was called unconditionally, preventing the
+    // menu's document-level keydown listener from ever firing.
+    const hasOpenDropdown = !!document.querySelector(
+      '.p-contextual-menu__dropdown[aria-hidden="false"]',
+    );
+    if (hasOpenDropdown) {
+      return;
+    }
     if ("nativeEvent" in event && event.nativeEvent.stopImmediatePropagation) {
       event.nativeEvent.stopImmediatePropagation();
     } else if ("stopImmediatePropagation" in event) {
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -2,7 +2,7 @@ import classNames from "classnames";
 import React, { useId, useRef, useEffect } from "react";
 import type { HTMLProps, ReactNode, RefObject } from "react";
 import { ClassName, PropsWithSpread } from "types";
-import { useOnEscapePressed } from "hooks";
+import { useEscapeStack } from "hooks";
 
 export type Props = PropsWithSpread<
   {
@@ -89,7 +89,7 @@ export const Modal = ({
     }
   };
 
-  useOnEscapePressed(() => close?.(), { isEnabled: !!close });
+  useEscapeStack(() => close?.(), { isEnabled: !!close });
 
   useEffect(() => {
     if (focusRef?.current) {
--- a/src/hooks/useEscapeStack.ts
+++ b/src/hooks/useEscapeStack.ts
@@ -1,0 +1,61 @@
+/**
+ * A module-level LIFO stack of Escape-key handlers.
+ *
+ * Any overlay component (Modal, ContextualMenu, Drawer, …) can register a
+ * handler here. Only the most recently registered handler fires when Escape
+ * is pressed, so nested overlays always dismiss in the correct order
+ * regardless of their DOM position or portal placement.
+ */
+
+import { useEffect } from "react";
+
+const handlers: Array<() => void> = [];
+
+function onKeyDown(e: KeyboardEvent): void {
+  if (e.key !== "Escape" || handlers.length === 0) return;
+  e.stopImmediatePropagation();
+  handlers[handlers.length - 1]();
+}
+
+/**
+ * Push a callback onto the global escape-key stack.
+ * Returns a cleanup function that removes the handler (suitable for
+ * use as a `useEffect` cleanup return).
+ * Handlers are invoked in LIFO order — the last one pushed runs first,
+ * mirroring the visual stacking of overlays.
+ */
+export function pushEscapeHandler(handler: () => void): () => void {
+  if (handlers.length === 0) {
+    document.addEventListener("keydown", onKeyDown);
+  }
+  handlers.push(handler);
+  return () => {
+    const idx = handlers.lastIndexOf(handler);
+    if (idx !== -1) handlers.splice(idx, 1);
+    if (handlers.length === 0) {
+      document.removeEventListener("keydown", onKeyDown);
+    }
+  };
+}
+
+/**
+ * React hook that registers an Escape-key handler on the global LIFO stack
+ * for the lifetime of the component (or while `isActive` is true).
+ *
+ * The most recently registered handler always fires first, so nested overlays
+ * naturally dismiss in the correct order regardless of DOM structure.
+ *
+ * @param handler - Callback invoked when Escape is pressed and this handler
+ *   is at the top of the stack.
+ * @param options.isActive - When `false` the handler is not registered
+ *   (defaults to `true`).
+ */
+export const useEscapeStack = (
+  handler: () => void,
+  { isActive } = { isActive: true },
+): void => {
+  useEffect(() => {
+    if (!isActive) return undefined;
+    return pushEscapeHandler(handler);
+  }, [handler, isActive]);
+};
--- a/src/hooks/useEscapeStack.ts
+++ b/src/hooks/useEscapeStack.ts
@@ -27,6 +27,9 @@ function onKeyDown(e: KeyboardEvent): void {
  * mirroring the visual stacking of overlays.
  *
  * Safe to call in SSR environments — it no-ops when `document` is unavailable.
+ *
+ * This is the imperative primitive behind `useEscapeStack` and is not part
+ * of the public package API. React consumers should use `useEscapeStack`.
  */
 export function pushEscapeHandler(handler: () => void): () => void {
   if (typeof document === "undefined") return () => undefined;
@@ -46,7 +49,7 @@ export function pushEscapeHandler(handler: () => void): () => void {
 
 /**
  * React hook that registers an Escape-key handler on the global LIFO stack
- * for the lifetime of the component (or while `isActive` is true).
+ * for the lifetime of the component (or while `isEnabled` is true).
  *
  * The most recently registered handler always fires first, so nested overlays
  * naturally dismiss in the correct order regardless of DOM structure.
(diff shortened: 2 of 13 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -91,6 +91,19 @@ export const Modal = ({
...
+    // no child overlay is open. This fixes the bug where a ContextualMenu
+    // rendered inside a Modal cannot be closed with the Escape key because
+    // stopImmediatePropagation() was called unconditionally, preventing the
+    // menu's document-level keydown listener from ever firing.
+    const hasOpenDropdown = !!document.querySelector(
+      '.p-contextual-menu__dropdown[aria-hidden="false"]',
+    );
+    if (hasOpenDropdown) {
+      return;
+    }
```
jmuzina: This would make this behavior reliant on maintaining a list of other components that register their own escape focus handlers within Modal - making it quite tightly coupled with other components and tough to maintain. It would also reduce interoperability with custom consumer components that might register their own escape focus handlers (as they are not included in this query selector). I think any solution to this issue needs to be agnostic to the specific relationships between components.

@edlerd posted a thorough [exploration](https://github.com/canonical/react-components/issues/1305#issuecomment-3969146219) of this problem with some options to try out - maybe that can help improve this solution

jmuzina (changes requested):
Thanks for the contribution @koushik717 and apologies that this slipped beneath the radar - leaving a comment below with a suggestion for improvement!

src/components/Modal/Modal.tsx
koushik717: Thanks for the thorough feedback! I've reworked the fix completely  the CSS selector approach is gone.

The new implementation introduces a **component-agnostic global LIFO escape-key stack** (`src/hooks/useEscapeStack.ts`), following the approach @edlerd outlined. A single `document` listener sits in front of the stack and invokes only the top handler per keypress no component needs any knowledge of any other.

Both `Modal` (via `useOnEscapePressed`) and `usePortal` (with an open-aware `useEffect`) now register through the same stack. The most recently opened overlay always handles Escape first, regardless of component type, DOM position, or portal placement. Custom consumer overlays can join the same stack via the exported `useEscapeStack` hook or `pushEscapeHandler` utility.

All 837 tests pass. Happy to adjust anything further!

koushik717:
@jmuzina — branch rebased onto latest main (853 tests pass) and ready for re-review. Happy to adjust anything further!

koushik717:
@jmuzina — just a gentle follow-up in case this got lost in the queue. The LIFO escape-stack approach from the last revision fully addresses the coupling concern, and the branch is rebased on current `main` with all 853 tests passing. Whenever you have a moment for a second look it would be much appreciated!

koushik717:
@kimanhou @Ninfa-Jeon — apologies for the extra ping, but in case it's helpful: this PR fixes a real accessibility bug (#1305) where pressing Escape inside a Modal dismisses the Modal instead of the open ContextualMenu first. The fix uses a component-agnostic LIFO escape-key stack (same pattern as Radix UI / Headless UI), all 853 tests pass, and the branch is rebased on current `main`. If either of you has a moment to take a look it would be greatly appreciated!

jmuzina:
Hi, apologies for the delay. I'll get you a re-review today!

koushik717:
Thanks @Copilot  all four points addressed in the latest commit (16813f7):

- **Capture phase**: `pushEscapeHandler` now registers with `{ capture: true }` so the stack fires before any bubble-phase listener and cannot be bypassed by `stopPropagation()` on a child element.
- **SSR guard**: `pushEscapeHandler` no-ops when `document` is undefined.
- **Stable ref pattern**: Both `useEscapeStack` and `useOnEscapePressed` now keep the latest callback in a ref and register a stable wrapper, so inline callbacks do not cause re-registration on every render.
- **`closePortal` ref in usePortal**: The escape-stack effect no longer lists `closePortal` as a dependency  a ref tracks the latest value so identity changes to `onClose` don't re-register and reorder the stack.
- **Test updated**: The regression test now clicks to open the ContextualMenu after the Modal mounts, which accurately reflects real-world usage and correct LIFO registration order.

src/hooks/useEscapeStack.ts
```diff
@@ -0,0 +1,79 @@
...
+ * incorrectly move the overlay to the top of the stack.
+ *
+ * @param handler - Callback invoked when Escape is pressed and this handler
+ *   is at the top of the stack.
+ * @param options.isActive - When `false` the handler is not registered
+ *   (defaults to `true`).
+ */
+export const useEscapeStack = (
+  handler: () => void,
+  { isActive } = { isActive: true },
```
jmuzina: curious for your thoughts on naming here - why `isActive` here considering we use `isEnabled` in `useOnEscapePressed`? What's your opinion?

src/hooks/useOnEscapePressed.ts
```diff
@@ -1,26 +1,27 @@
...
 
 /**
  * Handle the escape key pressed.
+ * Registers the callback on the global escape-key stack so that nested
+ * overlays (modals, dropdowns, drawers, …) always dismiss in the correct
+ * LIFO order, regardless of DOM position or portal placement.
+ *
+ * A stable wrapper is kept in a ref so that inline callbacks do not cause
+ * the handler to be unregistered/re-registered on every re-render (which
+ * would incorrectly move this overlay to the top of the stack).
```
jmuzina: Noticed a small problem with implementing this on `useOnEscapePressed` as is (which is consumed by many other components) in the way you've done here:

If the consuming component uses this hook without a `isEnabled` guard (as SearchAndFilter and Navigation currently do) it will be added to the LIFO stack from `useEscapeStack` on mount regardless of whether it is actually open. 

Say for example a page has a Navigation and SearchAndFilter on the same page. Navigation registers first, then SearchAndFilter. You open Navigation's search input, press escape, and the escape event is routed to SearchAndFilter even though it's not the component you're interacting with (because it's at the top of the stack). 

I vibed up a brief reproducible example of this [here](https://github.com/jmuzina/react-components/commit/3a1045c8db63c42e79ac46722e14aca4f6e8e802) if you'd like to see more detail / try it out for yourself. As-is, this causes regressions for both SearchAndFilter and Navigation. 

So this solution relies heavily on the `isActive` passed to useEscapeStack being kept accurate. This is fine I think as it helps us avoid relying on DOM structure for stack order (which we should be doing to account for detached popups for instance) but we should note that it changes how important it is for callers of `useOnEscapePressed` to use `isEnabled` and could be a breaking change for [external consumers of this hook](https://github.com/search?q=%40canonical%2Freact-components+useonescapepressed+-repo%3Acanonical%2Freact-components+&type=code) that could see the same faulty behavior I found with SearchAndFilter and Navigation.

What do you think?

src/hooks/useEscapeStack.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+  handlers.push(handler);
+  return () => {
+    const idx = handlers.lastIndexOf(handler);
+    if (idx !== -1) handlers.splice(idx, 1);
+    if (handlers.length === 0) {
+      document.removeEventListener("keydown", onKeyDown, true);
+    }
+  };
+}
```
jmuzina, line 87: Why publicly export this function? The other hooks in this folder only export the hook function and maybe some types for it, this is an outlier. useEscapeStack is not a hook, it seems like a utility built for interacting with state from a hook (`useEscapeStack`). 

My first gut here is to use a global context provider instead, that holds the LIFO stack and also exposes this utility. Consumers would need to manually wrap their UI in this context provider to use the escape stack. This would help manage the severity of the breaking change I noted in my other comment, but also risks unexpected behavior if someone forgets to use this provider (though maybe this is less of a long term problem as https://github.com/canonical/pragma will be the new Canonical library in the long-term).  

What are your thoughts?

koushik717:
Thanks @jmuzina — really appreciate the careful re-review! All addressed in the latest commit (82eb54b):

- **Naming**: Renamed `isActive` -> `isEnabled` in `useEscapeStack` to match `useOnEscapePressed`.
- **Navigation/SearchAndFilter regression**: Great catch, and you were right, this was a real bug. Both components now pass `isEnabled` tied to their actual open state (`searchOpen` for Navigation, `!filterPanelHidden` for SearchAndFilter), so they only occupy a slot on the stack while their own overlay is actually visible. Added regression tests to both test files reproducing the scenario from your example.
- **`pushEscapeHandler` export**: Agreed it was an outlier — removed it from the public package exports (`hooks/index.ts` / `src/index.ts`). Only `useEscapeStack` (the hook) is part of the public API now, consistent with the rest of the hooks folder. `pushEscapeHandler` stays as an internal primitive used by `useOnEscapePressed`/`usePortal`. Happy to explore a context-provider approach as a follow-up for `pragma` if that's still useful, but wanted to keep this fix minimal and non-breaking for now.

All 856 tests pass, lint and typecheck are clean. Let me know if you'd like anything else adjusted!

jmuzina:
> this was a real bug. Both components now pass isEnabled tied to their actual open state (searchOpen for Navigation, !filterPanelHidden for SearchAndFilter), so they only occupy a slot on the stack while their own overlay is actually visible. Added regression tests to both test files reproducing the scenario from your example.

@koushik717  This works for these two components within React Components, but it would still be a breaking change for consumers of `useOnEscapePressed` who are outside of this repo (see [projects](https://github.com/search?q=useOnEscapePressed+%40canonical%2Freact-components+-repo%3Acanonical%2Freact-components&type=code)).

Before we write any more code: how would you recommend proceeding? Can this fix can be implemented without a breaking change to `useOnEscapePressed`? 

A breaking change is OK if absolutely necessary, we just need to be sure of it amongst ourselves and within Canonical's React Working Group

koushik717:
@jmuzina good question, and yes  this can be done with zero breaking change to `useOnEscapePressed`. Pushed in 99f88d1:

- **`useOnEscapePressed` is reverted to its exact pre-PR implementation**  a plain per-component `document` keydown listener, completely independent of the new stack. Behavior for every existing consumer (internal or external, with or without `isEnabled`) is byte-for-byte identical to before this PR. No one needs to change anything.
- **`Modal` now joins the LIFO stack via `useEscapeStack` directly** instead of going through `useOnEscapePressed`. Modal is the component that actually needs LIFO ordering against Portal-based overlays (`ContextualMenu` via `usePortal`, etc.), so it opts in explicitly.
- `usePortal` already used `pushEscapeHandler` directly (unchanged).
- The `isEnabled`-based changes to `Navigation`/`SearchAndFilter` from the last commit are harmless either way (the original `useOnEscapePressed` already supported `isEnabled` and just skips registering the listener), so I left those in as a small hygiene improvement, but they're no longer load-bearing for correctness.

So in short: the new LIFO stack (`useEscapeStack`/`pushEscapeHandler`) is purely additive and currently only opted into by `Modal` and `usePortal`. `useOnEscapePressed` is untouched/non-breaking, so there's no change in behavior for `SearchAndFilter`, `Navigation`, or any external consumer of that hook.

All 856 tests pass, lint and typecheck clean. Let me know if this addresses the concern or if you'd like the LIFO opt-in surfaced differently!

koushik717:
@jmuzina @Copilot — Copilot's two new comments were both legit, and the first one in particular gets right at the heart of your "before we write any more code" question, so let me address it properly here rather than with another drive-by patch.

**Copilot's finding #1 was correct, and worse than it sounds.** I reproduced it locally before fixing: a single open `ContextualMenu` (no Modal anywhere) was silently blocking an unrelated `useOnEscapePressed`-based listener elsewhere on the page, because `pushEscapeHandler`'s `onKeyDown` called `stopImmediatePropagation()` whenever the stack was non-empty — not just when nested overlays were involved. That's a much bigger blast radius than the original bug (#1305 was specifically about ContextualMenu *inside* Modal).

One nuance though: Copilot's framing ("reintroduces the original layered-dismiss issue") isn't quite right for **Modal** — Modal has *always* called `stopImmediatePropagation()` in capture phase while open, even on its own (that's pre-existing behavior, covered by the "should stop immediate Esc press propagation to other document listeners" test that predates this PR). The genuinely *new* over-blocking came from `usePortal` adopting the same exclusive behavior, which it never had before.

**Fix (pushed in d5bc53b):** each escape-stack entry now has an `exclusive` flag:
- `Modal` (via `useEscapeStack`) stays `exclusive: true` — unchanged, long-standing behavior.
- `usePortal` (ContextualMenu, Dropdown, etc., via `pushEscapeHandler`) is now `exclusive: false` — it handles its own dismissal but does **not** call `stopImmediatePropagation()`, so it no longer silences anything else when used standalone.
- When a non-exclusive entry sits on top of an exclusive one (ContextualMenu opened inside Modal), the first Escape is handled by the ContextualMenu only and the Modal underneath is skipped for that press — the two-step dismiss for #1305 still works.

This also directly answers your open question: **`useOnEscapePressed` remains completely untouched and non-breaking** — it's not part of the stack at all, so this `exclusive` flag is purely an internal detail of `useEscapeStack`/`pushEscapeHandler`, which are brand new in this PR (no external consumers yet).

**Copilot's finding #2** (`lastIndexOf(handler)` could remove the wrong entry if the same function reference were pushed twice) is also fixed — `pushEscapeHandler` now wraps each registration in a unique entry object, so unregistering always removes the exact entry that was pushed, regardless of reference equality.

**New tests** cover: exclusive vs non-exclusive ordering, the duplicate-reference unregister case, and a regression test (`ContextualMenu.test.tsx`) reproducing Copilot's exact scenario — a standalone open ContextualMenu no longer blocks an unrelated `useOnEscapePressed` listener.

All 860 tests pass, lint and typecheck clean.

@jmuzina — given this addresses both the breaking-change question and Copilot's findings without touching `useOnEscapePressed` at all, does this seem like a reasonable place to land, or would you prefer a different shape (e.g. surfacing `exclusive` differently, or going further toward unifying `useOnEscapePressed` onto the stack as a separate, deliberate breaking change for a future major version)? Happy to go either way — just want to make sure we're aligned before any further changes.

koushik717:
@jmuzina — no rush at all, just bumping this in case it slipped off the radar. Happy to answer any questions about the `exclusive`/non-exclusive approach in d5bc53b, or adjust the shape if you'd prefer something different. Let me know whenever you get a chance!

PR #1341, opened 2026-04-03
feat(a11y): add disabled label for a11y on buttons [AC-4127]

Stefan3002:
## Done

- Added a label visible only to assistive technology when a button is disabled to explain why it is disabled.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Use a ConfirmationButton
- Make it disabled and pass the disabledReasonLabel prop
- The button should be linked to the label via aria-describedby and the label should not be visible on the screen.

### Percy steps

## Fixes

Fixes: #[AC-4127](https://warthogs.atlassian.net/browse/AC-4127)

[AC-4127]: https://warthogs.atlassian.net/browse/AC-4127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ActionButton/ActionButton.tsx
+++ b/src/components/ActionButton/ActionButton.tsx
@@ -176,6 +176,7 @@ const ActionButton = ({
       className={buttonClasses}
       ref={ref}
       onClick={isDisabled ? onClickDisabled : onClick}
+      aria-describedby={buttonProps["aria-describedby"]}
       aria-disabled={isDisabled || undefined}
       style={
         height && width


edlerd (commented):
We tend to not use the aria-describedby and a hidden div but a bit more simple: Add a title to the button with the reason. That makes the reason also discoverable to anyone  on mouse over.

If you also use this approach, we probably don't need the changes proposed here. Wdyt @Stefan3002 ?

jmuzina:
I've referred this a11y best practices question to @paul-geoghegan for his thoughts!

I would suggest to avoid using a Tooltip as it would complicate usage if the consumer already uses a tooltip on this button, and would introduce a dependency from Button to Tooltip. I would lean towards using an `aria-describedby` and a hidden div then - the question then becomes how do we give this same information to users not using AT - maybe that's not something the component needs to solve at all, and we can expect the caller to create a visible label with the same or similar content  as `disabledReasonLabel`.

Stefan3002:
Hey @jmuzina and @paul-geoghegan . How does this look now? Thanks for the reviews and insights

src/components/ActionButton/ActionButton.tsx
```diff
@@ -78,7 +78,6 @@ const ActionButton = ({
   const [showSuccess, setShowSuccess] = useState(false);
   const ref = useRef<HTMLButtonElement>(null);
   const startLoadTime = useRef<Date | undefined>(undefined);
-
```
paul-geoghegan, line 81: Any particular reason we are removing this line?

Stefan3002: :laughing: no, my mistake.

Stefan3002: Fixed @paul-geoghegan

paul-geoghegan: I assumed as much but like to make sure, if you could revert that change that would be great!

edlerd:
I think the current state of the PR is ok as is and can be merged.

Further, I have some concerns about the higher-level problem we are trying to solve here. Adding an explanation as to why a button is not enabled should be as simple as passing in a string to the button itself. And this should be consistent among all our applications. So I think a solution in this library is best.

If we go away from using the title for clarifying why a button is not enabled, then I would like the `Button` and `ActionButton` components to offer a replacement prop to pass in the string explanation. With the current suggestion above made by Julie, this leaves too much for the use case to implement the aria-described-by reference in every usage.

Stefan3002:
Hey all! @jmuzina , @edlerd , @paul-geoghegan so if we are not going to implement "the fix" in this repo, I guess we can close this PR here as we can already use aria-describedby with no special handling thanks to ```{...buttonProps}```. What do you think?

Stefan3002:
Closing this PR as we agreed to not "fix" it ourselves but rather leave it to the users to fix themselves

PR #1367, opened 2026-04-29
feat: multiselect improvements

Ninfa-Jeon:
## Done

- Exposed search text as controlled/uncontrolled:
  - onSearchChange(value: string)
- Expose lifecycle callbacks:
  - onOpen
  - onClose
- Add native empty state support:
  - emptyState node or emptyMessage string
- New tests and stories for this functionality

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Review that the newly exposed values and callbacks are actually available (check storybook)

### DIFF
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -31,12 +31,19 @@ export type MultiSelectProps = {
   renderItem?: (item: MultiSelectItem) => ReactNode;
   dropdownHeader?: ReactNode;
   dropdownFooter?: ReactNode;
+  emptyState?: ReactNode;
+  emptyMessage?: string;
   showDropdownFooter?: boolean;
   variant?: "condensed" | "search";
   scrollOverflow?: boolean;
   isSortedAlphabetically?: boolean;
   hasSelectedItemsFirst?: boolean;
   id?: string;
+  searchValue?: string;
+  onSearchChange?: (value: string) => void;
+  onOpen?: () => void;
+  onClose?: () => void;
+  onResetSearch?: () => void;
 };
 
 type ValueSet = Set<MultiSelectItem["value"]>;
@@ -54,6 +61,8 @@ type MultiSelectDropdownProps = {
   onDeselectItem?: (item: MultiSelectItem) => void;
   onSelectItem?: (item: MultiSelectItem) => void;
   footer?: ReactNode;
+  emptyState?: ReactNode;
+  emptyMessage?: string;
   groupFn?: GroupFn;
   sortFn?: SortFn;
   hasSelectedItemsFirst?: boolean;
@@ -98,6 +107,8 @@ export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
   onDeselectItem,
   isOpen,
   footer,
+  emptyState,
+  emptyMessage,
   sortFn = sortAlphabetically,
   groupFn = getGroupedItems,
   hasSelectedItemsFirst = true,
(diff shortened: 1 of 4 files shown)


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -228,6 +272,20 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
...
+      return;
+    }
+
+    if (isDropdownOpen) {
+      onOpen?.();
+    } else {
+      onClose?.();
+    }
+    previousOpenState.current = isDropdownOpen;
+  }, [isDropdownOpen, onClose, onOpen]);
```
edlerd: This is pretty complicated logic. Can we instead have a wrapper around `setIsDropdownOpen` so it becomes internal and emits the callbacks? Or a bit simpler: whenever we call `setIsDropdownOpen`, we also trigger the callbacks for `onOpen` / `onClose`.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -201,25 +221,49 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
...
+  onResetSearch,
   help,
   helpClassName,
 }: MultiSelectProps) => {
   const buttonRef = useRef(null);
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
-  const [filter, setFilter] = useState("");
+  const [internalSearchValue, setInternalSearchValue] = useState("");
+  const previousOpenState = useRef(isDropdownOpen);
+  const filter = searchValue ?? internalSearchValue;
```
edlerd: I am not a fan of this duplicate approach. It introduces complexity now that we can either track the state internally or externally. How about always tracking it internally via `filter`, `setFilter` as before, but on every update we call `onSearchChange` with the new value? So any usage can subscribe to the search value and track it as well, but this component always tracks it internally. Wdyt?

edlerd (commented):
Generally LGTM, though some ideas and suggestions to simplify.

src/components/MultiSelect/MultiSelect.tsx
Ninfa-Jeon: You're right. This simplified the logic by a lot. Thanks!

src/components/MultiSelect/MultiSelect.tsx
Ninfa-Jeon: I have made some changes. What do you think?

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -201,25 +220,53 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
...
+    onSearchChange?.(value);
+  };
+
+  const resetSearch = () => {
+    if (!filter.length) {
+      return;
+    }
+    updateFilter("");
+    onResetSearch?.();
+  };
```
edlerd, line 265: I wonder if we need the `resetSearch`, it will trigger both, `onSearchChange` with empty string and `onResetSearch`. Maybe we can remove the `onResetSearch` and always use `onSearchChange`

edlerd (approved):
LGTM, one more idea to simplify the interface.

src/components/MultiSelect/MultiSelect.tsx
Ninfa-Jeon: Done, thanks!

edlerd (approved):

PR #1368, opened 2026-05-01
fix: searchbox cta warnings in form

Ninfa-Jeon:
## Done

The `SearchBox` component and consequently the `MultiSelect` component were throwing warnings when rendered within a formik form. This was due to the fact that the search button within `SearchBox` was being rendered without an explicit `type="submit"` prop.

- Resolved the above by setting `type="submit"` by default on `SearchBox` search button
- Exposed a `searchButtonType` prop in case the button type needs to change
- Exposed `dropdownClassName`, `inputClassName` and `footerClassName` props to fine tune `MultiSelect` as needed. This is because the dropdown is being rendered within a portal, making global override the only other solution.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- View the storybook story to verify that the custom classname props work
- Change the `FormikField` to render multiselect and try clicking the search button on it. The following error should not appear in console:
```
Warning: You submitted a Formik form using a button with an unspecified type attribute. Most browsers default button elements to type="submit". If this is not a submit button, please add type="button"
```

### DIFF
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -31,6 +31,9 @@ export type MultiSelectProps = {
   renderItem?: (item: MultiSelectItem) => ReactNode;
   dropdownHeader?: ReactNode;
   dropdownFooter?: ReactNode;
+  footerClassName?: string;
+  inputClassName?: string;
+  dropdownClassName?: string;
   emptyState?: ReactNode;
   emptyMessage?: string;
   showDropdownFooter?: boolean;
@@ -59,6 +62,8 @@ type MultiSelectDropdownProps = {
   onDeselectItem?: (item: MultiSelectItem) => void;
   onSelectItem?: (item: MultiSelectItem) => void;
   footer?: ReactNode;
+  footerClassName?: string;
+  dropdownClassName?: string;
   emptyState?: ReactNode;
   emptyMessage?: string;
   groupFn?: GroupFn;
@@ -105,6 +110,8 @@ export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
   onDeselectItem,
   isOpen,
   footer,
+  dropdownClassName,
+  footerClassName,
   emptyState,
   emptyMessage,
   sortFn = sortAlphabetically,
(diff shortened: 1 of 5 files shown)


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -355,7 +365,7 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
               required={required}
               type="text"
               value={filter}
-              className="multi-select__input"
+              className={classNames("multi-select__input", dropdownClassName)}
```
huwshimi: I think this one needs another prop: `inputClassName` and `dropdownClassName` should be applied to the element with `multi-select__dropdown`.

src/components/SearchBox/SearchBox.tsx
```diff
@@ -159,6 +159,7 @@ const SearchBox = React.forwardRef<HTMLInputElement, Props>(
           className="p-search-box__button"
           disabled={disabled}
           onClick={triggerSearch}
+          type="button"
```
huwshimi: There's a chance that people will be relying on the current behaviour to submit forms, so maybe to prevent a breaking change we should introduce a new prop, something like `searchButtonType` that defaults to `submit`.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -216,3 +216,37 @@ export const EmptyStateNode: Story = {
     ),
   },
 };
+
```
huwshimi: I don't think we need to have stories for these class props.

huwshimi (commented):
A few small changes here.

src/components/SearchBox/SearchBox.tsx
Ninfa-Jeon: does it make sense to default the button type to `submit`?...i know it could be a breaking change but a searchbox cta couldnt be used as submit under normal conditions

src/components/MultiSelect/MultiSelect.stories.tsx
Ninfa-Jeon: removed

src/components/SearchBox/SearchBox.tsx
Ninfa-Jeon: as per our discussion, i have added a prop defaulting to `submit`

src/components/MultiSelect/MultiSelect.tsx
Ninfa-Jeon: done.

huwshimi (approved):
Thanks for the updates!

PR #1374, opened 2026-05-08
fix(Modal): call close when pressing Esc

kimanhou:
## Done

- fix(Modal): call close when pressing Esc

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: # .

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -124,11 +124,11 @@ export const Modal = ({
       return listener && listener(event);
     };
 
-    document.addEventListener("keydown", keyDown);
+    document.addEventListener("keydown", keyDown, true);
     return () => {
-      document.removeEventListener("keydown", keyDown);
+      document.removeEventListener("keydown", keyDown, true);
     };
-  });
+  }, [close]);
 
   const handleContentOnMouseDown = () => {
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -114,21 +114,22 @@ export const Modal = ({
...
 
-    document.addEventListener("keydown", keyDown);
+    document.addEventListener("keydown", keyDown, true);
+
     return () => {
-      document.removeEventListener("keydown", keyDown);
+      document.removeEventListener("keydown", keyDown, true);
     };
-  });
+  }, [close]);
```
edlerd, line 131: Isn't this logically equivalent to what was there before? Only change I see is the additional `true` parameter on the `addEventListener` and `removeEventListener` calls and the `close` dependency. If those two are the fix, shall we keep the rest as it was?

kimanhou: Sure

edlerd (approved):
LGTM

PR #1384, opened 2026-06-18
fix(CheckboxInput & RadioInput): add Field to expose error help and other props used by Input

kimanhou:
## Done

- Add Field wrapper around `CheckboxInput` and `RadioInput` to expose `error`, `help` and other props used by Input.
This enhancement will allow us to stop using `<Input type="checkbox" />` and `<Input type="radio" />` and only use the dedicated components `CheckboxInput` and `RadioInput` instead.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

- No visual changes expected on existing stories.
- New stories added for CheckboxInput and RadioInput

### DIFF
--- a/src/components/CheckboxInput/CheckboxInput.stories.tsx
+++ b/src/components/CheckboxInput/CheckboxInput.stories.tsx
@@ -52,7 +52,41 @@ export const Default: Story = {
   name: "Default",
 };
 
-export const Children: Story = {
+export const Caution: Story = {
+  render: () => (
+    <CheckboxInput label="Caution" caution="This is a caution message" />
+  ),
+  name: "Caution",
+};
+
+export const Disabled: Story = {
+  render: () => <CheckboxInput label="Disabled" disabled />,
+  name: "Disabled",
+};
+
+export const Error: Story = {
+  render: () => (
+    <CheckboxInput label="Error" error="This is an error message" />
+  ),
+  name: "Error",
+};
+
+export const Help: Story = {
+  render: () => <CheckboxInput label="Help" help="This is a help message" />,
+  name: "Help",
+};
+
+export const Indeterminate: Story = {
+  render: () => <CheckboxInput label="Indeterminate" indeterminate />,
+  name: "Indeterminate",
+};
+
+export const Inline: Story = {
+  render: () => <CheckboxInput inline label="Inline" />,
+  name: "Inline",
+};
+
+export const LabelWithChildElements: Story = {
   render: () => (
     <CheckboxInput
       label={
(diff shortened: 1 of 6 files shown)


src/components/Input/Input.tsx
```diff
@@ -142,56 +141,66 @@ const Input = ({
...
+      >
+        <input
+          className={classNames("p-form-validation__input", className)}
+          ref={inputRef}
+          type={type}
+          {...commonProps}
+        />
+      </Field>
     );
   }
```
edlerd: Let's move this out of the else block to clarify this is the default. Wdyt?

src/components/CheckboxInput/CheckboxInput.stories.tsx
```diff
@@ -52,7 +52,41 @@ export const Default: Story = {
...
+};
+
+export const Disabled: Story = {
+  render: () => <CheckboxInput label="Disabled" disabled />,
+  name: "Disabled",
+};
+
+export const Error: Story = {
+  render: () => (
+    <CheckboxInput label="Error" error="This is an error message" />
```
edlerd, line 69: The bottom spacing seems a bit too low when comparing to the text input:

Image

Image

edlerd: I wonder if the left alignment of the error should be changed. Though it seems the error message is neither aligned with the label nor the contentin the input example. So probably fine as is.

edlerd (commented):
Thanks for initiating the streamlining process. Two comments below.

src/components/CheckboxInput/CheckboxInput.stories.tsx
kimanhou: I took <Input type="checkbox" error="lorem ipsum" /> as reference as it's what we currently use in LXD UI and I think it looks correct, no ?

image

image

edlerd: I think the current spacing is off, maybe this has to be a follow up fix upstream though:

image

kimanhou: It's because the checkbox has `bottom: 0.2em;` whereas the input is at the bottom of its element.
If I manually apply a `padding-bottom: 0.4rem` to `.p-checkbox`, this is how it looks:
image

This fix has to be done in Vanilla tho.

edlerd: Maybe best to keep it here as is (like merge the PR without changing the spacing). And notifying in vanilla about the findings for a potential follow up.

edlerd (approved):
LGTM, thanks.

PR #1398, opened 2026-07-15
feat(MultiSelect): Allow ReactNode to be used as checkbox label

ethanashaw:
## Done

- Add a `node` prop to `MultiSelectItem` which overrides the `label` prop for the checkbox label. The `label` will still be used for alphabetical sorting and showing the list of selected items. This prop is intended to show additional information about the item beyond just its name.

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### Percy steps

- Added a informational label to some items in the default story.

### DIFF
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -10,6 +10,7 @@ import { FadeInDown } from "./FadeInDown";
 export type MultiSelectItem = {
   label: string;
   value: string | number;
+  node?: ReactNode;
   group?: string;
 };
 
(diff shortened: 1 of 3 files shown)


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -10,6 +10,7 @@ import { FadeInDown } from "./FadeInDown";
 export type MultiSelectItem = {
   label: string;
   value: string | number;
+  node?: ReactNode;
   group?: string;
 };
```
edlerd, line 15: The new naming is slightly inconsistent with the [CustomSelect](https://github.com/canonical/react-components/blob/main/src/components/CustomSelect/CustomSelectDropdown/CustomSelectDropdown.tsx#L17-L20). There we have a `label: ReactNode` and a `text?: string` with a programmatic guard in `getOptionText` to ensure a string value is provided. Can we use the same approach here?

edlerd (commented):
I see why we'd need the flexibility, but there are some concerns on aligning the changes with existing components. See details below.

PR #1401, opened 2026-07-18
feat: add segmented control component

### LINKED ISSUE #1365: Implement Segmented Control component
Vanilla's [segmented control](https://vanillaframework.io/docs/patterns/segmented-control) component is not currently implemented in React Components. Implementing it in RC will prevent teams from needing to separately implement it.

mushroomgenie:
## Done

- This PR implements the Segmented Control component from Vanilla Framework in RC.

## QA

### Storybook

```shell
yarn start
```
segmented-control

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

`yarn build` and `npm back` is successfull.

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

## Fixes

Fixes: #1365  .

### DIFF
--- a/src/components/SegmentedControl/SegmentedControl.tsx
+++ b/src/components/SegmentedControl/SegmentedControl.tsx
@@ -1,0 +1,74 @@
+import classNames from "classnames";
+import React, { ReactNode, HTMLProps, useState } from "react";
+import type { ClassName } from "types";
+
+export type Segments<P = null> = {
+  /**
+   * Label to be displayed inside the segment.
+   */
+  label: ReactNode;
+  /**
+   * Content to be displayed inside the segment.
+   */
+  segmentContent: ReactNode;
+  /**
+   * Icon to be displayed alongside the label of the segment.
+   */
+  segmentIcon?: ReactNode;
+} & (HTMLProps<HTMLElement> | P);
+
+export type Props<P = null> = {
+  /**
+   * Optional classes applied to the parent element.
+   */
+  className?: ClassName;
+  /**
+   * List of segments present in the element.
+   */
+  segments: Segments<P>[];
+};
+/**
+ * This is the [React](https://reactjs.org/) component for Vanilla [SegmentedControl](https://vanillaframework.io/docs/patterns/segmented-control).
+SegmentedControl organises and allows navigation between groups of content that are related and at the same level
+of hierarchy.
+ */
+const SegmentedControl = ({
+  className,
+  segments,
+}: Props): React.JSX.Element => {
+  const [activeIndex, setActiveIndex] = useState<number>(0);
+  return (
+    <div
+      className={classNames("p-segmented-control", className)}
+      data-testid="segmented-control-div"
+    >
+      <div className={classNames("p-segmented-control__list")} role="tablist">
+        {segments.map((segment, i) => {
+          return (
+            <button
+              aria-selected={activeIndex === i}
+              className={classNames("p-segmented-control__button")}
+              role="tab"
+              key={i}
+              onClick={() => setActiveIndex(i)}
+            >
+              {segment.segmentIcon ? (
+                <>
+                  {segment.segmentIcon}
+                  <span>{segment.label}</span>
+                </>
+              ) : (
+                segment.label
+              )}
+            </button>
+          );
+        })}
+      </div>
+      <div tabIndex={activeIndex} role="tabpanel" data-testid="text-content">
+        {segments[activeIndex].segmentContent}
+      </div>
+    </div>
+  );
+};
+
+export default SegmentedControl;
--- a/src/components/SegmentedControl/SegmentedControl.tsx
+++ b/src/components/SegmentedControl/SegmentedControl.tsx
@@ -63,11 +63,7 @@ const SegmentedControl = ({
           );
         })}
       </div>
-      <div
-        tabIndex={activeIndex}
-        role="tabpanel"
-        aria-labelledby={segments[activeIndex].label}
-      >
+      <div role="tabpanel" aria-labelledby={segments[activeIndex].label}>
         {segments[activeIndex].content}
       </div>
     </div>
--- a/src/components/SegmentedControl/SegmentedControl.tsx
+++ b/src/components/SegmentedControl/SegmentedControl.tsx
@@ -1,23 +1,24 @@
 import classNames from "classnames";
-import React, { ReactNode, HTMLProps, useState } from "react";
-import type { ClassName } from "types";
+import Icon, { ICONS } from "components/Icon";
+import React, { ReactNode, useState } from "react";
+import type { ClassName, ValueOf } from "types";
 
-export type Segments<P = null> = {
+export type Segments = {
   /**
    * Label to be displayed inside the segment.
    */
-  label: ReactNode;
+  label: string;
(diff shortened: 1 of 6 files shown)


src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+ */
+const SegmentedControl = ({
+  className,
+  segments,
+}: Props): React.JSX.Element => {
+  const [activeIndex, setActiveIndex] = useState<number>(0);
+  return (
+    <div
+      className={classNames("p-segmented-control", className)}
+      data-testid="segmented-control-div"
```
edlerd:
```suggestion
```
We should use other ways of selecting and avoid selectors solely for tests.

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+    >
+      <div className={classNames("p-segmented-control__list")} role="tablist">
+        {segments.map((segment, i) => {
+          return (
+            <button
+              aria-selected={activeIndex === i}
+              className={classNames("p-segmented-control__button")}
+              role="tab"
+              key={i}
```
edlerd:
```suggestion
              key={segment.label}
```
Better use an id from the contents. Using the number leads to rendering problems when the contents of the segments array changes. As when a single entry is removed or added to it..

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+   */
+  label: ReactNode;
+  /**
+   * Content to be displayed inside the segment.
+   */
+  segmentContent: ReactNode;
+  /**
+   * Icon to be displayed alongside the label of the segment.
+   */
+  segmentIcon?: ReactNode;
```
edlerd:
```suggestion
  iconName?: ValueOf<typeof ICONS> | string;
```
I would avoid passing in a random react node here and only allow a reference to the internal icon component. So below we should use `<Icon name={segment.iconName} />`. Wdyt?

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export type Segments<P = null> = {
+  /**
+   * Label to be displayed inside the segment.
+   */
+  label: ReactNode;
+  /**
+   * Content to be displayed inside the segment.
+   */
+  segmentContent: ReactNode;
```
edlerd:
```suggestion
  content: ReactNode;
```
This is the segment type, no need to be verbose on the name here.

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+                  <span>{segment.label}</span>
+                </>
+              ) : (
+                segment.label
+              )}
+            </button>
+          );
+        })}
+      </div>
+      <div tabIndex={activeIndex} role="tabpanel" data-testid="text-content">
```
edlerd:
```suggestion
      <div tabIndex={activeIndex} role="tabpanel" aria-labelledby={...}>
```
We don't need the test id here.

We should add a `aria-labelledby` referencing the id of the button for this segment.

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [activeIndex, setActiveIndex] = useState<number>(0);
+  return (
+    <div
+      className={classNames("p-segmented-control", className)}
+      data-testid="segmented-control-div"
+    >
+      <div className={classNames("p-segmented-control__list")} role="tablist">
+        {segments.map((segment, i) => {
+          return (
+            <button
```
edlerd, line 46: the button should have an id.

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
+import classNames from "classnames";
+import React, { ReactNode, HTMLProps, useState } from "react";
+import type { ClassName } from "types";
+
+export type Segments<P = null> = {
```
edlerd:
```suggestion
export type Segments = {
```
Why do we need the generic `P` type here?

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+  label: ReactNode;
+  /**
+   * Content to be displayed inside the segment.
+   */
+  segmentContent: ReactNode;
+  /**
+   * Icon to be displayed alongside the label of the segment.
+   */
+  segmentIcon?: ReactNode;
+} & (HTMLProps<HTMLElement> | P);
```
edlerd:
```suggestion
};
```
We ignore any of the props, except the ones explicitly mentioned above. Why the additional type here? It seems to announce non-supported features.

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export type Props<P = null> = {
+  /**
+   * Optional classes applied to the parent element.
+   */
+  className?: ClassName;
+  /**
+   * List of segments present in the element.
+   */
+  segments: Segments<P>[];
```
edlerd:
```suggestion
  segments: Segments[];
```
The segment does not need the generic.

edlerd (commented):
Thank you for the contribution!

Quite a few concerns and thoughts on the proposed component can be found below.

mushroomgenie:
@edlerd thank you for your review, I'll take a look at the proposed changes and update the PR.

mushroomgenie:
@edlerd made the changes, can you please check ?

src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,77 @@
...
+                </>
+              ) : (
+                segment.label
+              )}
+            </button>
+          );
+        })}
+      </div>
+      <div
+        tabIndex={activeIndex}
```
edlerd:
```suggestion
```
We probably shouldn't set a tab index here and rely on the default tab order. Why did you decide to include this?

edlerd (commented):
Thanks for the adjustments. Nearly perfect, just one question left below then it can be merged from my side.

Also, can you please rebase on main to get the latest changes in?

src/components/SegmentedControl/SegmentedControl.tsx
mushroomgenie: since tabIndex is a state variable, I decided to set it here. I had a thought that by default the first index would be the one selected. Should I remove it ?

mushroomgenie:
@edlerd removed the tabIndex

edlerd (approved):
LGTM, thanks for the follow-up fixes!

mushroomgenie:
@edlerd the snapshot failed, I'll re run the test and push it.

mushroomgenie:
@edlerd thanks for reviewing, I have pushed the updated snapshot.

edlerd:
Thanks for your contributions @mushroomgenie !

PR #1411, opened 2026-08-12
feat(SearchAndFilter): use Chip's appearance

kimanhou:
## Done

- Add optional use of Chip's `appearance` in SearchAndFilter

## QA

### Storybook

```shell
yarn start
```

### QA in your project

from `react-components` run:

```shell
yarn build
npm pack
```

```shell
yarn add <path-to-tarball>
```

### QA steps

- Steps for QA.

### Percy steps

- New story for SearchAndFilter with Chip appearance

## Screenshots

Screenshot from 2026-08-10 17-16-04

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -130,9 +130,9 @@ const Chip = ({
   );
 
   const chipClassName = classNames(
+    "p-chip",
     {
       [`p-chip--${appearance}`]: !!appearance,
-      "p-chip": !appearance,
       "is-dense": isDense,
       "is-readonly": isReadOnly,
       "is-inline": isInline,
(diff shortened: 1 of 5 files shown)


src/components/Chip/Chip.tsx
```diff
@@ -130,9 +130,9 @@ const Chip = ({
   );
 
   const chipClassName = classNames(
+    "p-chip",
     {
       [`p-chip--${appearance}`]: !!appearance,
-      "p-chip": !appearance,
```
edlerd, line 135: This seems like a regression, we should only have one `p-chip` prefixed class on the chip, right? Should this be reverted?

edlerd (commented):
Mostly LGTM, one question below.

src/components/Chip/Chip.tsx
kimanhou: Some CSS only applies to `p-chip`. I need to make this change because otherwise, the chips weren't vertically aligned with one another.

edlerd (approved):

PR #1416, opened 2026-08-25
fix(step): ensure step status icon style is not overwritten by icon's

lorumic:
## Done

- Ensured that the step status icon style is not overwritten by the icon's.

# Before

image

# After

image


tarek-hanna (approved):
LGTM thx Michele 👍

PR #1422, opened 2026-09-08
feat(OutputField): add truncate prop for overflowing values

omarelkashef:
## Done

- Added an optional `truncate` prop to `OutputField`.
- Applied a dedicated truncation class to the `<output>` element when enabled.
- Set the full output value as the `title` attribute when truncation is enabled.
- Added a Storybook `Truncate` example and test coverage for the class and title attribute.
- Fixed overflowing values by making the normally inline `<output>` a bounded block box, allowing `text-overflow: ellipsis` to apply.

## QA

### Storybook

    yarn start

Open the `OutputField` `Truncate` story and verify a long value is displayed on one line, clipped with an ellipsis when constrained, and exposes the complete value in its native title tooltip.

### QA in your project

from `react-components` run:

    yarn build
    npm pack

    yarn add <path-to-tarball>

### QA steps

- Render an `OutputField` with a long `value` and `truncate`.
- Constrain the field container to a width narrower than the value.
- Verify the value is displayed with an ellipsis rather than overflowing.
- Hover the value and verify the native title tooltip shows the complete value.
- Render the same field without `truncate` and verify its existing display behavior is unchanged.

### Percy steps

- The new `OutputField` `Truncate` Storybook story is expected to display a single-line, ellipsized long value.


src/components/OutputField/OutputField.tsx
```diff
@@ -45,7 +51,13 @@ export const OutputField: FC<Props> = ({
       className="output-field"
       required={required}
     >
-      <output id={id} className="mono-font u-sv2">
+      <output
+        id={id}
+        title={truncate ? value : undefined}
+        className={classNames("mono-font", "u-sv2", {
+          "output-field-truncate": truncate,
```
edlerd, line 58:
```suggestion
          "u-truncate": truncate,
```
Can we use the existing helper from vanilla?

kimanhou:
I thought the original idea was for the component to accept classnames ? It would be more flexible.

src/components/OutputField/OutputField.tsx
```diff
@@ -23,6 +24,10 @@ export type Props = {
    * Whether the output field is required.
    */
   required?: boolean;
+  /**
+   * Whether the value should be truncated with an ellipsis when it overflows, showing the full value in a title tooltip.
+   */
+  truncate?: boolean;
```
edlerd: I wonder if we should make it truncate by default. Or even remove this prop and always have it truncate. I can't imagine a case where we would not want this to be enabled.

omarelkashef:
> I thought the original idea was for the component to accept classnames? It would be more flexible.

I found out that giving the component a `u-truncate` is not enough because the the `output` element is not block so truncation won't work on it. Plus, I think truncation should be supported automatically rather than having to pass a classname that makes `output` block and add trunctaion.

edlerd (approved):
LGTM, thanks

Issue #1225, opened 2025-07-15
Paginated MainTable sorting only sorts the visible page


steverydz:
When using the `MainTable` component with `sortable` and pagination, only the visible rows are sorted, not the whole data set.

Perhaps this is intended behaviour, but we've had users complain that they expect the whole data set to be sorted so if we don't want to change the current behaviour, we need an option to enable sorting of the whole data set.

gbosetti:
@steverydz @edlerd what do you thnk of the approch in that PR? adding a dropdown so both sorting strategies can coexist (in case people got used to the current way - making them and those wanting a ful dataset sort, happy)

edlerd:
> [@steverydz](https://github.com/steverydz) [@edlerd](https://github.com/edlerd) what do you thnk of the approch in that PR? adding a dropdown so both sorting strategies can coexist (in case people got used to the current way - making them and those wanting a ful dataset sort, happy)

Thanks for looking into this and proposing a solution.

I think sorting only the current page is never the user intent and we always want to sort across all pages. Personally I don't see a situation where I'd want to sort only the current page.

Shall we simplify the solution and not have a switch between the two sorting strategies? Or do you see a reason to have both available?

canonical-steverydz:
Thanks for looking at this! Personally I'd expect a UI to sort the whole dataset, so I agree with @edlerd suggestion to simplify :)

gbosetti:
@edlerd thanks for the quick reply. My only concern was people who have been relying on the current behaviour and might treat sorting just the visible page as intended.

But since you and @canonical-steverydz agree, I've simplified it and pushed: sortable tables inside TablePagination now always sort the whole data set, with no switch. See #1426. :)

Stefan3002:
Well, not including a flag *could* potentially break workarounds people might have used. So while fixing the broken behavior, we are essentially breaking fixes for the broken behavior 😆 (that sounds funny). But yes, I agree with @edlerd and @canonical-steverydz , we should fix the broken behavior mostly because it goes against what users would expect i.e: sort the whole dataset, even if we *could* break workarounds. Leaving a flag there means the default behavior must be set to the broken one in order to keep backwards compatibility, but that is just wrong, we would have a broken component that users can fix with a flag. 
It sounds weird right?

I suggest we definitely fix it with no flag.

PR #1426, opened 2026-09-16
feat(TablePagination): sort the whole data set before pagination

gbosetti:
Fix attempt for #1225

## Description

When a sortable `MainTable` is wrapped in `TablePagination`, clicking a header only sorts the rows on the current page. Now the whole data set is sorted before it is paginated, so the order is correct across all pages.

An opt-in dropdown (sort this page / sort all pages) was considered first, but as agreed in #1225, sorting only the visible page is never what users want, so there is no switch: sortable tables inside `TablePagination` always sort all pages.

- `TablePagination` sorts the full data set for a sortable `MainTable` child and keeps the header arrows in sync. Non-sortable tables and externally controlled pagination are unchanged.
- `useSortTableData` hook (ported from lxd-ui, see #1225) does the sorting and is also exported for anyone paginating differently.
- `sortRows` moved out of `MainTable.tsx` so both share the same comparison. No behaviour change to `MainTable`.
- Tests for the hook and for `TablePagination`.

#### Demo

| Problem | Alternative Solution | Final Solution
| - | - | - |
| https://github.com/user-attachments/assets/a21aebd5-9cf7-439c-8724-67792463b688 | https://github.com/user-attachments/assets/36f6442f-5ed8-41ce-b8f8-94f177e4320b | https://github.com/user-attachments/assets/79573daf-b978-47e3-9181-49bda6282c11 |
| Sorting by cores only reorders the two rows on the page, so the row with the most cores never shows up. | Choosing "Sort all pages" sorts every row first, so the row with the most cores comes to the top. |  No dropdown: sorting always covers every row. |

## QA steps

- `yarn start`, open [TablePagination → RenderAbove](http://localhost:9009/?path=/story/components-tablepagination--render-above), pick `2/page` and click "Cores" twice.
- Page 1 shows Waiting (8 cores), Ready (2 cores). On `main` it shows Ready, Idle because only the visible page was sorted.
- Click "Cores" a third time: the sort is cleared and the original order comes back.


src/components/TablePagination/TablePagination.tsx
```diff
@@ -149,14 +162,42 @@ const TablePagination = (props: Props) => {
...
     pageData: internalData,
-  } = usePagination(externallyControlled ? [] : data, {
-    itemsPerPage: internalPageSize,
-    autoResetPage: true,
-  });
+  } = usePagination(
+    externallyControlled ? [] : sortable ? sortedData.rows : data,
+    {
+      itemsPerPage: internalPageSize,
+      autoResetPage: true,
```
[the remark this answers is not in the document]
gbosetti: The PR description still described an earlier version with a "sort scope" dropdown. That option was dropped after discussion in #1225, where the maintainers asked to always sort the whole data set with no switch. Title and description are now updated to match the code.

src/hooks/useSortTableData.ts
```diff
@@ -0,0 +1,119 @@
...
+   * `defaultSortDirection` prop.
+   */
+  defaultSortDirection?: SortDirection;
+  /**
+   * A custom function to be used when sorting, with the same signature as the
+   * MainTable `sortFunction` prop.
+   */
+  sortFunction?: MainTableProps["sortFunction"];
+};
+
```
edlerd: This is duplicating the interface documentation in lines 58 to 72. Rest of the repo is using `@param`, so I think we should prefer that.

src/hooks/useSortTableData.ts
```diff
@@ -0,0 +1,119 @@
...
+ * @param {Object} options
+ * @param {MainTableRow[]} options.rows - The rows to sort.
+ * @param {string} [options.defaultSort] - The key to sort the rows by initially.
+ * @param {SortDirection} [options.defaultSortDirection] - The initial sort direction, as for MainTable.
+ * @param {Function} [options.sortFunction] - A custom sort function, as accepted by MainTable.
+ */
+export const useSortTableData = ({
+  rows,
+  defaultSort = null,
+  defaultSortDirection,
```
edlerd: There should be a default applied here, no?

src/hooks/useSortTableData.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Reset the sort if the defaults change after mount, as MainTable does.
+  const resetSort = useEffectEvent(() =>
+    setSort((current) =>
+      current.sortKey === defaultSort &&
+      current.sortDirection === defaultSortDirection
+        ? current
+        : { sortKey: defaultSort, sortDirection: defaultSortDirection },
+    ),
+  );
+  useEffect(() => resetSort(), [defaultSort, defaultSortDirection]);
```
edlerd: This seems a bit odd. If the defaults change, we probably shouldn't set them as the current sorting order. A user might have changed it to customize, and we should not override it to what the new default is.

